PR #1966, opened 2018-02-17
Tests: Don't hardcode errno constants

jwilk:
The values of `ENOENT` and `ESRCH` are architecture-dependent, so don't assume they're always 2 and 3.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


pbeckingham:
Thank you.

PR #1971, opened 2018-02-27
[feat] count newlines in PS1 for burndown max height

a-tal:
#### Description

The burndown charts blindly assume a single line PS1. This patch accounts for multi-line PS1's.

```
$ ./problems
Failed:
burndown.t                          4
project.t                           4

Unexpected successes:

Skipped:

Expected failures:
```

### DIFF
--- a/src/commands/CmdBurndown.cpp
+++ b/src/commands/CmdBurndown.cpp
@@ -171,7 +171,17 @@ Chart::Chart (char type)
   // How much space is there to render in?  This chart will occupy the
   // maximum space, and the width drives various other parameters.
   _width = Context::getContext ().getWidth ();
-  _height = Context::getContext ().getHeight () - 1;  // Allow for new line with prompt.
+
+  // check the PS1 to determine how many newlines to leave for the prompt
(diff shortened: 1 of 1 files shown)


pbeckingham:
Have you seen `reserved.lines` in `man taskrc`?

a-tal:
maybe this is better suited to be a function in `Context.cpp`, up to you... I haven't noticed anywhere else yet where this is needed, but I just started using taskwarrior

pbeckingham:
Point being, we already support multiple line prompts, in a more portable way.

a-tal:
no, I saw the problem and started grep'ing code for `- 1` 😄 

not sure if that's actually more portable though. another thing to change/forget to change, rather than read from the source

pbeckingham:
It's portable to any shell, and usable for purposes other than prompts.  $PS1 is not.

a-tal:
sure thing.

do you want me to open a bug report for `reserved.lines` not affecting the burndown height then?

ad-si:
@a-tal I opened an issue.

PR #1974, opened 2018-03-04
Docs: improve verbose description in taskrc.5.in

rudis:
Hi,

This tries to improve the verbose description to make it more clear what is affected by the header/footnote settings. As first time taskwarrior user I was confused what falls under those settings. I'm not entirely happy with the change though, so feel free to improve it.

Regards
Simon


pbeckingham:
Good changes, thanks.

PR #1978, opened 2018-03-07
Use nullptr instead lf C-styled NULL

omtcvxyz:
Replace C-styled `NULL` with `nullptr`, which should be used in modern C++ projects.

This patch was generated by running clang-tidy [modernize-use-nullptr](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html) check.


pbeckingham:
Thank you.

PR #1980, opened 2018-03-08
Update notes regarding C++14 support

omtcvxyz:
4a929197ae8c32d62a109e1ab063217d0c57c01b introduced C++14 support. After these changes, the build will now terminate unless the compiler supports C++14 dialect. The documentation does not reflect that: this commit updates few reference files to actually say that the building process requires a C++14-compliant compiler.


pbeckingham:
Thank you.

PR #2009, opened 2018-04-09
Docs: mention "urgency" as possible attribute for filters

rudis:
Hello,

I noticed that the urgency attribute is not defined in the man page, this patch adds it.

However I don't know if there are additional unlisted attributes which can be used in filters. If so, please tell me and I'll update the patch.

Regards
Simon


pbeckingham:
Thank you.

PR #2051, opened 2018-07-27
doc/man: ready tasks sorted with started tasks 1st

sim590:
Documentation of the default sort method for `task ready` being *started-then-urgent*.


pbeckingham:
Thanks.

PR #2053, opened 2018-08-11
Do not set CMP0037 on CMake 3.11.0 and above

janikrabe:
This policy is no longer necessary because CMake issue #16062, which caused incorrect warnings to be shown, has been resolved in 3.11.0.


pbeckingham:
Thanks.

Issue #1917, opened 2018-02-14
[TW-1904] wrong order under projects command


taskwarrior:
_Beka on 2017-04-02T14:07:57Z says:_ 

 When having both projects 'a.b' and project 'a-b', the subproject indentation is wrong. For example:

```
$ task rc.data.location=/tmp/ add pro:a-b test1
$ task rc.data.location=/tmp/ add pro:a.b test2
$ task rc.data.location=/tmp/ projects         
Using alternate data.location /tmp/

Project Tasks
a           1
a-b         1
  b         1

3 projects (2 tasks)
```

Note that the indented 'b' subproject is under 'a-b', and not under 'a' as it is it's parent project.

taskwarrior:
Migrated metadata:
```
Created: 2017-04-02T14:07:57Z
Modified: 2017-04-02T16:36:30Z
```

taskwarrior:
_Paul Beckingham on 2017-04-02T16:36:30Z says:_ 

 Good catch, thanks.

PR #2073, opened 2018-11-03
Fix issue #1917 [TW-1904] wrong order under projects command

grimmn:
#### Description

Fixes incorrect ordering of projects as reported in TW-1904.  Default
string comparison was not respecting the higher sorting precedence
needed for the project delimiter character.

Failed:

Unexpected successes:

Skipped:

Expected failures:

### DIFF
--- a/src/commands/CmdProjects.cpp
+++ b/src/commands/CmdProjects.cpp
@@ -36,6 +36,21 @@
 #include <main.h>
 
 ////////////////////////////////////////////////////////////////////////////////
+template <char delimiter>
+struct ProjectCompare
+{
+  bool operator() (std::string a, std::string b)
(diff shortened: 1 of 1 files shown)


mrossinek:
Sorry Jonathan, I did not see your PR before working on my version #2093.
I quite like your approach here! Initially I tried to come up with something similar but didn't know enough details about the map interface on the spot. Maybe we can get some input from @tbabej with which of these two approaches we should continue.

grimmn:
@mrossinek feel free to copy my approach here in your own PR if you'd like.

mrossinek:
I'll add a comment to my PR and let the maintainers have a look at it. They should decide, imo

PR #2093, opened 2018-12-31
Fix issue #1917 [TW-1904] wrong order under projects command

mrossinek:
#### Description

This pull request fixes the issue _[TW-1904] wrong order under projects command_ #1917.
The issue arises due to the fact that the ASCII code of the '.' (dot) is bigger than that of the '-' (hyphen). Thus, the latter will be processed first when traversing a map of strings.
If this is to be fixed without major changes to the underlying Table class (of libshared), the (in my opinion, safe) assumption that nobody will use the '=' (equal) sign within project names comes in handy and can be used to replace all '-' occurrences by '=', pushing these project names behind those including '.' due to the higher ASCII code. Once such a name is encountered the substitution is reverted.
This pull request also extends the same functionality to the summary command.

Unit-tests for the working bug fix have also been added to the corresponding files.

  I am currently trying to clean up the results of the unit tests. Thus, the output of ./problems might vary for you. Simply note the added tests in ./project.t and ./summary.t:
```
> ./problems
Failed:                        

Unexpected successes:          

Skipped:                       

Expected failures:             
project.t                           3
```
```
> ./project.t | grep 1904
ok 9 - project.t: 1904: verify correct order under projects command
```
```
> ./summary.t | grep 1904
ok 1 - summary.t: 1904: verify correct order under summary command
```

### DIFF
--- a/src/commands/CmdProjects.cpp
+++ b/src/commands/CmdProjects.cpp
@@ -115,7 +115,8 @@ int CmdProjects::execute (std::string& output)
     sort_projects (sorted_view, unique);
 
     // construct view from sorted list
-    for (auto& item: sorted_view) {
+    for (auto& item: sorted_view)
+    {
       int row = view.addRow ();
       view.set (row, 0, (item.first == ""
                           ? "(none)"
--- a/src/commands/CmdProjects.cpp
+++ b/src/commands/CmdProjects.cpp
@@ -112,29 +112,7 @@ int CmdProjects::execute (std::string& output)
 
     // create sorted list of table entries
     std::list <std::pair<std::string, int>> sorted_view;
-    for (auto& project : unique)
-    {
-      const std::vector <std::string> parents = extractParents (project.first);
-      if (parents.size ()) {
-        // if parents exist: store iterator position of last parent
-        std::list<std::pair<std::string, int>>::iterator parent_pos;
-        for (auto& parent : parents)
-        {
-          parent_pos = std::find_if (sorted_view.begin (), sorted_view.end (),
-              [&parent](const std::pair<std::string, int>& element) { return element.first == parent; }
-          );
-          // if parent does not exist yet: insert into sorted view
-          if (parent_pos == sorted_view.end()) {
-            sorted_view.insert (parent_pos, std::make_pair(parent, 1));
-          }
-        }
-        // insert new element below latest parent
-        sorted_view.insert (++parent_pos, project);
-      } else {
-        // if has no parents: simply push to end of list
-        sorted_view.push_back (project);
-      }
-    }
+    sort_projects (sorted_view, unique);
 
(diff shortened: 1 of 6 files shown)


tbabej:
This does not really solve the underlying issue, as there are other characters than `-` that will cause problems, i.e:

```
Project          Tasks
(none)               1
a                    2
a%b                  1
a+b                  1
a,b                  1
  b                  1
```

Additionally, the assumption that users won't use `=` in the project names might be too strong. Maybe what needs to be done here is that the entries in the list need to be properly sorted not alphabetically, but according to the tree structure.

mrossinek:
That is true..
I guess I made my life a little easy here.

What do you suggest? Should we sort them locally into an auxiliary data structure and add them to the viewing table afterwards or should we instead extend the functionality of the table to also `get` lines such that we could swap them in-place?

tbabej:
First option is what I had in mind. Second would also work, but you still don't get around having to construct a proper tree with sorted leafs somewhere.

mrossinek:
Sorry for the spam of commits... I did a rebase before continuing to work on my branch..

Well, I fixed the sorting now. At least for the `projects` command. If you are happy with this implementation I will do the same for the `summary` command and then this PR should be good to be merged.

mrossinek:
Okay I believe the rebase rendered this PR useless..
The number of changed files blew up to almost 400..

I will leave this open to finalize the discussion on the actual implementation and start a new PR once the bug is properly fixed. Sorry about that

tbabej:
We should be able to fix this branch. Try the following:

```
git checkout fix-1917
git remote add upstream git@github.com:GothenburgBitFactory/taskwarrior.git
git pull --rebase upstream 2.6.0
```

mrossinek:
@tbabej Ah thanks a lot! I rebased from `origin 2.6.0` which was the problem. Didn't think to rebase from `upstream`..
I hope force-pushing didn't break anything ^^

tbabej:
@mrossinek Pardon me if this is wrong (I had only a cursory glance), but maybe we could avoid code duplication between CmdProjects and CmdSummary by refactoring the sorting logic into a common function, implemented perhaps in the `sort.cpp` utility module?

mrossinek:
@tbabej Done! Do you have an idea, how we could circumvent the function overloading for the two different map types (one with string+int and one with string+bool). This is required for the two different commands using the sorting function. I have tried templating the function but can't seem to get it working properly. I believe that is due to having to template inside of the templated map... (if that makes clear what I mean by that)

mrossinek:
@tbabej There was another PR (#2073) also addressing this issue.. @grimmn took a much different (and simpler) approach, which I also really like. Maybe someone should also look into that? I believe he closed his PR after noticing this one received more attention.
Let me know, with which one of these we should proceed.

PR #2094, opened 2019-01-01
Multi-OS Travis-based CI service

tbabej:
Since [Flod2](https://central.tasktools.org) is not operational for about a year, development of Taskwarrior is hindered due to inability to test the funcionality across multiple platforms.

Instead of maintaining additional project and associated infrastructure, I've taken the approach of leveraging existing tools - namely Docker and Travis-CI to provide a replacement for Flod.

An example of a test build triggered after a commit [can be seen here](https://travis-ci.org/tbabej/taskwarrior/builds/474181465).


recollir:
Doing things based on Docker is a +1 from me. If we can get GitHub actions activated for the taskwarrior repo those containers could also run the tests as containers there and the final flod stage where the reports would compiled as well. Results could be hosted then via Github pages. No need for DO anymore.

tbabej:
Yes, I was thinking we could integrate here the final flod report generation stage and statically host it to get kind of presentation (which was imho superior to this). The `after_script` phase of the Travis CI job is meant to perform these tasks.

recollir:
👍

ddeimeke:
@tbabej Great job!! Thank you.

Just for the [bus factor](https://en.wikipedia.org/wiki/Bus_factor): Would it be possible to migrate everything to a GothenburgBitFactory-account?

tbabej:
@ddeimeke Not sure I follow - now that it's merged. Travis CI is enabled now also for the [GothenburgBitFactory account](https://travis-ci.org/GothenburgBitFactory/taskwarrior). I guess it is confusing because my fork also has CI enabled :)

ddeimeke:
@tbabej Please forget about me. I saw the screenshot with your name and not our team name.

PR #2095, opened 2019-01-02
TW-1511: Interpret string subtraction as concanetation with a dash

### LINKED ISSUE #1537: [TW-1511] Project titles not properly parsed if they contain hyphens
_Leon Feng on 2015-01-14T17:59:14Z says:_ 

 bq. The expression could not be evaluated.
Thus saith the warrior, when I enter 
`task pro:<any-project-title-with-hyphens-in-it>`.

The end.

(Related to TW-1479, I presume.)

tbabej:
Fixes #1537 (previously known as TW-1511).

### DIFF
--- a/src/Variant.cpp
+++ b/src/Variant.cpp
@@ -1278,7 +1278,18 @@ Variant& Variant::operator-= (const Variant& other)
     break;
 
   case type_string:
-    throw std::string (STRING_VARIANT_SUB_STRING);
+    switch (right._type)
+    {
+    case type_string:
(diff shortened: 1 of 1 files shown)


mrossinek:
I believe the test case in `variant_subtract.t.cpp` on line 134 needs updating now because this test number 32 fails:
```
> ./variant_subtract.t | grep not
not ok 32 - foo - foo --> error
```

tbabej:
@mrossinek Good catch - could you fire a PR for that?

Issue #1641, opened 2018-02-14
[TW-1617] Can't search for multi-word project


taskwarrior:
_Audrey Easterday on 2015-05-13T20:07:46Z says:_ 

 I create a project called 'Mike Visit 2015'.  

However, when I try to search for tasks in the project by typing -   

task project:'Mike Visit 2015' 
* 
 I get an error that says "The expression could not be evaluated".

I was going to ask this on the question board and not post it as a bug, but I can't seem to log in to the question board even though it will let me log in to this part of Jira just fine.

taskwarrior:
_Paul Beckingham on 2015-05-14T01:41:21Z says:_ 

 Thanks for the report.

Yes, the answers.tasktools.org site has lost it's authorization to get into the bug database system, and it's proving a pain to fix.  Apologies.

tbabej:
Regression in 2.5.2, 2.6.0.

tbabej:
Actually this is resolved on 2.6.0.

mrossinek:
Ah yes. This was mainly about the clarifying the syntax which happened in https://github.com/GothenburgBitFactory/taskwarrior/commit/d85901e7ff0ce0aec8e6090dd601552b08f61007, right?

Issue #1505, opened 2018-02-14
[TW-1479] Task arguments not properly parsed if they contain a space


taskwarrior:
_Tomas Babej on 2014-12-26T17:08:23Z says:_ 

 I have a task which description contains a space:

```
$ [tbabej] task 13 export
{"id":13,"description":"1.st task","entry":"20141226T165008Z","modified":"20141226T165008Z","status":"pending","uuid":"256721ea-6e02-4385-b39f-ea7b27592262","urgency":"0"}
```

This is 2.4.0 dev. version
```
$ [tbabej] task _version
e924913
$ [tbabej] task --version
2.4.0.beta3
```

When trying to query for the task by description, the following works:

```
$ [tbabej] task 1.st
[task next description ~ '1.st']

ID Age  Recur Description Urg 
13 9min       1.st task      0

1 task
There are local changes.  Sync required.
```

Also, the following works:

```
$ [tbabej] task description:1.st                                                                                                                                                                                                                    
[task next description = 1.st]

ID Age   Recur Description Urg 
13 10min       1.st task      0

1 task
There are local changes.  Sync required.
```

However, if I pass the space as well (by escaping it):
```
$ [tbabej] task description:1.st\ task
[task next description = 1.st task]
The expression could not be evaluated.
```

```
$ [tbabej] task description:"1.st task"                                                                                                                                                                                                                             
[task next description = 1.st task]
The expression could not be evaluated.
```

If I omit the description keyword, the exact match is no longer used, and we have a match:

```
$ [tbabej] task "1.st task"                                                                                                                                                                                                                                         
[task next description ~ '1.st task']

ID Age   Recur Description Urg 
13 10min       1.st task      0

1 task
There are local changes.  Sync required.
```

I think this is inconsistent and we should be able to query for the tasks even with description: prefix.

taskwarrior:
Migrated metadata:
```
Created: 2014-12-26T17:08:23Z
Modified: 2015-07-13T16:06:25Z
```

taskwarrior:
_Tomas Babej on 2014-12-26T17:09:05Z says:_ 

 Seems I can't create a issue without a issue :) this one should be 'Core' component.

taskwarrior:
_Paul Beckingham on 2014-12-27T16:25:41Z says:_ 

 These work:

```
$ task '/1.st task/'
$ task '1.st task'
$ task description == "'1.st task'"
```

taskwarrior:
[...]

Issue #1481, opened 2018-02-14
[TW-1455] Filter parser does not properly handle parentheses in attributes


taskwarrior:
_Tomas Babej on 2014-11-11T15:12:19Z says:_ 

 The TW cannot handle extra parentheses in filters. Minimal example is here:

```
[tbabej]$ task 'project:School)'
[task next project = School)]
Mismatched parentheses in expression
 
[tbabej]$ task 'project:School\)'
[task next project = School\)]
Mismatched parentheses in expression
 
[tbabej]$ task 'project:School\\)'
[task next project = School\\)]
Mismatched parentheses in expression
 
[tbabej]$ task 'project:School\\\)'
[task next project = School\\\)]
Mismatched parentheses in expression
 
[tbabej]$ task 'project:School\\\\)'
[task next project = School\\\\)]
Mismatched parentheses in expression
 
[tbabej]$ task 47 list
 
ID Age  D Project Description Urg
47 8min   School) test           1
 
1 task
```

taskwarrior:
Migrated metadata:
```
Created: 2014-11-11T15:12:19Z
Modified: 2015-07-13T15:57:42Z
```

taskwarrior:
_Paul Beckingham on 2014-11-29T21:46:07Z says:_ 

 I am not surprised.

PR #2097, opened 2019-01-02
Fix variant_subtract test case after #2095

mrossinek:
#### Description

Fix variant_subtract test case after merging of PR #2095 
Also fixes wait.t to expect correct output

```
Failed:                        
project.t                           2

Unexpected successes:          

Skipped:                       

Expected failures:             
```

### DIFF
--- a/test/project.t
+++ b/test/project.t
@@ -381,6 +381,7 @@ class TestBug1455(TestCase):
     def setUp(self):
         self.t = Task()
 
+    @unittest.expectedFailure
     def test_project_hierarchy_filter(self):
         """1455: Test project:school)
 
--- a/test/variant_subtract.t.cpp
+++ b/test/variant_subtract.t.cpp
@@ -139,9 +139,10 @@ int main (int, char**)
   try {Variant v32 = v3 - v2; t.fail ("foo - 3.14 --> error");}
   catch (...)                {t.pass ("foo - 3.14 --> error");}
 
-  // string - string   -> ERROR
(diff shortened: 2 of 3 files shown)


mrossinek:
Some comments regarding the still failing test cases:

* `version.t`: normal due to new commit ID
* `search.t`: I am not sure whether this test case still applies. According to the discussion in #1505 the 'is' modifier will be deprecated, is that still true? Using the search without the specific 'description' attribute filter results in the expected behavior.. Maybe this needs to be updated?
* `project.t`: the first failing case is to be expected according to #1481. Should this be marked as such? The second case regards #1641 which is closed but is actually not fixed (no variation works for me at least)
* `filter.t`: this requires some more attention imo. I might be able to have look at this once I finish my other PR...

tbabej:
Looks good to me, I'll let the CI confirm and merge :) As far as the other failing tests go, I agree that at least #1481 should be marked as expected and perhaps time should be never spent on fixing it. I wonder who submitted that issue..

As far as #1641 goes I managed to search for a multi word project yesterday using the following bash syntax:

`task project:\"three\ four\"`

tbabej:
Escaping the space makes the whole argument arrive as one list element in the `*argv` as far as the TaskWarrior is concerned - that most likely plays a role in how it interprets it.

Marking as expected fail is the correct move here. IMHO, to extract value out of CI, we should mark such cases as expected failures such that we can quickly validate new contributions. I think @pbeckingham needs to comment on what is the expected syntax here.

PR #2109, opened 2019-02-01
[Waiting] Drop support for Python 2.7

janikrabe:
#### Description

The EOL date for Python 2.7 has been set to January 1, 2020. This means that Python 2 will no longer be maintained after the end of this year.

This pull request consists of three commits.

- Update shebangs to explicitly use Python 3
- Drop backports and `ImportError` hacks for legacy Python support
- Replace the [deprecated](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex) `assert[Not]RegexpMatches`method with `assert[Not]Regexp`

Unfortunately, the deprecation warnings are currently making the output of the test suite almost impossible to read with Python 3.

#### Additional information

  There are unrelated failing tests.


janikrabe:
You're right. It looks like it still is, and I can't find any authoritative information or timeline from Apple regarding when (or if) this is going to change. It might be worth holding off on merging this until we know more (or at least until Python 2.7 is no longer maintained).

pbeckingham:
Python 2.7 is still the only Python installed on OSX by default, which is my primary platform.  Until this changes, I am not merging this, it would serve only to prevent me from running the unit tests.

JonasDralle:
###  [Instructions for installing Python3 on OSX](https://docs.python-guide.org/starting/install3/osx/)

Excerpt:
`
The version of Python that ships with OS X is great for learning, but it’s not good for development. The version shipped with OS X may be out of date from the official current Python release, which is considered the stable production version.
`

with homebrew you can install Python via `brew install python`.

pbeckingham:
Taskwarrior will support Python 2.7 and 3.0 until the need no longer exists.

It is immaterial whether Python 2.7 is "not good for development".  Taskwarrior is not developed in Python, it is simply tested using Python.

Issue #2081, opened 2018-12-10
Depends stack urgency


a7hybnj2:
If a task has multiple other tasks that depend on it the urgency should stack. Currently it does not and a task that has 1 or 2+ depends = the same urgency.

sampablokuper:
Depends on your config. Try adding this to your `.taskrc`:

```
urgency.inherit=on
urgency.blocked.coefficient        =  0.0
urgency.blocking.coefficient       =  0.1
urgency.uda.priority.H.coefficient =  1.0
urgency.uda.priority.M.coefficient =  0.0
urgency.uda.priority.L.coefficient = -1.0
urgency.age.coefficient            =  0.0
urgency.tags.coefficient           =  0.0
urgency.project.coefficient        =  0.0
urgency.annotations.coefficient    =  0.0
urgency.active.coefficient         =  0.5
```

a7hybnj2:
@sampablokuper That did not fix the issue:

```
task add parent00
task add parent01
task add child00
task add child01
task add child02
task 'parent00' modify depends 'child00','child01','child02'
task 'parent01' modify depends 'child00'
```
child 00, 01, 02 have same urgency even though child 00 has 2 tasks depend on it.

sampablokuper:
Hm. WFM.

Perhaps someone else can help, but you'll likely get more meaningful replies if you provide more info, e.g. the output of `task diagnostics`.

pbeckingham:
This is implemented.

a7hybnj2:
I feel like this didn't format very well but here it is.

```
~ ▶︎︎ sh .task/urgencySettings.sh 
-3.0
-5.0
-en   urgency.active.coefficient          =
-en   urgency.age.coefficient             =
-en   urgency.age.max                     =
-en   urgency.annotations.coefficient     =
-en   urgency.blocked.coefficient         =
-en   urgency.blocking.coefficient        =
-en   urgency.due.coefficient             =
-en   urgency.inherit                     =
-en   urgency.project.coefficient         =
-en   urgency.scheduled.coefficient       =
-en   urgency.tags.coefficient            =
-en   urgency.uda.priority.H.coefficient  =
-en   urgency.uda.priority.L.coefficient  =
-en   urgency.uda.priority.M.coefficient  =
-en   urgency.user.tag.next.coefficient   =
-en   urgency.waiting.coefficient         =
1.0
1.0
1.0
1.8
12.0
15.0
2.0
3.9
365
[...]

PR #2113, opened 2019-02-07
New feature: urgency accumulation

sampablokuper:
#### Description

Allow blocking tasks' urgency values to reflect the total urgency of the blocked tasks, rather than just the urgency of the most urgent among the blocked tasks.

N.B. There seems to be a bug (maybe a race condition of some sort) in the way TaskWarrior attempts to recursively identify blocking tasks and to get their urgency values. As a result of that bug, some "leaf" tasks (in graph-theoretic terminology) end up being assigned the same urgency as "branch" tasks. Until the bug is resolved, the functionality added in this commit will not work *optimally* (nor exactly as people are likely to expect); though it is still *useful*.

Fixes [#2081](https://github.com/GothenburgBitFactory/taskwarrior/issues/2081).

needsTest.

```
Failed:                        
calendar.t                          2

Unexpected successes:          

Skipped:                       
rx.t                                1

Expected failures:             
color.rules.t                       2
```

### DIFF
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -1159,6 +1159,17 @@ urgency.blocked.coefficient to 0.0 in order for this setting to
 be the most useful.
 .RE
 
+.B urgency.accumulate=off
+.RS
+Not actually a coefficient. When enabled, and if urgency.inherit is also
+enabled, blocking tasks inherit all of the urgency values of all of the tasks
+they block (calculated recursively), in addition to their own existing urgency
+value. I.e. the resulting urgency value is the sum of: the urgency the task
+would have if urgency inheritence were disabled, plus the urgency values of all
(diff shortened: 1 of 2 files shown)


pbeckingham:
Wim, there's a reason we didn't do this, can you recall?

wbsch:
Yes, I do - mostly because I've outlined the reasons in the ticket @sampablokuper linked here, https://github.com/GothenburgBitFactory/taskwarrior/issues/2081#issuecomment-450728997

This pull request might still be a viable config option even if I personally think that it's the wrong approach to task management, but I'm very interested to hear what those "race conditions" mentioned are.

Xerus2000:
I have left another comment on that issue outlining ideas that might re-conciliate both sides

tbabej:
Given the original author deleted their Github account, we'll need somebody else to take this over. That would require a new PR.

PR #2114, opened 2019-02-08
task UI: prevent ID/UID being followed by period.

corbolais:
This enables copy and paste in case phrase collation class contains the dot character.

Signed-off-by: <corbolais@gmail.com>

#### Description

When working with tw I usually cnp the task ID or UID.

Task commands print the ID often followed by a period ".". Usually the dot is part of the collation char class for mouse selection by double-clicking, however. This means, not only the ID/UID is being marked and copied but the following "." is also. This regularly yields unecessary clutter requiring explicit clean up by hand.

### DIFF
--- a/src/commands/CmdAdd.cpp
+++ b/src/commands/CmdAdd.cpp
@@ -64,20 +64,20 @@ int CmdAdd::execute (std::string& output)
 
   if (Context::getContext ().verbose ("new-uuid") &&
            status != Task::recurring)
-    output += format ("Created task {1}.\n", task.get ("uuid"));
+    output += format ("Created task UID \"{1}\".\n", task.get ("uuid"));
 
   else if (Context::getContext ().verbose ("new-uuid") &&
            status == Task::recurring)
-    output += format ("Created task {1} (recurrence template).\n", task.get ("uuid"));
+    output += format ("Created task UID \"{1}\" (recurrence template).\n", task.get ("uuid"));
 
   else if (Context::getContext ().verbose ("new-id") &&
       (status == Task::pending ||
        status == Task::waiting))
-    output += format ("Created task {1}.\n", task.id);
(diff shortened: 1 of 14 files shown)


corbolais:
the PR compiles locally.
what does not compile locally are the tests.
why does travis-ci fail?

corbolais:
This is a CI failure:

!!! Multiple package instances within a single package slot have been pulled
!!! into the dependency graph, resulting in a slot conflict:
dev-lang/perl:0
  (dev-lang/perl-5.26.2:0/5.26::gentoo, ebuild scheduled for merge) pulled in by
    =dev-lang/perl-5.26* required by (virtual/perl-Parse-CPAN-Meta-2.150.10-r1:0/0::gentoo, ebuild scheduled for merge)
    ^              ^^^^^                                                                                                
    (and 22 more with the same problem)
  (dev-lang/perl-5.24.4:0/5.24::gentoo, ebuild scheduled for merge) pulled in by
    dev-lang/perl:0/5.24= required by (virtual/perl-CPAN-Meta-2.150.5-r1:0/0::gentoo, installed)
                 ^^^^^^^^                                                                        
    =dev-lang/perl-5.24* required by (virtual/perl-CPAN-Meta-2.150.5-r1:0/0::gentoo, installed)
    ^              ^^^^^

pbeckingham:
This is a change in Taskwarrior behavior so that the user can avoid modifying xterm click selection characters.  Why not simply change xterm settings instead?

corbolais:
Hi,
Here, I use xterm, konsole uxterm et al. It would require reconfiguration of working selection character sets. This in case would break with specific setups.
Is there any kind of dependency that forces TW to keep the status quo regarding this PR?

pbeckingham:
It's just on principle: Taskwarrior does not need to change because a user refuses to configure a third-party tool.

PR #2117, opened 2019-02-17
Workaround for alias grep=rg in fish

Pastafarianist:
#### Description

If the user has `alias grep rg` in Fish, completions are going to fail as the meaning of `-E` option is different in `grep` and `rg`. This ensures that completions bypass aliasing.

```
$ cd test && ./problems
Traceback (most recent call last):
  File "./problems", line 60, in <module>
    with open(cmd_args.tapfile) as fh:
FileNotFoundError: [Errno 2] No such file or directory: 'all.log'
```

Sorry, the tests seem to be broken.

### DIFF
--- a/scripts/fish/task.fish
+++ b/scripts/fish/task.fish
@@ -186,7 +186,7 @@ end
 
 function __fish.task.list.command
   # ignore special commands
-  __fish.task.list._command $argv | grep -Ev '^_'
+  __fish.task.list._command $argv | command grep -Ev '^_'
 end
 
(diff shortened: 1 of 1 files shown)


pbeckingham:
Nice work.

pbeckingham:
Dammit, this is on the master branch, and I didn't notice.  Reverting.

pbeckingham:
Please put this change on the 2.6.0 DEVELOPMENT branch, not master.

PR #2118, opened 2019-02-24
Fix #2002

### LINKED ISSUE #2002: Add more hints on how to reset the current context
When I need to look at the tasks outside of the project, I have to reset/change the current context. As I do this not so frequently and I'm relatively new user, I can't remember how exactly to reset the context.

**What command(s) did you run?**
- `task context reset`
- `task context default`
- `task context help`
- `task context -h`

**What did you expect to happen?**
`task context` should give me advice on how to remove the current context, like `task context <name>` does.

So, I suggest to modify an output of the `task context` to also show the following:
```
Use 'task context none' to remove active context.
```

mrossinek:
#### Description

Simply add the suggestion from #2002 

I did not implement another 'task context help' subcommand since no taskwarrior command has this. Simply calling 'task context' will give the hint on how to unset the context.

### DIFF
--- a/src/commands/CmdContext.cpp
+++ b/src/commands/CmdContext.cpp
@@ -73,7 +73,7 @@ int CmdContext::execute (std::string& output)
   else
   {
     listContexts (out);
-    out << "Use 'task context none' to unset the current context.";
+    out << "Use 'task context none' to unset the current context.\n";
   }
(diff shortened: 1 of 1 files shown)


pbeckingham:
Nice work.

PR #2116, opened 2019-02-14
[WIP] Fix rc.verbose documentation in taskrc(5)

janikrabe:
This commit fixes what I believe to be two mistakes in `taskrc(5)`, specifically, the documentation of the `verbose` setting.

I would appreciate if someone more familiar with this setting could review these changes.

PR #2122, opened 2019-02-27
Fix rc.verbose default value in taskrc(5)

janikrabe:
As promised in #2116, this reverts one of the changes I made in f6b2a6541c462eab1e89c3bbcb4610eca84ec295.

It looks like `verbose=1` was actually the correct default; I had been using an older version. Sorry!


pbeckingham:
Thanks.

PR #2152, opened 2019-04-27
fix stale reference to empty priority color config

thehunmonkgroup:
#### Description

color.pri.none is now color.uda.priority.none

I have tested the above change with TaskWarrior 2.5.1, and it works correctly. The old 'color.pri.none' option does not work.

Not applicable.


pbeckingham:
Thank you.

thehunmonkgroup:
I think this can also be merged w/ the 2.5.2 branch in prep for the next release?

pbeckingham:
That is correct.  As I slowly fix all the broken tests, I’ll also be cherry-picking from 2.6.0 the fixes that can and should be released.  I’ve already done a lot of that.

Paul

PR #2154, opened 2019-04-28
Make completed projects appear in autocompletion

heiderich:
#### Description

The aim of this pull request is to make completed projects appear in the bash auto completion.

```
Failed:                        
dates.t                             1

Unexpected successes:          

Skipped:                       
rx.t                                1

Expected failures:             
color.rules.t                       2

```

### DIFF
--- a/scripts/bash/task.sh
+++ b/scripts/bash/task.sh
@@ -65,7 +65,7 @@ _task_offer_priorities() {
 }
 
 _task_offer_projects() {
-    COMPREPLY=( $(compgen -W "$($taskcommand _projects)" -- ${cur/*:/}) )
+    COMPREPLY=( $(compgen -W "$($taskcommand _unique project)" -- ${cur/*:/}) )
 }
 
 _task_offer_contexts() {


pbeckingham:
This change breaks functionality.  There is a configuration setting `list.all.projects`, obeyed by the `_projects` command that already does this. This change eliminates control.

heiderich:
> This change breaks functionality. There is a configuration setting list.all.projects, obeyed by the _projects command that already does this.

I was unaware of this configuration option. Thanks for letting me know.

There are probably no "typical" use cases. Personally I would prefer "task projects" to print only the active projects. (I mostly use "task projects" to get an overview about my projects requiring action.) Therefore I would prefer the default setting `list.all.projects=no`.

I however think that bash auto completion is a different use case and I am not sure whether the same configuration option should be applied in both situations. I often find myself in need to refer to projects that do not have active tasks anymore. This pull request would allow this and it would not change the behavior of the `_projects` and `projects` commands themselves.

> This change eliminates control.

You are right. It eliminates control over the behavior of the auto completion. At the other side it allows the auto completion to behave differently as the `_projects` command (admittedly with less control).

Would you consider adding a feature that would them to behave differently?

pbeckingham:
Yes, I'm happy with letting the user control this.  But it doesn't require such a big change to the completion script.  All it requires is the original command with `rc.list.all.projects=1` inserted. The idea being that smaller changes are often better.

That said, this does change behavior, by default, of bash completion. It's been an important rule I follow that it's wrong to make changes based on the wishes of one person, no offence intended. We have seen somewhere north of 10^5 tarball downloads from our site, and who knows how many package installs there have been. The point being that it's greater than one, so this is by definition a minority opinion, unless there is support for this change out there.

The best approach is probably to document how to change the bash completion script to achieve both forms of behavior, but leave the behavior as it is. I would argue that project name completion is used significantly more often on pending tasks than completed tasks, which is why the default is what it is.

heiderich:
> Yes, I'm happy with letting the user control this. But it doesn't require such a big change to the completion script. All it requires is the original command with rc.list.all.projects=1 inserted. The idea being that smaller changes are often better.

My motivation for using `_unique project` was the information from the man page that `_projects` is deprecated in favor of `_unique`. What is the difference between

`task _unique project`

and

`task _projects rc.list.all.projects=1`?


> That said, this does change behavior, by default, of bash completion. It's been an important rule I follow that it's wrong to make changes based on the wishes of one person, no offence intended. We have seen somewhere north of 10^5 tarball downloads from our site, and who knows how many package installs there have been. The point being that it's greater than one, so this is by definition a minority opinion, unless there is support for this change out there.

I understand. I was probably to quick opening this PR.

PR #2171, opened 2019-05-19
Completely refine zsh completion for 2.5.2

doronbehar:
#### Description

This is a complete refinement to the ZSH completion, including:

- No environment leakage - A.K.A global variables.
- All filter attributes are computed on demand.
  * This is important in the common case where a project is being added when a new task is added and you want `task project:` to suggest the new project. Up until now, the completion function completed only projects it knew of from the first time the completion function was loaded.
- Added support for UDAs and their labels in addition to normal attributes.
- Removed the now unnecessary hard coded priority completion since this is supported by the UDAs completion.

### DIFF
--- a/scripts/zsh/_task
+++ b/scripts/zsh/_task
@@ -65,12 +65,6 @@ _task_zshcmds=( ${(f)"$(task _zshcommands)"} sentinel:sentinel:sentinel )
 
 _task_aliases=($(task _aliases))
 
-_task() {
-    _arguments -s -S \
-        "*::task default:_task_default"
-    return 0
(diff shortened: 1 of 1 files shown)


doronbehar:
Thanks! I hope the CLI syntax won't change much in `2.6.0`. If so, please let me know by mentioning me in a new issue. I'll be happy to take over it when the time comes.

pbeckingham:
Thank you!  I expect only minimal changes.  2.6.0 will likely also be called 3.0.0.  I will notify you, thanks for the offer.

Issue #2023, opened 2018-05-13
Layout breaks if report does not fit in the available width


ad-si:
I'm getting the warning `The report has a minimum width of 225 and does not fit in the available width of 211.` when running `task all` because of a really long line:

<img width="422" alt="screen shot 2018-05-13 at 07 23 52" src="https://user-images.githubusercontent.com/36796532/39964925-a3fc7580-567e-11e8-92cc-7d3e23ff3b96.png">

The native wrapping of the the terminal then unfortunately breaks the whole layout, as every line which was just one line height before is now two lines high:

<img width="591" alt="screen shot 2018-05-13 at 07 27 33" src="https://user-images.githubusercontent.com/36796532/39964950-31e1d43a-567f-11e8-8256-f5617363a287.png">

The problem is that every line is padded to same length as the longest line.
I guess the correct fix for this would be to support native wrapping in taskwarrior, so that the description line gets wrapped properly. However, I don't know how time-consuming this will be.
Maybe as a quick-fix there could be a config flag to not pad lines with spaces?
This will make the highlighted lines look jagged, but still better than this awful wrapping.

What do you think?

pbeckingham:
It's a bug alright.  There is support for hyphenation, but somewhere along the line, it broke.  Hyphenation is supposed to only kick in when this situation happens.  Thanks for reporting.

ad-si:
So if this is a bug, when is the "… does not fit in the available width" warning then actually necessary?

chaseadam:
looks like proper hyphenation support has been in backlog for a while: #74

tbabej:
This is now fixed on our latest releases. Tested on 2.5.2, 2.5.3, 2.6.0.

```
[warrior@tw-v2.5.2 ~]$ task add test
Created task 1.
[warrior@tw-v2.5.2 ~]$ task 1 ann https://github.com/GothenburgBitFactory/taskwarrior/issues/2023/liebherr.com/shared/media/aerospace-and-transportation/aerospace/downloads/magazines/aets-magazines-recent/liebherr-aerospace-magazine-2016-de.pdf
Annotating task 1 'test'.
Annotated 1 task.
[warrior@tw-v2.5.2 ~]$ task
[task next]

ID Age  Description                                                                                                       Urg 
 1 4s   test                                                                                                               0.8
          2021-06-19
[...]

PR #2223, opened 2019-10-11
Poor implementation of linewrap restored to task

0xACE:
This addresses #2023

For many years linewrap has been broken in taskwarrior, rendering the
application basically useless in terminals with short widths and tasks
containing URLs in them... This janky line wrap patch sort of works but
introduces other minor graphical glitches.

But seeing as no one took the time to fix it over the years, here some a
small salvation to those who suffer as much as me.

One problem of this patch is, when a line is wrapped, it may be 1 column
smaller than the terminal width, resulting in each line not fully
utilizing the terminal width.

This patch is not polished and could probably need to be tended by
someone who cares more.

This patch has been tested on v2.5.2

### DIFF
--- a/src/ViewTask.cpp
+++ b/src/ViewTask.cpp
@@ -208,7 +208,6 @@ std::string ViewTask::render (std::vector <Task>& data, std::vector <int>& seque
   // Not enough for minimum.
   else if (overage < 0)
   {
-    context.error (format (STRING_VIEW_TOO_SMALL, sum_minimal + all_extra, _width));
     widths = minimal;
   }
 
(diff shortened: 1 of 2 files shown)


liskin:
Is it okay that this was merged to master but not to 2.5.2 nor 2.6.0?

liskin:
See also https://github.com/GothenburgBitFactory/taskwarrior/pull/2280#issuecomment-654676538.

0xACE:
> Is it okay that this was merged to master but not to 2.5.2 nor 2.6.0?

Iirc the reason why the PR was made the way it is:

I originally wrote the patch for `2.6.0`, but seeing as `2.6.0` isn't labeled "published" as stable, I backported it for `2.5.1` so it could be used today, because it is a ongoing problem for existing users.

I could submit the patch for version `2.6.0` if it is necessary (I'm not sure if the merge between `master` and `2.6.0` will overwrite my changes)

Cheers

PR #2228, opened 2019-11-08
Correctly parse projects with spaces in zsh completion

mcornella:
#### Description

This is the same as how the `_task_zshids` array is created.
See https://github.com/ohmyzsh/ohmyzsh/issues/8249

This PR doesn't need to run the test suite. I have tested it manually with the following steps:

1. Use zsh with the completion file loaded
2. Create the task projects: `task add TEST project:"Test Project"`
3. Try to complete: `task project:[TAB]`

### DIFF
--- a/scripts/zsh/_task
+++ b/scripts/zsh/_task
@@ -24,7 +24,7 @@
 # https://www.opensource.org/licenses/mit-license.php
 #
 typeset -g _task_cmds _task_projects _task_tags _task_config _task_modifiers
-_task_projects=($(task _projects))
+_task_projects=(${(f)"$(task _projects)"})
 _task_tags=($(task _tags))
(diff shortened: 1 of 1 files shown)


mcornella:
The Travis CI build failed but not due to the PR:
```
ERROR: Service 'test-gentoo' failed to build: COPY failed: stat /var/lib/docker/overlay2/68b9bf8571acbff0ad3957a321de5646d0bf1e56fdc4026d80b229bdbd408847/merged/usr/portage: no such file or directory
The command "if [[ $CONTAINER != "osx" ]]; then docker-compose build test-$CONTAINER ; fi" failed and exited with 1 during .
```
This is a simple fix that should be fast-tracked.

PR #2234, opened 2020-01-12
Update JP holidays 2020

tknv:
Update JP holidays for 2020.


tbabej:
Hi @tknv, thanks for this contribution! Maybe we should make it clear in the README, but the holidata files are actually autogenerated from our holidata project, which contains definitions for each country's holiday logic, such that we can generate the holiday files for each year automatically.

See the [following example for Canada](https://github.com/GothenburgBitFactory/holidata/blob/master/holidata/holidays/en-CA.py). I can see [we currently don't have Japanese holiday definition file](https://github.com/GothenburgBitFactory/holidata/tree/master/holidata/holidays). Would you consider contributing that instead? That way we would need to only update the underlying definition file if the logic behind the holidays changes.

PR #2237, opened 2020-01-27
Fix add.t Test1549 test failure.

vrusinov:
The test was failing on my machine on fresh checkout. It failed because
somehow both new-id and new-uuid verbosity were set during the test, so
TW defaulted to printing out uuid, which did not match the assertion.

Test fixed by explicitly setting verbosity to new-id.

#### Description

Replace this text with a description of the PR.


pbeckingham:
Thank you.

PR #2305, opened 2020-07-07
Test: Add UUID test demonstrating a "Cannot subtract strings" bug

liskin:
```
$ task 0000000d-000a-0000-0000-000000000000 export
Cannot subtract strings
$ task 0000000d-0000-0000-0000-000000000000 export
The expression could not be evaluated.
```

### DIFF
--- a/test/lexer.t.cpp
+++ b/test/lexer.t.cpp
@@ -421,6 +421,7 @@ int main (int, char**)
 
     // UUID
     { "ffffffff-ffff-ffff-ffff-ffffffffffff",         { { "ffffffff-ffff-ffff-ffff-ffffffffffff",         Lexer::Type::uuid         }, NO, NO, NO, NO }, },
+    { "0000000d-0000-0000-0000-000000000000",         { { "0000000d-0000-0000-0000-000000000000",         Lexer::Type::uuid         }, NO, NO, NO, NO }, },
     { "00000000-0000-0000-0000-0000000",              { { "00000000-0000-0000-0000-0000000",              Lexer::Type::uuid         }, NO, NO, NO, NO }, },
     { "00000000-0000-0000-0000",                      { { "00000000-0000-0000-0000",                      Lexer::Type::uuid         }, NO, NO, NO, NO }, },
     { "00000000-0000-0000",                           { { "00000000-0000-0000",                           Lexer::Type::uuid         }, NO, NO, NO, NO }, },


pbeckingham:
Please mark this PR as not being "WIP" and I will merge.

liskin:
@pbeckingham but this is not a real PR to be merged, it's just a test demonstrating #2306. Until someone fixes the issue and makes the test not fail, this shouldn't be merged.

Btw now that you're working on taskwarrior again, I think I added some comments somewhere about several PRs being merged to master but not to 2.5.2 nor 2.6.0. Would be nice to cherry-pick those into at least 2.6.0, and then ideally mark 2.6.0 as the default branch and delete master, so that this mistake doesn't happen again.

Thanks!

tbabej:
@liskin The policy for the development branches of TW is to actually allow failed tests (and PRs contributing just failing tests were encouraged along with the bug reports). This goes way back when the project did not have a test suite that ran automatically on PRs.

tbabej:
@liskin Good point about the default branch, it actually is our default now :slightly_smiling_face:

liskin:
Oh, I see, okay. :-)

mrossinek:
> This goes way back when the project did not have a test suite that ran automatically on PRs.

I can see why this used to be the policy but I think that now that we are actually trying to use the CI it would make more sense to integrate such unittests while being flagged as unexpected failures. Then we can actually use the CI to detect whether some change breaks something and the expected failure flag can be removed once the bug is fixed.

What do you think? @tbabej @pbeckingham

mrossinek:
I don't fully understand all of the details which this PR entails. It resulted in 4 new failures:
```sh
% ./lexer.t | grep -C 4 'not ok'
ok 903 - Embedded Lexer::token(...) --> true
ok 904 -   token --> ffffffff-ffff-ffff-ffff-ffffffffffff
ok 905 -   type --> Lexer::Type::uuid
ok 906 - Isolated Lexer::token(...) --> true
not ok 907 -   token --> 0000000d
# expected: '0000000d-0000-0000-0000-000000000000'
#      got: '0000000d'
not ok 908 -   type --> Lexer::Type::duration
# expected: 0
#      got: 17
ok 909 - Embedded Lexer::token(...) --> true
not ok 910 -   token --> 0000000d
# expected: '0000000d-0000-0000-0000-000000000000'
#      got: '0000000d'
not ok 911 -   type --> Lexer::Type::duration
# expected: 0
#      got: 17
ok 912 - Isolated Lexer::token(...) --> true
ok 913 -   token --> 00000000-0000-0000-0000-0000000
```
Why are there 2 new failures designating a `duration` type?

liskin:
> Why are there 2 new failures designating a `duration` type?

Because 0000000d-0000-0000-0000-000000000000 is interepreted as "0 days - 0 - 0 - 0 - 0" instead of as an UUID.

Please do feel free to flag this as an expected failure with a comment. I never considered this would be merged as is, therefore I did not add any comments, and instead described the issue in #2306.

mrossinek:
I am honestly not sure how to mark this case as an expected failure. The lexer test cases are being compiled from `lexer.t.cpp`.
Do @tbabej or @pbeckingham know whether this is even supported?

tbabej:
Maybe we can make that a separate test :thinking:

mrossinek:
I can take a look at that although I am not sure whether that requires more effort than "just" fixing it :D

tbabej:
@liskin @mrossinek I implemented a mechanism to mark C++ tests as expected failures, see efc56d3ab6e676af17f8414ef8f1dc70649758c3. Then I used it to mark this UUID tests as currently expected failure in 9e9ef6a210c48f4c9c220420b5c2a82719aa0565.

Issue #2345, opened 2020-11-21
Tests currently failing on active development branches (2.5.2, 2.6.0)


tbabej:
Running `make test` shows that there are currently couple of tests broken on the 2.6.0 branch:

```
Failed:                        
filter.t                            5
project.t                           1
search.t                            1
undo.t                              2
```

To maximize utility of the CI, we should strive to get these fixed.

mrossinek:
I started to look into the other failing tests again. I also remembered the discussion which we had a while back over at #2097.

- [ ] `filter.t`: will update later
- [ ] `project.t`: this we had already discussed. We wanted to get @pbeckingham's input on what the expected syntax is for the user. We determined that `task project:\"three\ four\" list` works as expected in bash and zsh as it escapes the argument correctly. In order to reflect this in the test case we need to change it to be `r'project:\"three\ four\" list'` where we need to make it a raw string in order to ensure that Python does not interpret the backslashes itself. (Alternatively we could write `\\\"` but I find the raw string much more legible and clear). We should probably also document the requirements for this syntax.
- [ ] `search.t`: this essentially boils down to the exact same problem as in the project filter above: no clarity on the syntax when dealing with spaces. If we clarify it above, we can apply the same fix here.
- [ ] `undo.t`: This looks currently very broken. I am bisecting it as I write this.

mrossinek:
I don't think it makes much sense to open another issue for the discussion of the tests which fail on 2.5.2 branch. So I will add them here:
```
% ./problems
Failed:
diag.t                              1
filter.t                            5
project.t                           3
search.t                            1
version.t                           1
wait.t                              1
```

- `diag.t`: TODO
- `filter.t`: the same issues as on 2.6.0 branch. Detailed discussion in next comment.
- `project.t`: one overlap with 2.6.0 branch. TODO
- `search.t`: same issue as 2.6.0 branch (see previous comment)
- `version.t` can probably be addressed by cherry-picking #2343 on to the 2.5.2 branch
- `wait.t`: TODO

tbabej:
Yeah, my bad for the title being too specific. Fixed now :slightly_smiling_face:

mrossinek:
Discussion of the `filter.t` failures:
```
% ./filter.t | grep 'not ok'
not ok 21 - filter.t: filter - description contains +
not ok 24 - filter.t: 1630: Verify that 'due.before:9d' is correctly interpreted
[...]

PR #2343, opened 2020-11-21
Fix test suite for Python 3.9

tbabej:
This upgrades the test runner to support Python 3.9, mostly by replacing deprecated methods with their new equivalents.

Also updates the Copyright across the repo to 2020, thus fixing one more `version.t` test :slightly_smiling_face:

PR #2346, opened 2020-11-21
Fix failing tests on 2.6.0 branch

mrossinek:
#### Description

This PR attempts to fix some more of the failing tests on the 2.6.0 branch.

This is part of the discussion in #2345:
- ~`filter.t`:~ will be addressed on 2.5.2 branch
- ~`project.t`:~ will be addressed on 2.5.2 branch
- ~`search.t`:~ will be addressed on 2.5.2 branch
- [x] `undo.t`: After bisecting I found this commit to be the one which introduced the problem: https://github.com/GothenburgBitFactory/taskwarrior/commit/48dff115c2d237e0c94b97a868d81360358c460e. The problem occured from an unsafe creation of a `Task` object because the `prior` string can be empty.
- [x] `version.t`: Addressed in #2343


pbeckingham:
Excellent. Can you move this ver to hte 2.5.2 branch, which is the one going out first, and in most need?

mrossinek:
Ah yes that would make sense. I noticed that #2343 is addressed at branch 2.6.0 which is why I did the same for this PR.
I will open a new PR for branch 2.5.2 as the Undo command is not broken there. But the following failures are observed on that branch:
```
Failed:
project.t                           3
```

I will see about addressing the individual failures and also see whether we can backport some fixes from the 2.6.0 branch (as some appear to be fixed there).
We can definitely apply #2343 on the 2.5.2 branch, though!

pbeckingham:
This:
    Task old = Task ();

Should be:
    Task old;

mrossinek:
I fixed the remaining issues by clarifying the need for escaping filters correctly (note in man-page as well as fixing the unittests).
I also marked all remaining issues as expected failures. They should be fixed by closing the respective issues.

However, PR #2305 added some new unittests which also fail. We should clarify some comments over there and afterwards I believe that this PR is ready to go.

tbabej:
This is a great start, I'll try to pull the last strings together.

Issue #2344, opened 2020-11-21
CI matrix needs to be updated to support newer distro releases


tbabej:
Our Travis-based CI is relatively exhaustive, but contains outdated versions of popular distros (i.e. Fedora 28, 29, CentOS 7, Ubuntu 18.04 or Arch 2018). All of these have newer releases and some of the currently tested releases are no longer supported.

We should update the test matrix and the associated Dockerfiles to be based on a set of current distro releases.

kwoot:
How can I help?

8-bit-fox:
They reside in [`test/docker`](https://github.com/GothenburgBitFactory/taskwarrior/tree/2.5.2/test/docker)

tbabej:
@kwoot For most of the Dockerfiles, I think this comes down to:
1. updating the base image to more recent tag (i.e. from `fedora:29` to `fedora:33`) and
2. checking if the build still works (dependency package names could have changed, etc.)

As for what distros to test, I think we should keep testing the ones that are still supported, so that would mean:
* Replacing F28 and F29 with F32 and F33 (where F=Fedora)
* Keeping Centos 7, but adding Centos 8
* Updating Arch to a newer tag (maybe just a generic one like @8-bit-fox suggests)
* Replacing Ubuntu 16.04 with Ubuntu 20.04, keeping Ubuntu 18.04
... and similar for the other distros (according to their versions and support schedules)

8-bit-fox:
@tbabej, as I have pointed out in #2339 I think the Docker images themselves would need some refactoring as well. Currently, they contain state (the actual project-state), which requires to build the whole OS over and over again. Instead the project should be mounted as volume, which would remove the need for rebuilding the image. Except for important changes in the underlying OS, of course.

IMO, OS changes should also be tested against, but in a separate pipeline. This allows to distinguish between issues introduced by changes on the project and changes introduced by the underlying system.

For sure this can be done in multiple steps (1. Fixing current pipeline, 2. Improving CI).

sruffell:
> @tbabej, as I have pointed out in #2339 I think the Docker images themselves would need some refactoring as well. Currently, they contain state (the actual project-state), which requires to build the whole OS over and over again. Instead the project should be mounted as volume, which would remove the need for rebuilding the image. Except for important changes in the underlying OS, of course.

I'll second this point. I too believe the Docker images should be refactored so that they do not need to be rebuilt in order to to run the tests against the source in the current directory.

[...]

Issue #2339, opened 2020-11-18
Discussion: Surveying Continuous Integration providers


8-bit-fox:
Mentioned as part of #2338 , @pbeckingham [mentioned](https://github.com/GothenburgBitFactory/taskwarrior/issues/2338#issuecomment-729067633) that the project has lost its testing infrastructure.

The project is now looking for a new infrastructure setup. Requirements are apparently
* multi-platform testing, [even among linuxes](https://github.com/GothenburgBitFactory/taskwarrior/tree/2.6.0/test/docker)
* Automated builds of PRs
* Free of charge

This issue aims at collecting options to revive the continuous testing for taskwarrior.

Collected options:

- [ ] [SuSE OBS](https://build.opensuse.org/)
  - Linux based only
  - Packaging mechanism included
  - No dedicated testing pipelines (must be part of the building/packaging)
- [ ] Github Actions
  - Only three hosted-VMs are available (Win10, MacOS, Ubuntu) in various versions
  - Self-hosted VMs likely produce additional cost and maintenance effort
  - [Docker container actions](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-docker-container-action) allow for running arbitrary Docker images, which are built during execution time.
- [ ] [TeamCity](https://www.jetbrains.com/teamcity/)
  - Free of charge for Open Source project
  - Project needs to meet JBs criteria (OSI + regular commits within last three months)
  - Closed-source vendor
- [ ] [AWS CodeBuild](https://aws.amazon.com/en/codebuild/)
  - Usage based pricing (per minute payment)
  - 100 free minutes included on the smallest node (2 vCPU + 3 Gi RAM)
- [ ] [Azure Devops Services](https://azure.microsoft.com/en-us/services/devops/)
- [ ] [Gitlab](https://about.gitlab.com/solutions/github/)
  - Can be integrated with Github for free for OpenSource projects
  - Offer only limited? 
    > So, for the next year we are making the GitLab CI/CD for GitHub feature a part of our GitLab.com Free tier. `
- [ ] [Travis CI](https://travis-ci.org/)
  - Ubuntu or Windows as Host-System
  -  Docker available only for linux
  - Already in use
- [ ] Cirrus CI

kwoot:
May I suggest SuSE OBS? I am no OBS expert but I think it offers what is needed. Even building on non-SuSE distros and IBM mainframes :-).

8-bit-fox:
[...]

Issue #2338, opened 2020-11-17
O, Project Activity Where Art Thou?


corbolais:
This issue likes to bundle all the past, present and future issues regarding the projects past, present and future.

You feel your PRs are not acknowledged, though they got reviewed but are pending a merge since..?
-> This is the issue to head for and lament about it.

There might be a multitude of other related issues or feelings deserving mention here. Just Do It!

Issue #2248 was not the first one getting hijacked because of lack of this repo's inactivity. It just was it's starter: https://github.com/GothenburgBitFactory/taskwarrior/pull/2248#issuecomment-727277557

As I'm grateful for TW's existance and for the PO's work, a fork is kind of a last resort. Nevertheless, I'd be willing to undertake the initial step.

@jrabinow mentioned an article presenting an interesting concept or procedure to possibly find a way to unstall a project: [The Pull Request Hack](https://felixge.de/2013/03/11/the-pull-request-hack/).

Let's see how this all turns out -- for the better, please.

Edit: For starters, how about >= 380 open issues?

corbolais:
Skimming through the first couple of issue pages, there are already many issue regarding the project activity, documentation issues, issues issues etc: #2311, #2147, #2259, ...

The sheer number of issues about existing functionality, requests for enhancements, improvements and plain old bug reports is **HUGE**. And all too often gray-bearded.

TW is a great piece of SW, attracting alot of apt people on the way. All too often turning them away for (no) good reason. I'd love to see this project rise from it's issues. ;-)

kwoot:
When I look at the log I see that the last commit was:  #4b67f10cb13017 That one is dated Fri Jan 17 19:16:57 2020 -0600. So activity may be low, but the project is not dead yet (I've seen worse). Let's try to contact the maintainer first.

pbeckingham:
Project is not dead, it just looks like it. Working with Tomâš to get this going.

corbolais:
If you look closer: 14 pending PRs, two projects, last updated 2018/2019.

Although @pbeckingham as PO is far from being the only dev (has support from six other devs), his activity in general was last higher in [2016](https://github.com/pbeckingham?tab=overview&from=2016-12-01&to=2016-12-31). The profile tells you "I may be slow to respond.".

[...]

PR #2347, opened 2020-11-21
Remove project content from test images

8-bit-fox:
#### Description
This PR aims at removing the project code from the test images. Reasons have been explained in #2344 .

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -1,0 +1,10 @@
+version: '3'
+services:
+  tw-build-and-test:
+    image: ghcr.io/8-bit-fox/tw-arch:base
+    container_name: tw-build-and-test
+    entrypoint: /src/entrypoint.sh
+    working_dir: /src
+    volumes:
+    - ./:/src
+
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+git submodule init
+git submodule update
+cmake -DCMAKE_BUILD_TYPE=debug .
+make -j2
+make install
+task --version
+
+cd test || exit
+make
+./run_all -v
+./problems
+make clean
+
+cd ../
+make clean
--- a/test/test-containers/Dockerfile-arch
+++ b/test/test-containers/Dockerfile-arch
@@ -1,0 +1,12 @@
+FROM archlinux:base
+
+RUN pacman -Sy --noconfirm archlinux-keyring &\
+    pacman -Syyu --noconfirm &\
+    pacman -S --noconfirm gnutls util-linux bash-completion cmake python git libfaketime &\
+    pacman -S --noconfirm make gcc
+
+# Setup language environment
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+
+CMD ["/bin/bash"]
--- a/test/test-containers/Dockerfile-centos7
+++ b/test/test-containers/Dockerfile-centos7
@@ -1,0 +1,28 @@
+FROM centos/devtoolset-7-toolchain-centos7:7
+
+USER root
+RUN yum update --assumeyes
+RUN yum install --assumeyes https://repo.ius.io/ius-release-el7.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm &\
+    yum install --assumeyes centos-release-scl &\
+    yum install --assumeyes git224 &\
+    yum install --assumeyes python make gnutls-devel libuuid-devel libfaketime &\
+    yum install --assumeyes epel-release &\
+    yum install --assumeyes which cmake3
+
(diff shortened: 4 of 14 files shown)


tbabej:
Welcoming this initiative, it could make our tests run faster too. While the images are currently written in such a way that they would reuse cached layers locally (and hence would not need to re-install OS packages each time), that does not seem to be happening in Travis by default.

Another way to achieve this would be to push the built test images and then use `--cache-from something/test-taskwarrior:ubuntu1804` to reuse the OS installation layers.

sruffell:
@8-bit-fox I think a change like this should go into the 2.6.0 branch first and then be backported to 2.5.2 if necessary. 

Otherwise, I have https://github.com/sruffell/taskwarrior/commits/wip-remove-project-content as an example, that does the centos7 and ubuntu1804 images, but otherwise I think it should wait for #2349 to be merged first before this is merged in.

8-bit-fox:
In general, I'd agree. However, in #2339 @pbeckingham suggested that hewould cherry-pick commits from 2.6.0 to 2.5.2 and rather create a 3.0.0. Thus I am finebasing it on master, but 2.6.0 just seems rebase overhead to me.

But despite that - is there any reason to have two different PRs? Rebasing and changing the draft PR is a matter of seconds.

sruffell:
> In general, I'd agree. However, in #2339 @pbeckingham suggested that hewould cherry-pick commits from 2.6.0 to 2.5.2 and rather create a 3.0.0. Thus I am finebasing it on master, but 2.6.0 just seems rebase overhead to me.

Maybe I'm confused, but it looks like your PR is scheduled to go into 2.5.2?  The current head of development is 2.6.0 is shouldn't new work go there and then be backported to the stable branch as necessary (in this case 2.5.2)

Do you have a link to the specific comment where @pbeckingham said he would cherry-pick from 2.6.0 onto 2.5.2 (which, again, seems to suggest that changes should go into 2.6.0)?  I couldn't find it when I did a cursory scan of #2339.

8-bit-fox:
> Do you have a link to the specific comment where @pbeckingham said he would cherry-pick from 2.6.0 onto 2.5.2 (which, again, seems to suggest that changes should go into 2.6.0)? I couldn't find it when I did a cursory scan of #2339.

Ah, my bad. I mistyped the issue number. It was #2338 . On mobile you don't get the pop-up with the issue titles and I didn't confirm. The specific comment is [here](https://github.com/GothenburgBitFactory/taskwarrior/issues/2338#issuecomment-729067633).

8-bit-fox:
> Maybe I'm confused, but it looks like your PR is scheduled to go into 2.5.2? The current head of development is 2.6.0. Shouldn't new work go into 2.6.0 and be backported to the stable branch as necessary (in this case 2.5.2)?

I must admit, I do not understand the branching strategy here yet. The common ancestor of master, 2.5.2 and 2.6.0 is  [Post251ToMaster](8f09a35ad20e94b7418ce2de4e875db261a6d397). According to the comment, it rather looks as if 2.5.2 would be the future and 2.6.0 would deprecate in favor of 3.0.0

IMO the only question is if SemVer is implemented here. If so, a major or minor version change would be the place to put this refactoring. Or - it is regarded as a bug fix (coupling of test infrastructure and project code being the bug :wink: ) so we should put it to 2.5.2 :slightly_smiling_face:

tbabej:
@8-bit-fox TW tries to follow semantic versioning. That said, general CI improvement like this one will be merged into all active development branches. As @sruffell mentions, the 2.6.0 is the development head, so there is a slight preference for this MR to be pointed towards 2.6.0, and it will be backported to 2.5.2 at the time of merge.

> Thus I am finebasing it on master, but 2.6.0 just seems rebase overhead to me.

As README mentions (let us know if it is not clear!), `master` in this project is actually last stable release. In this particular case that's end of 2.5.1 branch, so basing this MR on master would actually create more work (technically it would be quite difficult because Travis CI was never even introduced on 2.5.1 :slightly_smiling_face: ).

With respect to 2.6.0 / 3.0.0 discussion, there is a chance that 2.6.0 will be renamed to 3.0.0 at some point in the future, but that is I think irrelevant to the discussion at hand. The branch contant will stay the same during the change.

TL;DR: As a general rule, 2.6.0 should work as the target branch. But really, for the purpose of this MR, it does not matter much.

8-bit-fox:
Thanks, @tbabej , for clarification!

8-bit-fox:
@tbabej any reason for closing?

tbabej:
I don't remember doing so, I think this got automatically closed by Github because we deleted the target branch 2.5.2 after its release.

tbabej:
@8-bit-fox I am not able to change the target branch it seems, maybe you can? We should definitely reopen this PR or open its equivalent against 2.6.0. Good catch.

8-bit-fox:
I'll reopen it once I had the chance to rebase.

PR #2349, opened 2020-11-22
Added docker files for Fedora 32 and 33, Centos 8 and Ubuntu 20.04. Changed docker-compose and .travis.yml files

kwoot:
#### Description

Added docker files for Fedora 32 and 33, Centos 8 and Ubuntu 20.04. Changed docker-compose and .travis.yml files

- [x] Have you run the test suite? Nope. That is Travis job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,11 @@ matrix:
       os: linux
       env: CONTAINER=centos7
       services: docker
-    - name: "Fedora 28"
+    - name: "Centos 8"
+      os: linux
+      env: CONTAINER=centos8
+      services: docker
+     - name: "Fedora 28"
       os: linux
       env: CONTAINER=fedora28
       services: docker
@@ -18,7 +22,15 @@ matrix:
       os: linux
       env: CONTAINER=fedora28
       services: docker
-    - name: "Debian Stable"
+     - name: "Fedora 32"
+      os: linux
+      env: CONTAINER=fedora32
+      services: docker
+    - name: "Fedora 33"
+      os: linux
+      env: CONTAINER=fedora33
+      services: docker
+     - name: "Debian Stable"
       os: linux
       env: CONTAINER=debianstable
       services: docker
@@ -34,7 +46,11 @@ matrix:
       os: linux
       env: CONTAINER=ubuntu1804
       services: docker
-    - name: "OpenSUSE 15.0"
+    - name: "Ubuntu 20.04"
+      os: linux
+      env: CONTAINER=ubuntu2004
+      services: docker
+     - name: "OpenSUSE 15.0"
       os: linux
       env: CONTAINER=opensuse1500
       services: docker
--- a/test/docker/centos8
+++ b/test/docker/centos8
@@ -1,0 +1,30 @@
+FROM centos:8
+
+RUN yum update -y
+RUN yum install python git gcc gcc-c++ make gnutls-devel libuuid-devel libfaketime -y
+RUN yum install epel-release centos-release-scl -y
+RUN yum install which cmake3 devtoolset-8-gcc* -y
+RUN source scl_source enable devtoolset-8; gcc --version; cmake3 --version
+RUN ln -s /usr/bin/cmake3 /usr/bin/cmake
+
+# Setup language environment
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+
+# Setup taskwarrior
+ADD . /root/code/
+WORKDIR /root/code/
+RUN git clean -dfx
+RUN git submodule init
+RUN git submodule update
+RUN source scl_source enable devtoolset-8; cmake -DCMAKE_BUILD_TYPE=debug .
+RUN source scl_source enable devtoolset-8; make -j2
+RUN source scl_source enable devtoolset-8; make install
+RUN task --version
+
+# Setup tests
+WORKDIR /root/code/test/
+RUN source scl_source enable devtoolset-8; make
+
+CMD ["bash", "-c", "./run_all -v ; cat all.log | grep 'not ok' ; ./problems"]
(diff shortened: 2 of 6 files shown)


test/docker/ubuntu2004
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:20.04
```
8-bit-fox, line 1: May I suggest using
```suggestion
FROM ubuntu:latest
```
It seems to default to the latest LTS. In addition, we could add a Dockerfile for `ubuntu:devel`, which is the current Ubuntu version.

8-bit-fox (commented):
I found that gentoo requires some rework, as the base image repo is deprecated. I suggest
```Dockerfile
FROM gentoo/portage:latest as portage
FROM gentoo/stage3:x86
COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
```

Archlinux also has a deprecation notice on `archlinux/base`. Instead, one should use `archlinux:base`:
```Dockerfile
FROM archlinux:base
```

test/docker/ubuntu2004
```diff
@@ -0,0 +1,27 @@
...
+
+RUN apt-get update
+RUN apt-get install -y build-essential cmake git uuid-dev libgnutls28-dev libfaketime
+RUN apt-get install -y python
```
8-bit-fox, line 5: When building the image locally, I had the issue that `tzdata` installation (likely as dependency) required manual intervention. This can be circumvented by setting 
```
DEBIAN_FRONTEND="noninteractive" 
```
in the corresponding `RUN` line.

8-bit-fox:
Maybe we should remove ubuntu:16.04? It is quite old and reaches EoL in April 2021 (5 months).

tbabej:
What is the relationship between #2347 and #2349? Is one replacing the other or are they complementary and should be both considered for merging?

8-bit-fox:
I suggest merging both.
#2349 provides container definitions for current OSs. #2347 aims at refactoring the setup. In terms of priority, I'd go for merging #2349 first and #2347 at a later stage (currently still a draft and wip).

tbabej:
> Maybe we should remove ubuntu:16.04?

I'd suggest we leave it there for now, at least until EOL is reached. Taskwarrior has a history of trying to be supportive of older setups, and especially with recent changes like enabling C++17 support it would be nice to have coverage over older popular distro releases.

test/docker/ubuntu2004
kwoot: I do not think this is a good idea. It would mean we would introduce a dynamic variable in the travis configuration that would change with every new release. I think it's better to keep it clearly distinguishable.

test/docker/ubuntu2004
kwoot: Will change

tbabej: This is the only outstanding issue here, otherwise this looks good to merge.

tbabej: Needs an update here.

.travis.yml
```diff
@@ -10,15 +10,27 @@ matrix:
...
+      env: CONTAINER=centos8
+      services: docker
+     - name: "Fedora 28"
       os: linux
       env: CONTAINER=fedora28
       services: docker
     - name: "Fedora 29"
       os: linux
       env: CONTAINER=fedora28
       services: docker
```
mrossinek, line 24: Both of these have reached their EOL. Thus, I think we are safe to remove them.
If not, we should at least fix the F29 one to actually built that rather than F28.
I.e.
```suggestion
     - name: "Fedora 28"
      os: linux
      env: CONTAINER=fedora28
      services: docker
    - name: "Fedora 29"
      os: linux
      env: CONTAINER=fedora29
      services: docker
```

test/docker/centos8
```diff
@@ -0,0 +1,30 @@
+FROM centos:8
+
+RUN yum update -y
+RUN yum install python git gcc gcc-c++ make gnutls-devel libuuid-devel libfaketime -y
+RUN yum install epel-release centos-release-scl -y
+RUN yum install which cmake3 devtoolset-8-gcc* -y
```
mrossinek, line 6: Similar as in #2357 the following may be necessary:
```suggestion
RUN yum install python3 git gcc gcc-c++ make gnutls-devel libuuid-devel -y
RUN yum install epel-release centos-release-scl -y
RUN yum install which cmake3 devtoolset-8-gcc* libfaketime -y
```

test/docker/fedora32
```diff
@@ -0,0 +1,26 @@
+FROM fedora:32
+
+RUN dnf update -y
+RUN dnf install python git gcc gcc-c++ cmake make gnutls-devel libuuid-devel libfaketime -y
```
mrossinek, line 4: The same can be done here, just to be sure it works:
```suggestion
RUN dnf install python3 git gcc gcc-c++ cmake make gnutls-devel libuuid-devel libfaketime -y
```

test/docker/fedora33
```diff
@@ -0,0 +1,26 @@
+FROM fedora:33
+
+RUN dnf update -y
+RUN dnf install python git gcc gcc-c++ cmake make gnutls-devel libuuid-devel libfaketime -y
```
mrossinek, line 4: And here, too...
```suggestion
RUN dnf install python3 git gcc gcc-c++ cmake make gnutls-devel libuuid-devel libfaketime -y
```

test/docker/ubuntu2004
mrossinek, line 5: And finally:
```suggestion
RUN apt-get install -y build-essential cmake git uuid-dev libgnutls28-dev faketime
RUN apt-get install -y python3
```

mrossinek (changes requested):
We need to make sure that the same patches as introduced to the existing docker containers in #2357 are applied here, too.

tbabej:
I attempted to review and merge, but some of the docker images in this PR do not successfully build (i.e. the centos8). However, I was able to use the learnings, and good tips from @mrossinek to get F31-F33, Centos 8 and Ubuntu 20.04 working.

As we currently have working CI with modern distributions, I'm closing this PR. Great work everybody.

Issue #1928, opened 2018-02-14
[TW-1915] Filter combinations do not work


taskwarrior:
_Florian on 2017-06-12T13:27:40Z says:_ 

 Create the following tasks:

`$ task log project:A item1`
 `$ task log project:B item2`
 `$ task add project:A item3`
 `$ task add project:B item4`

I expect all of the following filters to be equivalent. The following filters work and yield item3 and item4:

`$ task status:pending "(project:A or project:B)" all`
 `$ task status:pending "( project:A or project:B )" all`
 `$ task status:pending '(project:A or project:B)' all`
 `$ task status:pending BACKSLASH( project:A or project:B BACKSLASH) all`
 `$ task "(project:A or project:B) and status:pending" all`

The following filters do not work and yield »No matches«:

`$ task "status:pending and (project:A or project:B)" all`
 `$ task 'status:pending and (project:A or project:B)' all`
 `$ task "status:pending and ( project:A or project:B )" all`

A non-working example »`status:pending and (project:Home or project:Garden)`« is listed in [https://taskwarrior.org/docs/filter.html]

taskwarrior:
Migrated metadata:
```
Created: 2017-06-12T13:27:40Z
Modified: 2017-06-12T13:27:40Z
```

Issue #1654, opened 2018-02-14
[TW-1630] "Due" parsing behavior seems inconsistent


taskwarrior:
_Jim B on 2015-07-08T07:48:02Z says:_ 

 This behavior occurs on my Cygwin install, as well as the online demo at wbsch.de.

I can't make sense of the way "due" modifiers are working here:

I am attempting to create a task which is due in 7 days (i.e., "now" + 7 days) and then searching for it using "due" modifiers -- in this example, I expect the output to show me all tasks which are due in 10 days or less.

demo(at)taskwarrior:~$ task add due:7d TEST TASK
Created task 1.
demo(at)taskwarrior:~$ task due.before:10d
[task next due < 10d]
No matches.
demo(at)taskwarrior:~$ task due.after:10d
[task next due > 10d]

ID Age Due    Description Urg 
 1 12s 6 days TEST TASK    5.6

1 task
demo(at)taskwarrior:~$ task 1 mod due:15d
Modifying task 1 'TEST TASK'.
Modified 1 task.
demo(at)taskwarrior:~$ task
[task next]

ID Age Due     Description Urg 
 1 57s 2 weeks TEST TASK    2.4

1 task

I think the example shows that when a task is created or modified, reasonable behavior occurs; but when searching/matching tasks, something completely different happens as though "due.before" and "due.after" are doing the opposite of what one would expect.

taskwarrior:
Migrated metadata:
```
Created: 2015-07-08T07:48:02Z
Modified: 2015-08-05T21:47:32Z
```

taskwarrior:
_Paul Beckingham on 2015-08-05T21:01:54Z says:_ 

 Thanks!

mrossinek:
There are some remaining issues with this meaning that this issues should be re-opened for now.

timokau:
`due.before:3d` worked for me in 2.5.1 and now no longer works in 2.5.2. It is returning no results although there should be some. I'm not sure if that is the same problem that was described in this issue, since it worked in 2.5.1.

timokau:
As a workaround: `due.before:now+3d` still works.

tbabej:
This is currently describing same issue as #2390.

PR #2350, opened 2020-11-22
Fix tests on 2.5.2 branch

mrossinek:
#### Description

This PR backports some of the test fixes already present on the 2.6.0 branch onto the 2.5.2 branch.
It also fixes some other details in the tests which were outdated.

```
% ./problems
Failed:
```

The above failures are still related to the following problems:
- [ ] `project.t`, `search.t` and one in `filter.t`: related to a syntax issue. We can choose to document it properly and fix the unittests to reflect this syntax. Some discussion on this took place in #2345.
- [ ] `filter.t` part 1: 2 failures are related to #1654 which is closed but not fixed. I think we should re-open the issue and mark the failures as expected for now with a TODO label
- [ ] `filter.t` part 2: the remaining 2 failures are related to the open issue #1928. I suggest doing the same as above (mark as expected and add a TODO flag).

### DIFF
--- a/src/Variant.cpp
+++ b/src/Variant.cpp
@@ -1278,7 +1278,18 @@ Variant& Variant::operator-= (const Variant& other)
     break;
 
   case type_string:
-    throw std::string (STRING_VARIANT_SUB_STRING);
+    switch (right._type)
+    {
+    case type_string:
+      cast (type_string);   _string   += '-' + right._string;   break;
+    case type_boolean:
+    case type_integer:
+    case type_real:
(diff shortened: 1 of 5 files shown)


mrossinek:
@pbeckingham thanks for merging. Should we address the remaining failing tests which I described above, too? Especially the ones regarding the correct escaping syntax just need a decision how to be handled. And the other ones we can mark as expected failures and mark the respective issues with a `help wanted` label (re-opening if necessary).
What do you think?

PR #2354, opened 2020-11-24
Cherry-pick PRs accidentally merged to master into 2.5.2

liskin:
#### Description

#2223, #2280, #2320 were merged to master but not to 2.5.2 nor 2.6.0.
This PR cherry-picks #2280 and #2320 to 2.5.2.
PR #2223 is skipped for two reasons:

1. most commands now use Table instead of ViewTask, so the benefit of the change is quite limited
2. it really is a quick dirty hack that needs more review/work before being merged

### DIFF
--- a/doc/man/task.1.in
+++ b/doc/man/task.1.in
@@ -604,7 +604,7 @@ by third-party applications.
 Reports a unique set of attribute values. For example, to see all the active
 projects:
 
-  task +PENDING _unique projects
+  task +PENDING _unique project
 
 .TP
 .B task <filter> _uuids
--- a/src/TLSClient.cpp
+++ b/src/TLSClient.cpp
@@ -37,11 +37,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <arpa/inet.h>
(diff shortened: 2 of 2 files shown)


tbabej:
Thanks for keeping an eye on this :)

liskin:
Do I need to keep an eye on the 2.6.0 branch too or will everything in 2.5.2 be eventually merged there?

Oh and the master branch can now probably really be dropped, unless you wish to keep it around as a sort of "last released version", in which case it could use a force-push to discard the 5 or so patches merged after 2.5.1.

mrossinek:
With 2.5.2 now released it may be a good idea to do a last check whether 2.5.3 and 2.6.0 are now inline. Is there an easy way to do this or would that have to be done in a manual effort?

liskin:
There's git cherry (not cherry-pick) which hashes each individual commit and check if it's in the other branch, but when I tried it last week it flagged dozens/hundreds of patches as different between the branches, so it may be better to just git merge 2.5.2 into 2.6.0, which doesn't operate on individual patches but checks the final diff, and deal with the conflicts, if any, manually.

PR #2357, opened 2020-11-26
Update Docker images to use Python 3

mrossinek:
#### Description

As discussed in #2345, the CI currently fails not only due to problematic tests but also because none of our docker images actually satisfy the Python requirements.
This PR attempts to fix that by:
- [x] replacing `/usr/bin/env python` with `/usr/bin/env python3` in all tests
- [x] replacing `python` with `python3` in the docker configuration files

This PR will also look into:
- [ ] ensuring `faketime` or `libfaketime` is installed properly in each docker container

- Have you run the test suite? The purpose of this PR is to fix the Travis CI so I will be inspecting its logs closely :wink:


tbabej:
Thanks Max!

tbabej:
Looking at the results of the CI after merging this, it looks like F28 and F29 containers succeed? [1] Very surprising that they do not suffer from the Lexer test failing..

https://travis-ci.org/github/GothenburgBitFactory/taskwarrior/builds/746684706

mrossinek:
Yes, I noticed that too. I can look into it later this week but on a first glance I would assume that for some reason the `lexer.t` tests are not being run.
As a first check I would make sure that any of the compiled tests are run. If not this could point to a failure in compiling them..

mrossinek:
Indeed, it appears as though the Fedora28 container does not compile the tests: https://travis-ci.org/github/GothenburgBitFactory/taskwarrior/jobs/746684708#L1609

I also noticed, that the Fedora29 container is not configured correctly and actually uses the Fedora28 docker file, too. As seen on line 15 in https://travis-ci.org/github/GothenburgBitFactory/taskwarrior/jobs/746684709/config

However, I will leave these as is because hopefully #2349 will integrate Fedora 32 and 33 successfully, after which we should be okay to remove F28 and F29 anyways.

mrossinek:
There also appears to be a problem with the Ubuntu containers, as seen here:
https://travis-ci.org/github/GothenburgBitFactory/taskwarrior/jobs/746684712#L6150

The Ubuntu containers run into many errors of the following kind (https://travis-ci.org/github/GothenburgBitFactory/taskwarrior/jobs/746684712#L5751):
```python
not ok 8 - search.t: 1469: Explicit search, case insensitive, no regex
# ERROR: TypeError on file /usr/lib/python3.5/unittest/case.py line 58 in testPartExecutor: 'yield':
#        'TimeoutWaitingFor' object is not iterable
```

On first inspection I think it may be related to how the `_retrieve_output` method defined in `test/basetest/utils.py` handles such a timeout error. See here: https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/test/basetest/utils.py#L172 and the following extract:
```python
    # Thread died so we should have output
    try:
        # data = (stdout, stderr, exitcode)
        data = queue.get(timeout=timeout)
    except Empty:
        data = TimeoutWaitingFor("streams from TaskWarrior")

    return data
```
As you can see, the `TimeoutWaitingFor` error is not actually raised correctly resulting in a Python error as seen in Ubuntu rather than an actual timeout error.

I think changing the above to an actual `raise`-statement might help, but it won't fix why the timeout appears in the first place.

@tbabej What do you think about this?

tbabej:
@mrossinek This turned out to be a incorrectly configured locale, which ended up trying to run TW commands using ascii encoding. We need utf-8 correctly working to run certain tests. This breaking might be a byproduct of recent change to use Python3 for the test suite.

See bf99dedc073fdd854fe34b488703b50d9196c11a for the fix.

mrossinek:
@tbabej Awesome, thanks for figuring that out! :slightly_smiling_face:

PR #2362, opened 2020-12-06
Fix final failing test on 2.5.3

### LINKED ISSUE #2023: Layout breaks if report does not fit in the available width

mrossinek:
#### Description

This reverts commit 64243e6ec1919e5ce41f95b00cffe8b263ebb980 which was introduced by #2223.

As discussed in #2345 there is one final test on the 2.5.3 branch which is failing. This test is passing fine on 2.6.0 and also used to pass fine on 2.5.2 because the failure is caused by #2223 which was merged into `master` directly (which 2.5.3 is now based on).
#2223 attempted to fix #2023 by introducing a workaround to handle linewrapping. However, since the 2.5.2 release most commands now actually use a table rather than `ViewTask` (thanks @liskin).

Thus, reverting 64243e6 does not actually cause any problems as shown below:
screenshot_1607262855

```
% ./run_all
Passed:                          3776
Skipped:                            5
Expected failures:                 11
Runtime:                        20.24 seconds
```

### DIFF
--- a/src/ViewTask.cpp
+++ b/src/ViewTask.cpp
@@ -310,11 +310,10 @@ std::string ViewTask::render (std::vector <Task>& data, std::vector <int>& seque
       row_color.blend (rule_color);
     }
 
-    const int off = overage < 0 ? overage : 0;
     for (unsigned int c = 0; c < _columns.size (); ++c)
     {
       cells.push_back (std::vector <std::string> ());
-      _columns[c]->render (cells[c], data[sequence[s]], widths[c]+off, row_color);
+      _columns[c]->render (cells[c], data[sequence[s]], widths[c], row_color);
 
       if (cells[c].size () > max_lines)
         max_lines = cells[c].size ();


mrossinek:
A quick note: I still expect the CI to fail for this PR because the changes to the CI workflow have been made on the 2.6.0 branch.
@tbabej we should probably backport all of that onto this branch too. I.e. the new docker container configuration files as well as the usage of Python3 for the unittests.

tbabej:
Thank you! Yes, backporting the CI fixes is one of the tasks on the, ehm, todo list :)

mrossinek:
@tbabej I am happy to open a PR for the backporting later in the evening today, if you don't plan to tackle this earlier :+1:

tbabej:
Oops, saw this too late. Well, everything should be backported and working (see [the 2.5.3 commit log](https://github.com/GothenburgBitFactory/taskwarrior/commits/2.5.3?since=2020-12-06&until=2020-12-06)), now I'm just waiting for Travis to prove me correct :)

Issue #2322, opened 2020-09-06
Automatically remove wait date when completing a task


xerus2000:
Sometimes, I complete a task right off my someday-list. In that case, the wait date should be automatically deleted - keeping it doesn't make much sense.

skbolton:
This is a good use case for a hook if this is the behavior you want.

tbabej:
I think @xerus2000 is right about this one. At first I was considering we might want to keep it for historical purposes, but the information would be preserved by the change history.

tbabej:
Reproduced on 2.6.0.

mrossinek:
I opened #2371 to implement this :+1:

tbabej:
Implemented by @mrossinek.

PR #2371, opened 2020-12-12
Un-wait a waiting task when done

mrossinek:
#### Description

Implements #2322 

The unittest mirrors the `TestFeature608.test_done_stop` method in `start.t`.


mrossinek:
@tbabej Given your recent activity I noticed I still have this pending PR. I don't know why the CI never ran for this though and I cannot retrigger it.. Would you mind giving it a review as is? If you have some comments a further commit should trigger the CI again :+1:

tbabej:
I wonder if we want this behaviour also for `task delete`?

mrossinek:
> I wonder if we want this behaviour also for `task delete`?

@tbabej I extended the fix to `DeleteCommand` as well :+1:

tbabej (approved):
Looks good to me!

Issue #2283, opened 2020-05-24
task edit json encodes annotations


jspricke:
#### To report a bug...

Since https://github.com/GothenburgBitFactory/taskwarrior/commit/e4b9c1f01ec1b6c3a6c1f47597e1bde3788dd83a annotations are JSON encoded in the editor making them hard to parse and edit. According to the ChangeLog this is only to escape newlines. I would propose replace newlines by `\n` (or some other char) and undo it after the edit.
I'm happy to implement this if you can confirm that the JSON encoding is only for the newline and that `\n` would be acceptable.

* What command(s) did you run?
```
task add foo
task 1 annotate https://taskwarrior.org/
task 1 edit
```
* What did you expect to happen?

see https://taskwarrior.org/ in the editor window

* What actually happened?
```
Annotation:        XXX -- https:\/\/www.taskwarrior.org\/
```
(JSON encoded URL)

* Paste the output of the ```task diag``` command.
```
> task diag

task 2.5.1
   Platform: Linux

Compiler
    Version: 9.2.1 20200224
       Caps: +stdc +stdc_hosted +c8 +i32 +l32 +vp32 +time_t32
 Compliance: C++11

Build Features
      CMake: 3.16.3
    libuuid: libuuid + uuid_unparse_lower
  libgnutls: 3.6.12
 Build type: None

Configuration
       File: /home/jspricke/.taskrc (found), 859 bytes, mode 100600
       Data: /home/jspricke/.task (found), dir, mode 40700
    Locking: Enabled
         GC: Enabled
    $EDITOR: nvim
     Server:
      Trust: strict
Certificate: , not readable, 0 bytes
        Key: , not readable, 0 bytes
    Ciphers: NORMAL
      Creds:

Hooks
     System: Enabled
   Location: /home/jspricke/.task/hooks
     Active: on-add    (executable)
             on-modify (executable) (symlink)
   Inactive:

Tests
      $TERM: tmux-256color (357x112)
       Dups: Scanned 415 tasks for duplicate UUIDs:
             No duplicates found
 Broken ref: Scanned 415 tasks for broken references:
             No broken references found
```

tbabej:
@jspricke Confirming that this indeed seems to be the case. However, I'm not sure using an escape character is the way to go. What if the annotation itself already contains (escaped) `\n` character? Now both escaped and unescaped newlines would be represented by the same character.

[...]

PR #2383, opened 2021-01-05
Support editing multi-line annotations (Closes: #2283)

jspricke:
#### Description

Since e4b9c1f annotations where JSON encoded in task edit to escape
new lines (\n). But other strings where mangled as well, like https://
becoming https:\/\/, making it hard to edit.

This patch removes the JSON encoding and indents new lines instead.

Failed:

Unexpected successes:

Skipped:

Expected failures:
filter.t                            4

### DIFF
--- a/src/commands/CmdEdit.cpp
+++ b/src/commands/CmdEdit.cpp
@@ -173,23 +173,6 @@ std::vector <std::string> CmdEdit::findValues (
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-std::string CmdEdit::replaceString (
-  std::string text,
-  const std::string& search,
-  const std::string& replacement)
-{
-  std::string::size_type found = 0;
-
-  while ((found = text.find (search, found)) != std::string::npos)
-  {
-    text.replace (found, search.length (), replacement);
-    found += replacement.length ();
-  }
-
-  return text;
-}
-
-////////////////////////////////////////////////////////////////////////////////
 std::string CmdEdit::formatDate (
   Task& task,
   const std::string& attribute,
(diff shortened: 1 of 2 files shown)


tbabej (changes requested):
Tested and looks amazing! Two general comments:
* There should be no need to implement a new string replacement method, I believe `str_replace` [available from libshared's shared.cpp](https://github.com/GothenburgBitFactory/libshared/blob/a9226163/src/shared.cpp#L140) performs the same task.
* (Optional) Ideally, we'd also add a test for CmdEdit. Test suite so far has no test being able to fake actions takein in the `EDITOR` window, but I believe with bash_tap.sh interface it should be doable (i.e. see [an example of bash test here](https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/test/tw-1718.t)).

jspricke:
> * There should be no need to implement a new string replacement method, I believe `str_replace` [available from libshared's shared.cpp](https://github.com/GothenburgBitFactory/libshared/blob/a9226163/src/shared.cpp#L140) performs the same task.

No idea how I could have missed that, thanks! Updated.

> * (Optional) Ideally, we'd also add a test for CmdEdit. Test suite so far has no test being able to fake actions takein in the `EDITOR` window, but I believe with bash_tap.sh interface it should be doable (i.e. see [an example of bash test here](https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/test/tw-1718.t)).

There is one test already here: https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/test/edit.t#L62

tbabej:
Thanks for editing the PR!

> There is one test already here.
I meant a test that would specifically test what we added here, but I think it would be more tricky than the example already given in the codebase. I can add that myself as I already have an idea how to go about it.

tbabej (approved):

PR #2387, opened 2021-01-14
fix: remove lock file on interruption of 'task edit' (#2292)

### LINKED ISSUE #2292: Interruption should remove lock file
When the "Would you like to try again" prompt is interrupted with Ctrl-C, task should remove the temporary/lock file, otherwise you can't edit the task again:

```
./task edit 1                                                                                              2.6.0
Launching 'vi "task.2b9db3dd.task"' now...
Editing complete.
Edits were detected.
Error: Circular dependency detected and disallowed.
Taskwarrior couldn't handle your edits.  Would you like to try again? (yes/no) ^C

Interrupted: No changes made.

./task edit 1                                                                                              2.6.0
Filter: ( ( ( id == 1 ) ) )
Task is already being edited.
```

Steps to produce:

* Add task 1
* Add task 2 with dependency on 1
* `task 1 edit` and add dependency on task 2; save
* At prompt, press Ctrl-1
* `task 1 edit` fails because of lock file

Confirmed with GitHab 2.6.0 branch from today.

indev29:
Fixes #2292

### DIFF
--- a/src/commands/CmdEdit.cpp
+++ b/src/commands/CmdEdit.cpp
@@ -826,12 +826,16 @@ ARE_THESE_REALLY_HARMFUL:
     {
       std::cerr << "Error: " << problem << '\n';
 
-      // Preserve the edits.
-      before = after;
-      File::write (file.str (), before);
+      File::remove (file.str());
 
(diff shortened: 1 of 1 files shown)


tbabej:
Thanks for the fix!

PR #2401, opened 2021-01-27
Bash paths correctly dereferenced to account for spaces

nathaniel-c-wheeler:
#### Description
Previously, if the path of the repo had spaces in it (eg "~/Documents/Open Source/taskwarrior"), some of the test scripts would fail. This was due to the bash variables representing the binary not being correctly dereferenced as strings, which lead to bash interpreting the path as two arguments instead of one. The affected scripts have been updated to treat the binary as a string and should operate correctly regardless of special characters in the path.

The version test is currently failing upstream of my changes.

Failed:                        

Unexpected successes:          

Skipped:                       

Expected failures:             
filter.t                            2

### DIFF
--- a/scripts/bash/task.sh
+++ b/scripts/bash/task.sh
@@ -50,14 +50,16 @@
 #
 ################################################################################
 #the following variable is substituted for by ../../test/bash_completion.t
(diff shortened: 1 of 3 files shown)


tbabej:
Looks good. Thanks for making the test suite more robust!

PR #2402, opened 2021-01-28
Implement writeable context

### LINKED ISSUE #1955: [TW-1944] adding new tasks in context
_Jean-Francois Joly on 2017-04-15T12:59:06Z says:_ 

 Could tasks added in context have the proper tags added implicitly, in such a way that the tasks would show up as part of the current context without having to add them manually?

As an example, if I am in the +work context and I add a task, I would not have to specify +work every time.

tbabej:
Support for context for the `add` and `log` commands.

Closes #1955.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,6 +3,7 @@ contributions of the following people:
 
   Paul Beckingham    (Principal Author)
   Federico Hernandez (Principal Author)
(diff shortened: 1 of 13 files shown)


kwoot (commented):
Oops...

PR #2404, opened 2021-01-29
build: make sure internal libs are built as static

### LINKED ISSUE #2403: Undefined symbols in taskwarrior 2.5.3
Compilation fails with:
```
/usr/bin/ld: libtask.so: undefined reference to `Command::read_only() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::factory(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Command*, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Command*> > >&)'
/usr/bin/ld: libtask.so: undefined reference to `Command::accepts_filter() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::needs_gc() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::accepts_modifications() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::displays_id() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::accepts_miscellaneous() const'
/usr/bin/ld: libtask.so: undefined reference to `Command::uses_context() const'
/usr/bin/ld: libtask.so: undefined reference to `Column::factory(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Column*, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Column*> > >&)'
```

That's because ie CLI2.cpp using Command is built into libtask but Command.cpp is not. Trying to add Command.cpp triggers even more, though different unresolved symbols. If ie `CLI2::findCommand()` is not supposed to be used in libtask then perhaps it would be possible to extract unused part? If it is expected to be used then I suppose Command.cpp must be added to libtask.

Same applies to `Context::initialize()` which makes use of `Column::factory()` even though Column.cpp is not included in libtask.

jpalus:
these libararies are not installed so don't leave the decision up to
enviroment which might build shared libraries resulting in binary
with missing deps after installation

Fixes #2403

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,7 +6,7 @@ include_directories (${CMAKE_SOURCE_DIR}
                      ${CMAKE_SOURCE_DIR}/src/libshared/src
                      ${TASK_INCLUDE_DIRS})
 
-add_library (task CLI2.cpp CLI2.h
+add_library (task STATIC CLI2.cpp CLI2.h
                   Context.cpp Context.h
                   DOM.cpp DOM.h
                   Eval.cpp Eval.h
@@ -28,7 +28,7 @@ add_library (task CLI2.cpp CLI2.h
                   sort.cpp
                   util.cpp util.h)
 
-add_library (libshared libshared/src/Color.cpp         libshared/src/Color.h
+add_library (libshared STATIC libshared/src/Color.cpp         libshared/src/Color.h
                        libshared/src/Configuration.cpp libshared/src/Configuration.h
                        libshared/src/Datetime.cpp      libshared/src/Datetime.h
                        libshared/src/Duration.cpp      libshared/src/Duration.h


tbabej:
To force building shared libraries (in order to reproduce #2403), one can use `cmake -DBUILD_SHARED_LIBS=ON .` [1]. This change enforces static build (as behaviour is left to local settings otherwise, see [2])

@jpalus Thanks for identifying and fixing this!

[1] https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS
[2] https://cmake.org/cmake/help/latest/command/add_library.html

PR #2448, opened 2021-04-11
Parse 'context' verbose token correctly

pcapriotti:
`context` was not included among the possible verbosity tokens, resulting into the `verbose` setting being incorrectly set to `off` in some cases.


tbabej (approved):
Thanks for the patch. Good catch!

PR #2464, opened 2021-04-22
Support relative include for themes in taskrc

### LINKED ISSUE #1847: [TW-1831] Allow including files relative to $PREFIX/share/
_Georg Sauthoff on 2016-07-11T20:02:57Z says:_ 

 Selecting a color scheme or holidays via including a file from `/usr/share/doc/task/rc` is convenient.

But, when sharing a `~/.taskrc` between different systems, this isn't always portable. That means on systems where task warrior is installed under a non-standard prefix (i.e. $PREFIX != /usr), one has to adjust such settings.

Thus, it would be nice to be able to specify such includes relative to the task warrior system prefix share directory.

Related features are:
*  tilde expansion - task warrior already supports this
*  include files relative to the origin of the current file (e.g. this is how git/hg import relative paths) - doesn't help here, but is nice for other use cases
*  have a include search path (like e.g. with the C preprocessor) - would help if `$PREFIX/share/doc/task/rc` is then part of the default search path

oxalica:
#### Description

Support relative include for themes in taskrc. Fix #1847.
This is required for package manager like nix, which installs different version of a package to different paths. We need to keep paths of `include` in taskrc to be always valid between updates.

This PR depends on https://github.com/GothenburgBitFactory/libshared/pull/38 to be merged.

  All tests passed.

### DIFF
--- a/doc/man/task-color.5.in
+++ b/doc/man/task-color.5.in
@@ -236,11 +236,10 @@ include other files into the .taskrc file, different sets of color rules can
 be included.
 
 To get a good idea of what a color theme looks like, try adding this entry to
-your .taskrc file (note that your installation may use a slightly different path
-from the example):
+your .taskrc file:
 
 .RS
-include ${CMAKE_INSTALL_PREFIX}/${TASK_RCDIR}/dark-256.theme
+include dark-256.theme
 .RE
 
 You can use any of the standard Taskwarrior themes:
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -85,7 +85,8 @@ include <file>
 
 There may be whitespace around 'include' and <file>. The file may be an
 absolute or relative path, and the special character '~' is expanded to mean
-$HOME.
+$HOME. If relative path is specified, it will be searched in installation
+directory of configuration files, which contains themes.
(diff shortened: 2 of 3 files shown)


tbabej:
Hi @oxalica, thanks for the patch! I rejoice each time a PR comes with tests and doc updates :slightly_smiling_face: I wonder, if you considered implementing a more generic mechanism proposed in #1847, which would allow for usage of environment variables in the configuration files?

oxalica:
@tbabej  
> ...which would allow for usage of environment variables in the configuration files?

It requires modification on `libshared` side. I pushed a new commit to https://github.com/GothenburgBitFactory/libshared/pull/38 to support environment variable expansion and also search relative paths in the directory of the configuration file itself.

oxalica:
ping @tbabej

tbabej (approved):
@oxalica I think this is ready to go, we just need to update the `libshared` submodule as the commit SHA has changed after merging. The top commit is now b00c80c7a97103551acae7dbe47cf92cc57d308b

oxalica:
@tbabej Rebased.
I also fixed previously failed tests in #2493 . That is not related to code here.

tbabej:
Thank you for the prompt response :) merging both.

PR #2467, opened 2021-04-23
clang-tidy: add missing reserve calls

neheb:
Found with performance-inefficient-vector-operation

- [x ] Have you run the test suite?
```
Failed:

Unexpected successes:

Skipped:

Expected failures:
tw-2124.t                           1
```

probably a bogus failure.

### DIFF
--- a/src/Hooks.cpp
+++ b/src/Hooks.cpp
@@ -196,6 +196,7 @@ void Hooks::onExit () const
 
     // Convert to a vector of strings.
     std::vector <std::string> input;
+    input.reserve(tasks.size());
     for (auto& t : tasks)
       input.push_back (t.composeJSON ());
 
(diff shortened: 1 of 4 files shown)


neheb:
Hmm the constructor also accepts a size parameter. Which is better, reserve or constructor?

tbabej:
Hi @neheb, thanks for the patch!

> Hmm the constructor also accepts a size parameter. Which is better, reserve or constructor?

Indeed, you can use [fill constructor](https://www.cplusplus.com/reference/vector/vector/vector/) to initialize a vector of lenght N, however, that would fill out the vectors with some default value, requiring further changes to the code (namely, not using `push_back`, but directly rewriting elements of the vector.

IMHO your proposed changed is less invasive.

tbabej (approved):

neheb:
makes sense.

I played with this on godbolt. It seems for loops benefit from reserve but algorithms do not. These loops can easily be converted to use std::transform.

PR #2470, opened 2021-04-23
clang-tidy: use const references

neheb:
Found with performance-for-range-copy


tbabej (approved):
Looks good to me :slightly_smiling_face: thanks for the PR!

PR #2471, opened 2021-04-23
clang-tidy: use auto

neheb:
Found with modernize-use-auto


tbabej:
Looks good to me. While I'm not a fan of overusing `auto`, this does indeed increase readability.

neheb:
Right. With these, the types are duplicated otherwise.

PR #2472, opened 2021-04-23
clang-tidy: loop conversion

neheb:
Found with modernize-loop-convert

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -409,8 +409,6 @@ static const char* modifierNames[] =
   "noword"
 };
 
-#define NUM_MODIFIER_NAMES       (sizeof (modifierNames) / sizeof (modifierNames[0]))
-
 Context* Context::context;
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -564,8 +562,8 @@ int Context::initialize (int argc, const char** argv)
     //
     ////////////////////////////////////////////////////////////////////////////
 
-    for (unsigned int i = 0; i < NUM_MODIFIER_NAMES; ++i)
-      cli2.entity ("modifier", modifierNames[i]);
+    for (auto& modifierName : modifierNames)
+      cli2.entity ("modifier", modifierName);
 
     for (auto& op : Eval::getOperators ())
       cli2.entity ("operator", op);
--- a/src/Eval.cpp
+++ b/src/Eval.cpp
@@ -195,8 +195,9 @@ void Eval::debug (bool value)
 std::vector <std::string> Eval::getOperators ()
 {
   std::vector <std::string> all;
-  for (unsigned int i = 0; i < NUM_OPERATORS; ++i)
-    all.push_back (operators[i].op);
+  all.reserve(NUM_OPERATORS);
+  for (const auto &opr : operators)
+    all.push_back (opr.op);
 
   return all;
 }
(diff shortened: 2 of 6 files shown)


src/Eval.cpp
```diff
@@ -195,8 +195,9 @@ void Eval::debug (bool value)
 std::vector <std::string> Eval::getOperators ()
 {
   std::vector <std::string> all;
-  for (unsigned int i = 0; i < NUM_OPERATORS; ++i)
-    all.push_back (operators[i].op);
+  all.reserve(sizeof(operators));
```
tbabej, line 198: This is not what we want, `sizeof(operators)` will return memory-size of the array (including elements), not the number of elements. Why not use the `NUM_OPERATORS` constant which was made for that?

src/Eval.cpp
```diff
@@ -195,8 +195,9 @@ void Eval::debug (bool value)
...
+  for (const auto &opr : operators)
+    all.push_back (opr.op);
```
tbabej, line 200: I'd suggest using a more descriptive name, i.e. `operator` as the loop variable.

src/Context.cpp
```diff
@@ -564,8 +564,8 @@ int Context::initialize (int argc, const char** argv)
     //
     ////////////////////////////////////////////////////////////////////////////
 
-    for (unsigned int i = 0; i < NUM_MODIFIER_NAMES; ++i)
```
tbabej, line 567: `NUM_MODIFIER_NAMES` is unused after this,

src/Eval.cpp
```diff
@@ -206,9 +207,9 @@ std::vector <std::string> Eval::getOperators ()
 std::vector <std::string> Eval::getBinaryOperators ()
 {
   std::vector <std::string> all;
-  for (unsigned int i = 0; i < NUM_OPERATORS; ++i)
-    if (operators[i].type == 'b')
-      all.push_back (operators[i].op);
+  for (const auto &opr : operators)
```
tbabej, line 210: Should reflect the implementation above, that is, (a) use `NUM_OPERATORS` constant to resize upfront and (b) more readable loop variable.

test/lexer.t.cpp
```diff
@@ -487,33 +487,33 @@ int main (int, char**)
   };
   #define NUM_TESTS (sizeof (lexerTests) / sizeof (lexerTests[0]))
```
tbabej, line 488: This variable is unused after this PR is merged and can be removed.

tbabej (changes requested):
This one has a couple of issues that need to be addressed, see comments :speech_balloon:

src/Eval.cpp
neheb: Hmm I'll look into this. Originally, I tried to use operators.size() but that obviously failed.

How about turning operators into an std::array?

src/Eval.cpp
neheb: operator might be a reserved keyword. At least nvim highlighted it. I was hoping to avoid confusion.

neheb: yeah just tried it. Doesn't work.

src/Eval.cpp
neheb: reserving NUM_OPERATORS makes no sense as it's only doing a push_back on a subset of them.

src/Eval.cpp
neheb: Actually std::array would allow turning this loop into a single std::transform. Can't do that as is since auto is not available in lambdas with C++11.

tbabej (approved):
Thanks for responding to the review comments :slightly_smiling_face:

PR #2473, opened 2021-04-23
replace several loops with std::copy

neheb:
std::copy is usually faster and works the same as push_back underneath.

### DIFF
--- a/src/CLI2.cpp
+++ b/src/CLI2.cpp
@@ -1398,8 +1398,7 @@ void CLI2::desugarFilterAttributes ()
         // Getting this wrong breaks a whole lot of filtering tests.
         if (evalSupported)
         {
-          for (auto& v : values)
-            reconstructed.push_back (v);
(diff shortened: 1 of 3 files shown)


tbabej:
Do we know for a fact this is a valid performance improvement? My gut tells me this should be basically the same (as the `std::copy` will translate to roughly the same code it replaces), and anecdotal evidence sees to suggest so: https://hjortsberg.org/notes/C++-vector-insert-performance.html

neheb:
Based on godbolt testing, std::copy is better since it avoids having to call reserve. It's also a one liner.

tbabej:
Let's not do this for now, with C++20 ranges we should have more readable way of doing accomplishing the same. When it comes to performance improvements, according to the profiling information these codepaths are not critical.

neheb:
Maybe these could be replaced with the insert method instead.

PR #2476, opened 2021-04-24
convert loops to std::any_of

neheb:
Simpler. Should also be faster.


tbabej:
Thanks @neheb, I will be selectively implementing some of these manually in the code. In some cases, I'd argue these one-liners decrease readability of the code.

PR #2493, opened 2021-06-02
Prefer /usr/bin/env and searching from PATH

oxalica:
#### Description

- Prefer `/usr/bin/env` in shebangs, since `bash` may not be installed in `/bin` in some systems (like NixOS). And we already use `/usr/bin/env` in most of scripts.
- Use `echo` from `PATH` instead of hard-coded `/bin/echo`, for the same reason.


tbabej:
Thank you for helping make Taskwarrior more portable!

PR #2494, opened 2021-06-02
Remove build date/time from diagnostic message

oxalica:
#### Description

This message does not really help, and makes the binary non-deterministic.


tbabej:
Good point. Thanks for the patch!

Issue #2536, opened 2021-07-17
Feature: inclusive range-end attribute modifier 'by' so 'end of' named dates can be filtered inclusively


smemsh:
Due to a recent change in 2.6.0 branch, implemented in #2522, the `eod`, `eom`, `eoy` intervals now end on the last second of the day.  This in turn means filter modifiers like `.before:eom` will fail to match on the last second of the interval, because comparison uses `<` (see discussion in #2519).  For example:

```
task add test due:eoy
task due before:eoy
```
will not match the new task.  Anyone desiring an inclusive comparison to match would have to manually use `<=` on the command line and dispense with filter attribute modifers.

In order to cover this boundary-case and to give a quicker way to specify an inclusive range-end comparison more generally, the proposal is to add a new `by` filter attribute modifier which behaves similar to the existing `before` modifier, but also includes the endpoint of the range (ie, uses `<=` rather than `<` for comparison).

See full discussion in #2519

Issue #2519, opened 2021-06-29
Named date eod should be last minute of today and not first of tomorrow


vizcay:
```
$ task --version
2.5.3
$ task add XXX due:eod
$ task +TODAY list
No matches
```

Is completely unlogical for me.. if a task dues by the end of today, how is not due TODAY but TOMORROW?

Proposal: make eod set time to 23:59 that is what I manually end up doing.

tbabej:
Hi @vizcay, thanks for the bug report. This is actually behaving [according to the documentation](https://taskwarrior.org/docs/named_dates.html).

That said, I can see how this is not playing nice with other parts of TW, such as the `+TODAY` virtual tag or fuzzy matching for tasks that are due today in general.

Since Taskwarrior operates with second precision, I'd suggest modifying the proposal for `eod` and others be _last second_ of the given day, that is `23:59:59`.

tbabej:
For what it's worth, I actually always felt that the default time component of `00:00:00` for all named dates is a bit rigid. There is a hook that I authored that allows you to specify default hour in the day:

https://github.com/tbabej/task.default-date-time

This would allow you to configure your default "time" for the named dates to be, for example, 16h offset. Therefore doing `task add I have to do this today due:today` would have due date today at 4pm.

vizcay:
> Hi @vizcay, thanks for the bug report. This is actually behaving [according to the documentation](https://taskwarrior.org/docs/named_dates.html).

Hi @tbabej and thanks for your reply but documentation is conflicting then here, please check https://taskwarrior.org/docs/dates.html where it states:

> eod | Current local date, with time 23:59:59.

tbabej:
Thanks for the pointer. Looks like we should treat this as a regression then. Documentation will also need to be unified on this topic.

tbabej:
Indeed, this is a regression, as 2.5.1 behaves as if `eod` is the last second of the day:

```
[warrior@tw-v2.5.1 ~]$ task calc eod
2021-07-03T23:59:59
```

smemsh:
[...]

PR #2537, opened 2021-07-18
Add filter attribute modifier 'by'

smemsh:
#### Description

Adds the modifier 'by' to those possible for filter attributes.  This uses `<=` for comparison rather than `<` as the existing `before` modifier does.  Includes docs, tests and credit.  Also added a test for #2519 which isn't technically part of this patch per se, but was the genesis and is related so it makes sense to include it in this PR.

See #2519 for full discussion and #2536 for the issue this PR closes.

```
 $ ./problems
Failed:                        

Unexpected successes:          

Skipped:                       
tw-1999.t                           1

Expected failures:             
tw-2124.t                           1
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -156,6 +156,7 @@ The following submitted code, packages or analysis, and deserve special thanks:
   Anthony VB
   Julien Rabinow
   Daniel Mowitz
+  Scott Mcdermott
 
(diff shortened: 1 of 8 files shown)


tbabej (approved):
Very thoroughly written. Thank you for contributing this :slightly_smiling_face:

Issue #2532, opened 2021-07-12
show number of tasks to be modified


jaker-dotcom:
When entering a modification that is not bound to one or multiple tasks by their ID but by a filter (e.g. project and pending) one is asked whether all tasks that the criteria applies to should be modified. Otherwise one can confirm or decline per task. I find myself oftentimes going "yes .. yes .. yes .." because there's no way to tell whether the filter was correct (as in I didn't do a mistake there). If it showed the number of all tasks that the  criteria applies to it would be an indicator for whether it's right or wrong. Drawback from having a filter that is too broad (say forgot to add +PENDING) is that you have to run the undo command several times.

tbabej:
This is a great suggestion and relatively self-contained. Putting into our next milestone.

sebu06:
I had some code that also listed the number of changes for the tasks:
```
➜  build git:(2.6.0) ✗ ./src/task pro:A done
This command will alter 3 tasks.
  - End will be set to '2021-07-18'. (3 tasks)
  - Start will be deleted. (1 tasks)
  - Status will be changed from 'pending' to 'completed'. (3 tasks)
```

However, for this I changed the return value of taskDifferences to a std::vector<std::string> and had to call taskDifferences twice. In the end I didn't like this approach and reverted it. However, if that is a wanted feature, I could push this as well.

sebu06:
implemented with PR #2538

jaker-dotcom:
Great thanks!

PR #2538, opened 2021-07-18
Added output of number of tasks to be modified in batch mode

sebu06:
#### Description

Added number of tasks as output for batch operations, see issue #2532.

Runtime:                         0.35 seconds

### DIFF
--- a/src/commands/CmdAnnotate.cpp
+++ b/src/commands/CmdAnnotate.cpp
@@ -70,6 +70,9 @@ int CmdAnnotate::execute (std::string&)
   // Accumulated project change notifications.
   std::map <std::string, std::string> projectChanges;
 
+  if(filtered.size() > 1) {
+    std::cout << "This command will alter " << format(filtered.size()) << " tasks." << std::endl;
+  }
   for (auto& task : filtered)
   {
     Task before (task);
--- a/src/commands/CmdAnnotate.cpp
+++ b/src/commands/CmdAnnotate.cpp
@@ -71,7 +71,7 @@ int CmdAnnotate::execute (std::string&)
   std::map <std::string, std::string> projectChanges;
 
   if(filtered.size() > 1) {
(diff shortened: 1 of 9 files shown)


src/commands/CmdAnnotate.cpp
```diff
@@ -70,6 +70,9 @@ int CmdAnnotate::execute (std::string&)
   // Accumulated project change notifications.
   std::map <std::string, std::string> projectChanges;
 
+  if(filtered.size() > 1) {
+    std::cout << "This command will alter " << format(filtered.size()) << " tasks." << std::endl;
```
tbabej, line 74: Taskwarrior uses [verbosity tokens](https://taskwarrior.org/docs/verbosity.html) to allow users to control the amount of output and feedback messages the `task` commands are providing.

Hence we should not output this directly to `cout`, but rather make this output appear under the appropriate verbosity token. Looking at the list the `affected` seems to be the right match here (from `man taskrc`):

```
                  affected   Reports 'N tasks affected' and similar
```

There is already a convenience function called `feedback_affected` that does the check for you:

```c++
      feedback_affected ("This command will alter {1} tasks.", count);
```

tbabej (changes requested):

src/commands/CmdAnnotate.cpp
sebu06: Hi,
thanks - as you've seen I'm new to contributing to taskwarrior and still finding my way through the codebase. I included your request in the next commit. It is very positive for me to see that there are quick responses to pull requests - this definetely increases my motivation to contribute - thank you again for that.

tbabej (approved):
Looks good to me. Thanks for the PR!

src/commands/CmdAnnotate.cpp
tbabej: No worries! I'm happy to see new contributors to the project. Feel free to shoot me an email if you have any questions :slightly_smiling_face:

Issue #2208, opened 2019-09-07
Coloring dates with scheduled tasks in calendar report


justf0rfun:
#### To request an improvement...

Taskwarrior can be used to manage appointments, like meetings, dates etc. by scheduling tasks. Unfortunately dates with scheduled tasks are not highlighted by coloring in the calendar report.

I would like to see dates where tasks has been scheduled to be colored in the calendar report. As far as I can see (I only use them rarely) due dates are already colored in the calendar report. Scheduled dates should be colored, too, maybe in a different color. 

With scheduled dates highlighted in the calendar report, taskwarrior gets a better tool for all users managing scheduled tasks like appointments with taskwarrior.

As far as I can see, this improvement does not break any existing functionality and does not interfere or only slightly with existing installations. When deployed, either this coloring could be activated for all users by default or users would have to change their configuration to enable the new coloring.

pbeckingham:
Godo suggestion, thanks.

sebu06:
implemented with PR #2539

PR #2539, opened 2021-07-18
Added coloring of scheduled dates in calendar report

sebu06:
Color can be defined by color.calendar.scheduled, see issue #2208 

Runtime:                         0.35 seconds

### DIFF
--- a/src/commands/CmdCalendar.cpp
+++ b/src/commands/CmdCalendar.cpp
@@ -290,6 +290,7 @@ int CmdCalendar::execute (std::string& output)
   Color color_overdue    (config.get ("color.calendar.overdue"));
   Color color_weekend    (config.get ("color.calendar.weekend"));
   Color color_holiday    (config.get ("color.calendar.holiday"));
+  Color color_scheduled  (config.get ("color.calendar.scheduled"));
   Color color_weeknumber (config.get ("color.calendar.weeknumber"));
 
   if (Context::getContext ().color () && config.getBoolean ("calendar.legend"))
@@ -307,6 +308,8 @@ int CmdCalendar::execute (std::string& output)
           << color_duetoday.colorize ("due-today")
           << ", "
           << color_overdue.colorize ("overdue")
+          << ", "
+          << color_scheduled.colorize ("scheduled")
           << ", ";
 
     // If colorizing holidays, print legend
@@ -528,6 +531,7 @@ std::string CmdCalendar::renderMonths (
   Color color_overdue    (config.get ("color.calendar.overdue"));
   Color color_weekend    (config.get ("color.calendar.weekend"));
   Color color_holiday    (config.get ("color.calendar.holiday"));
+  Color color_scheduled  (config.get ("color.calendar.scheduled"));
   Color color_weeknumber (config.get ("color.calendar.weeknumber"));
 
   // Loop through months to be added on this line.
@@ -573,7 +577,7 @@ std::string CmdCalendar::renderMonths (
         if (config.get ("calendar.holidays") != "none")
         {
           auto dateFormat = config.get ("dateformat.holiday");
-		  for (auto& hol : config)
+          for (auto& hol : config)
           {
             if (hol.first.substr (0, 8) == "holiday.")
             {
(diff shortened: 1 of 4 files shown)


src/commands/CmdCalendar.cpp
```diff
@@ -606,6 +610,25 @@ std::string CmdCalendar::renderMonths (
         // colorize due tasks
         if (config.get ("calendar.details") != "none")
         {
+          config.set ("schdeuled", 0);
```
tbabej, line 613: There is a typo here in `scheduled`.

sebu06: Hi,
thanks - sorry I missed that - it is fixed now.

src/commands/CmdCalendar.cpp
```diff
@@ -606,6 +610,25 @@ std::string CmdCalendar::renderMonths (
...
+              std::string due = task.get ("scheduled");
+              Datetime duedmy (strtol (due.c_str(), nullptr, 10));
+
+              if (duedmy.sameDay (date))
+              {
+                cellColor.blend(color_scheduled);
+              }
+            }
+          }
+
```
tbabej, line 631: I think it would be a bit less repetitive here if we could roll the check for scheduled attribute into the code below. In particular:

```c++
            auto status = task.getStatus ();
            if ((status == Task::pending ||
                 status == Task::waiting ) &&
                !task.hasTag ("nocal")     &&
                task.has ("due"))
```

could be altered to omit the last condition and then we could split the logic within one cycle depending on whether "due" is defined or "scheduled" is defined (or both).

sebu06: Hi,
this is obviously correct - I changed the code as youu proposed. I switched the order of the due and scheduled section, since I think if a task is due on that day it is more important then when one is scheduled. Also, I added a flag that once a due task has been found on that day, no coloring of scheduled tasks will happen.

src/commands/CmdCalendar.cpp
```diff
@@ -603,40 +607,53 @@ std::string CmdCalendar::renderMonths (
         if (today.sameDay (date))
           cellColor.blend (color_today);
 
-        // colorize due tasks
+        // colorize due and scheduled tasks
         if (config.get ("calendar.details") != "none")
         {
           config.set ("due", 0);
+          config.set ("scheduled", 0);
+          bool dueTaskFound = false;
```
tbabej, line 615: Is the purpose of this variable to prevent the "scheduled" and "overdue" colors being applied on the same date?

sebu06: yes, if the date has a task has a due date, tasks appearing later in the task array should not overwrite the due color, since I think due dates are more important to be marked in the calendar than scheduled ones

tbabej: Understood and agreed. Two suggestions/observations:
* for a decision like that (due takes precendence over scheduled when coloring), we should at least add a comment to ensure future readers of this code understand this was a deliberate decision made here :slightly_smiling_face: 
* I don't think the code works correctly, as is. The `all` vector contains all the tasks, not just the tasks that are scheduled/due on the given date. This means that while the code is evaluating the color for the, say, July 27th, it can encounter a task with due date of Dec 31st, and it will still flip the value of the `dueTaskFound` to `true`. This will in turn prevent that date from being colored with the scheduled coloring, if there is a matching task.

The flag should be flipped only after the color was definitely applied (that is, after the `cellColor.blend` calls). Also I'd suggest a slightly more descriptive variable name, perhaps `coloredWithDue`?

sebu06: Thanks, that is (maybe) correct. I still do not completely understand the code, I think:

```
              if(task.has("due")) {
                std::string due = task.get ("due");
                Datetime duedmy (strtol (due.c_str(), nullptr, 10));

                if (duedmy.sameDay (date))
                {
                  switch (task.getDateState ("due"))
                  {
                  case Task::dateNotDue:
                    break;
                    ...
```

If `duedmy.sameDay (date)` evaluates to true, how could the case Task::dateNotDue happen? I thought this should not happen and thus It shouuld have been fine to set the value of the coloredWithDue before the whole switch section, right?

tbabej: You are correct that the case `Task::dateNotDue` will never happen, I believe that is there for completeness (or perhaps compiler nagging).

The previous version of the code, however, did not enforce that the flag was flipped only if the task is due on the same day, it was flipped if the task had a due attribute defined (with any value):

image

That was incorrect.

However, you're right that we could move the `coloredWithDue = true;` statement in front of the switch statement, which would simplify the code a bit. Good suggestion, I missed that :slightly_smiling_face:

tbabej: @sebu06 up to you if you want to refactor, I'm also happy merging as is (after the Changelog conflict is resolved).

tbabej (approved):
Looks good on my end. Thanks for following up with the review!

Issue #1921, opened 2018-02-14
[TW-1908] STDERR potentially abused for status messages while running 'task sync'


taskwarrior:
_rjc on 2017-04-19T04:34:00Z says:_ 

 Hi,

Task is potentially abusing STDERR for some of its status messages.
This issue is easily observed by running "task sync" from cron(8). Namely, when I run it from said cron(8) I would very much like to get an email only when there's a real issue running the above command. Obviously, I can just redirect STDOUT to /dev/null:

    task sync > /dev/null

However, I still get this message on STDERR output:

    Sync successful.  No changes.

OK, let's try the other way around - redirect STDERR to /dev/null:

    task sync 2> /dev/null

The problem with that approach is that I get another message, this time to STDOUT:

    Syncing with my.task.server:53589

Sure, I can redirect *both* to /dev/null like so:

    task sync > /dev/null 2>&1

and that's what I'm currently using but, in case there's a real error (i.e. error code != 0), I won't see it as I already redirected *both* STDOUT and STDERR to /dev/null, to merely silence the status messages of 'task sync'.

Currently I'm running something like this:

   task sync > /dev/null 2>&1 || echo "Something went wrong - run 'task sync' again to see what it was."

I understand that STDERR was probably chosen for a reason but it feels like one needs to jump through quite a number of hoops to get 'task sync' working from cron(8) without emails for every single sync *and* with emails *only* when there's a failure.
It's never too late to change it! :^)

Regards,

Raf

taskwarrior:
_rjc on 2017-10-12T03:39:54Z says:_ 

 Not sure why I would have marked it as _Major_ - must've selected it looking at the initial letter and thinking of _Minor_ as per your system:
 * Minor - Minor loss of function, or other problem where easy workaround is present.
 * Trivial - Cosmetic problem like misspelt words or misaligned text.

Certainly *not* _Trivial_.

In terms of running *task sync* from *cron* - I have *taskwarrior* set up on multiple machines so, if I'm working on another one where tasks are out of sync, I'd have to run *task sync* manually before doing anything anyway. How is running it from *cron* any different and a _bad idea_? The Android app does it automatically, too. Is there a recommended way of achieving the same thing?

taskwarrior:
_Paul Beckingham on 2017-12-02T19:31:49Z says:_ 

[...]

PR #2540, opened 2021-07-18
add convenience feedback funtion for sync messages

sebu06:
#### Description

I added a convenience function to output feedback from the sync command, which outputs to stdout instead of stderr, see issue #1921 . Howevver, I'm wondering if it is correct that footnotes are output to std::cerr

Context.cpp:674
```
    if (verbose ("footnote"))
    {
      for (auto& f : footnotes)
        if (color ())
          std::cerr << colorizeFootnote (f) << '\n';
        else
          std::cerr << f << '\n';
    }
```
and if for some reason the sync output should be treated as a footnote rather than usual output. In that case, I would simply check the verbosity token before the output section.

Runtime:                         0.42 seconds

### DIFF
--- a/src/commands/CmdSync.cpp
+++ b/src/commands/CmdSync.cpp
@@ -32,6 +32,7 @@
 #include <Context.h>
 #include <Filter.h>
 #include <Color.h>
+#include <main.h>
 #include <shared.h>
 #include <format.h>
 #include <util.h>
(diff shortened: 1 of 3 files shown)


sebu06:
@tbabej - could you maybe give me a hint why footnotes are printed to std::cerr instaad of std::cout or if this fix is good as it is?

djmitche:
@sebu06 would you like to rebase this against `develop`? The sync function has changed, but this helper would be useful all the same.

Issue #2136, opened 2019-03-31
Support config overrides in context configuration


thehunmonkgroup:
#### To request a feature...

* Clearly describe the feature.

Context configuration in ```.taskrc``` doesn't respect ```rc.XXXX``` overrides.

* Clearly state the use case. We are only interested in use cases, do not waste time with implementation details or suggested syntax.

Contexts could benefit from being able to support custom sorting on reports for that particular context.

It would be great if this worked:
```context.biz=project:Biz rc.report.next.sort:project+/,urgency-```

**Comment by tbabej** [comment]
Will be implemented as `context.<name>.config` configuration variable:

```
context.biz.config=report.next.sort:project+/,urgency-
```

PR #2543, opened 2021-07-19
Allow overriding configuration variables per context

sebu06:
#### Description

I would like to run a specific report for each context. I added a convenience function to libshared - see https://github.com/GothenburgBitFactory/libshared/pull/48

Passed:                          3810
Skipped:                           13
Expected failures:                 13
Runtime:                        12.02 seconds


tbabej:
I like the direction where this is going, but this is needlessly too specific. After this feature is released, users could rightfully ask for the option to override more configuration variables in a specific context.

Instead, we should go for a holistic solution which would allows us to override an arbitrary configuration variable. In 2.6.0, the context is already expanded to support `read` and `write` subvariables (this was implemented in #2402 )

```
context.<name>.read=...
context.<name>.write=...
```

we should also allow users to set a `config` subvariable

```
context.<name>.config="rc.default.command:done"
```

these could be injected into the parser, if defined.

sebu06:
If you think it is too specific, the option would be to simply change the `get` method in libshared - since I was a little bit afraid to mess things up, I created a new one named `getFromContext`. The basic idea was that in case a context is set, all config options can be overwritten in the following style:
```
context.<name>.<configoption>=<value>
```
where 
`<configoption>` is the original config option like for example `default.command`. The config would looks like this then:
```
[...]
default.command=all_report
context.home.default.command=home_report
context.work.default.command=work_report
context.work.color.uda.priority.H=rgb542
context.work.report.next.labels=ID,Name
[...]
```

tbabej:
I like that. I'd propose we move all that under `rc` namespace, like this:

```
[...]
default.command=all_report
context.home.rc.default.command=home_report
context.work.rc.default.command=work_report
context.work.rc.color.uda.priority.H=rgb542
context.work.rc.report.next.labels=ID,Name
[...]
```

the second open question is whether to treat all these as configuration overrides (that is, should the footnote info about the override appear?)

tbabej (changes requested):
Looks and works well! The only thing remaining to fix is that the ChangeLog entry is out of date - now setting any context-specific configuration value overrides is allowed.

sebu06:
I updated the Changelog and added a comment to the man page

tbabej (approved):

tbabej:
Closes #2136.

PR #2544, opened 2021-07-23
CmdInfo: fix missing spaces causing run-on in virtual tag list

smemsh:
#### Description

This patch fixes a cosmetic issue that would cause run-on tag names to be shown in the output when viewing a task and displaying its list of virtual tags.  Some lines were missing spaces, because their virtual tag definitions did not include a space like all the rest, and these spaces were relied on for output formatting.

> Have you run the test suite?

Have not run the test suite; presumably this is a cosmetic change.  Did test the fix itself and it does make the run-on tags disappear.


tbabej:
Thanks for catching that @smemsh !

tbabej (approved):
Works as expected.

PR #2555, opened 2021-08-02
Refactor to store tags as individual attributes

djmitche:
Each tag is stored as `tag_<tagname>: x`.  The `x` is required because
empty attributes are treated as nonexistent.

For compatibility, the `tags` attribute is updated in sync with the 
per-tag attributes.  This compatibility support may be dropped in later
versions.

Note that synchronization _updates_ use JSON format, which does not 
change with this patch, and thus no compatibility issues exist.  The 
synchronization _initialization_, however, uses FF4, meaning that a
sync server initialized from a version of `task` with this patch will
contain `tag_<tagname>` attributes, which will look like orphaned UDAs
to older versions.  However, as updates to tasks are synchronized via 
the sync server, the updates will not contain these attributes and they
will show as "deleted" in the `task info` display on the older version.
Aside from the noise in the `task info` output, this is harmless.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -630,6 +630,15 @@ void Task::parse (const std::string& input)
     parseLegacy (input);
   }
 
+  // for compatibility, include all tags in `tags` as `tag_..` attributes
+  if (data.find ("tags") != data.end ()) {
+    for (auto& tag : split(data["tags"], ',')) {
+      data[tag2Attr(tag)] = "x";
+    }
+  }
+  // ..and similarly, update `tags` to match the `tag_..` attributes
+  fixTagsAttribute();
+
   recalc_urgency = true;
 }
 
@@ -691,16 +700,6 @@ void Task::parseJSON (const json::object* root_obj)
           auto tag = (json::string*)t;
           addTag (tag->_data);
         }
-      }
-      // This is a temporary measure to accomodate a malformed JSON message from
-      // Mirakel sync.
-      //
-      // 2016-02-21 Mirakel dropped sync support in late 2015. This can be
-      //            removed in a later release.
-      else if (i.first == "tags" && i.second->type() == json::j_string)
-      {
-        auto tag = (json::string*)i.second;
-        addTag (tag->_data);
       }
 
       // Dependencies can be exported as an array of strings.
@@ -906,6 +905,10 @@ std::string Task::composeJSON (bool decorate /*= false*/) const
     if (! i.first.compare (0, 11, "annotation_", 11))
       continue;
 
+    if (i.first == "tags" || isTagAttr (i.first))
+      // Tags are handled below
+      continue;
+
     // If value is an empty string, do not ever output it
     if (i.second == "")
         continue;
@@ -944,26 +947,6 @@ std::string Task::composeJSON (bool decorate /*= false*/) const
           << "\":"
           << i.second;
 
-      ++attributes_written;
-    }
-
-    // Tags are converted to an array.
-    else if (i.first == "tags")
-    {
-      auto tags = split (i.second, ',');
-
-      out << "\"tags\":[";
-
-      int count = 0;
-      for (const auto& i : tags)
-      {
-        if (count++)
-          out << ',';
-
-        out << '"' << i << '"';
-      }
-
-      out << ']';
       ++attributes_written;
     }
 
@@ -1045,6 +1028,25 @@ std::string Task::composeJSON (bool decorate /*= false*/) const
     }
 
     out << ']';
+  }
+
+  auto tags = getTags();
+  if (tags.size() > 0)
+  {
+    out << ','
+        << "\"tags\":[";
+
+    int count = 0;
+    for (const auto& tag : tags)
+    {
+      if (count++)
+        out << ',';
+
+      out << '"' << tag << '"';
+    }
+
+    out << ']';
+    ++attributes_written;
   }
 
 #ifdef PRODUCT_TASKWARRIOR
@@ -1257,8 +1259,13 @@ std::vector <Task> Task::getDependencyTasks () const
 ////////////////////////////////////////////////////////////////////////////////
 int Task::getTagCount () const
 {
-  auto tags = split (get ("tags"), ',');
-  return (int) tags.size ();
+  auto count = 0;
+  for (auto& attr : data) {
+    if (isTagAttr (attr.first)) {
+      count++;
+    }
+  }
+  return count;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


djmitche:
This is *almost* there, at least as far as the test suite is concerned.

The tricky bit here seems to be around getting and setting `tags` as a user-level property:
```
task 1 modify tags:foo,bar
task 1 _get tags
task 2 modify tags:1.tags
```

The first was pretty easy to fix, and I suppose could be kept as useful syntax.

For the moment, the second works because we're reflecting `tags` into the new format for compatibility.  It's useful for tests.  We could probably special-case it in `_get`, or just deprecate that with some notice (I assume `_get` is internal and not part of the API contract?)

The third one was kind of tricky.  It seems that `1.tags` is a DOM expression which evaluates (again for compatibility) to a comma-separated string full of tags.  But that's not parsed?  The parsing in `src/columns/ColTags.cpp` was doing some funny stuff (parsing the whole expression within a loop over parts of the expression) so I refactored it a bit.

In general, what do you think we should do here?  Try to "emulate" a comma-separated DOM tags?  Or drop tags from the DOM?

djmitche:
I'll make another PR for depends, once this one is done.

djmitche:
Looks like I need to look at
```
not ok 13 - import.t: Test import same task twice
```
but I'd like some general feedback first..

djmitche:
OK, that test should be fixed.

djmitche:
Hm, in trying to test sync'ing this between old and new versions, I'm not seeing the `tag_foo: "x"` attributes on the old version.

new:
```
funk ~/p/taskwarrior [issue2553] $ cat ~/.task/pending.data 
[description:"one" entry:"1627853804" modified:"1627950549" start:"1627864360" status:"pending" tags:"c,e,foo,fromnew,fromold" tags_c:"x" tags_e:"x" tags_foo:"x" tags_fromnew:"x" tags_fromold:"x" uuid:"05346776-c59b-41ae-a386-9b329bd53fab"]
```
old:
```
cleave ~ $ cat tmp/task/pending.data 
[description:"one" entry:"1627853804" modified:"1627950549" start:"1627864360" status:"pending" tags:"c,e,foo,fromnew,fromold" uuid:"05346776-c59b-41ae-a386-9b329bd53fab"]
```

I would expect the older version to just treat the `tag_foo` as an unknown UDA (and in fact the new version does, too, in the `info` output).  Hmm.

djmitche:
Huh, the taskserver is getting JSON data, straight from `backlog.data`:
```
9616771a-0d9e-49ca-8e2f-19b830839f1c
{"description":"one","entry":"20210801T213644Z","modified":"20210803T003752Z","start":"20210802T003240Z","status":"pending","uuid":"05346776-c59b-41ae-a386-9b329bd53fab","tags":["baz","blergh","c","e","foo","fromnew","fromold"]}
{"description":"one","entry":"20210801T213644Z","modified":"20210803T004028Z","start":"20210802T003240Z","status":"pending","uuid":"05346776-c59b-41ae-a386-9b329bd53fab","tags":["baz","blergh","c","e","foo","fromnew","fromold","yikes"]}
```
so, these tags never even show up there.  It looks like they *were* copied there on the `sync init`, but deleted in subsequent modifications.

I think this means it all just "works" as far as this inter-version compatibility stuff goes.

tbabej:
I still need to test this, but from reading the code earlier today the implementation seemed OK. One thing that I think is missing is handling the `tags_X` attributes in `task info`. I suspect they might show up individually on the bottom if they're being interpreted as UDAs?

djmitche:
> I still need to test this, but from reading the code earlier today the implementation seemed OK. One thing that I think is missing is handling the `tags_X` attributes in `task info`. I suspect they might show up individually on the bottom if they're being interpreted as UDAs?

I spotted this and handled it.

tbabej (approved):
Works as expected in my testing. I think the only outstanding (related, but not part of this PR) issue is that for an older `tags` attribute, we will not convert it into `tags_<name>` attribute array. Perhaps that's something that a `task upgrade` could take care of.

djmitche:
That should already be handled - the parsing reads both tags and tags_foo
and updates the in memory data structures correctly.  Did you see otherwise?

tbabej:
Well, it does - but the current behavior is that the in-memory data structures get reflected on the disk:

```
[warrior@tw-2.6.0 ~]$ task add hello +test
Created task 1.
[warrior@tw-2.6.0 ~]$ cat ~/.task/pending.data 
[description:"hello" entry:"1628292231" modified:"1628292231" status:"pending" tags:"test" tags_test:"x" uuid:"8472ba6a-149c-4fc9-b9a9-be8d73aa796f"]
[warrior@tw-2.6.0 ~]$ task 1 mod +test2
Modifying task 1 'hello'.
Modified 1 task.
[warrior@tw-2.6.0 ~]$ cat ~/.task/pending.data 
[description:"hello" entry:"1628292231" modified:"1628292250" status:"pending" tags:"test,test2" tags_test:"x" tags_test2:"x" uuid:"8472ba6a-149c-4fc9-b9a9-be8d73aa796f"]
```

Notice that both `tags` and `tags_X` attributes are present. This is a clean DB, so no historical data present.

While we can get this fixed easily, I'm thinking perhaps we should not be merging these data-model related changes to 2.6.0 at this stage of the branch. I am trying to get 2.6.0 out of the door, and as such these patches might seem limited (manual) testing. One alternative would be to create a new branch (perhaps labeled as `3.0`) for the data-model changes related OT work. Though that would increase a bit the development inconvenience (I offer take care of the rebasing :slightly_smiling_face: ), it seems like the right move here.

Some parts of the changes could still go into 2.6.0, for example changing the behaviour of `wait` and the unwaiting mechanism. Thoughts?

djmitche:
I'm not sure what issue you're seeing here.  The double-storage is intentional, to avoid the need for any explicit migration step.  It means that a user can smoothly upgrade a single replica from 2.5.3 to 2.6.0 without issue, and in fact can connect 2.5.3 and 2.6.0 replicas via taskd with only minimal issues (specifically, a tag removed with 2.5.3 will still appear after a sync to 2.6.0).  Using both 2.5.3 and 2.6.0 on the same replica would have the same minimal issue (but I think we should advise upgrading a replica to any new version in one go, rather than accessing the same task_dir with both versions concurrently).

You're right that we could "fix" the double-storage easily, but I think that would reduce usability.

As for getting 2.6.0 out, I think that it's fine to hold off merging this for a little while (a week or two) in the interest of ensuring quality for what's on the branch already.  Aside from that, I would argue that a project at this level of development bandwidth is not a good fit for multiple active branches and lots of inter-branch merging - that's a lot of toil for someone (you) and a lot of coordination of a loosely-affiliated group of part-time devs.  I think this is a better fit for an "always shippable `main`" approach, with some thoughtful attention to when `main` actually gets shipped and maybe delaying some PRs as necessary, such as to get a patch-level release out before a major release.

I suspect that the 3.0 breaking point will come when we drop support for taskd and add support for the new sync backend.  Until then, I think the changes are non-breaking, modulo some suboptimal behaviors when syncing between versions (and in my experience, syncing between any of the recent versions of TW has been a source of data corruption, so suboptimality seems acceptable).

PR #2557, opened 2021-08-05
use emplace

neheb:
Allows removing the constructor name as emplace forwards the arguments
directly.


tbabej (changes requested):
It looks like the changes to the `libshared` submodule might be unrelated here and cause the build to fail.

neheb:
huh. wonder why they're here. will fix.

tbabej (approved):
Looks good. Thanks for fixing the libshared problem :slightly_smiling_face:

PR #2562, opened 2021-08-07
Replace status:waiting with +WAITING

djmitche:
With this change, WAITING becomes only a virtual tag, dynamically
determined based on a task's `wait` property and the current time.  This
avoids the need to modify ("un-wait") a task, something which previously
occured during gc and thus could happen on multiple replicas or might
not occur when the user expected.

The user-visible change is that filters using `status:pending` which
assume waiting tasks are not included will need to be modified to
include `-WAITING`.


tbabej:
This looks like a well rounded MR! The change in behaviour of filtering borth for `status:waiting` and `status:pending` is a bit worrying consequence. I wonder if we could avoid breaking third-party tools and user configuration by allowing `status` to have virtual values:

```
@@ -301,12 +301,22 @@ Task::status Task::getStatus () const
   if (! has ("status"))
     return Task::pending;
 
+  if has ("wait") {
+    Datetime now;
+    if (get_date ("wait") > now)
+      return Task::waiting;
+  }
+
   return textToStatus (get ("status"));
 }
 
 void Task::setStatus (Task::status status)
 {
+  // Waiting tasks are considered pending on disk
+  if (status == Task::waiting)
+    status = Task::pending;
+
   set ("status", statusToText (status));
 
   recalc_urgency = true;
@@ -1615,12 +1625,6 @@ void Task::validate (bool applyDefault /* = true */)
   }
 */
 
-  // Tasks with a wait: date get a special status.
-  else if (status == Task::pending &&
-           has ("wait")            &&
-           get ("wait") != "")
-    status = Task::waiting;
-
```

and then making sure that the rest of the code base does not directly interact with the `status` attribute if not necessary (which seems to _mostly_ be the case already).

Assuming you haven't tried that approach, I can take a stab at prototyping and seeing if this leads anywhere.

djmitche:
I feel like that's complicating the definition of "status" for the sake of a relatively minor change in behavior.  I think this could be handled with a well-crafted release note?  Especially if that's in 3.0.0.

tbabej:
It all comes down to the definition of individual states should be. The original separation of `status:pending` and `status:waiting` implies that "pending" tasks are the tasks that are actionable.

This was introduced in 1.8.0 (released in 2009), however, later changes do not conform to this definition. For example, a "blocked" task (=a task with non-empty `depends` attribute) does not get a designated status, while still being arguably non-actionable (on its own). This is likely due to the fact that mapping all possible values of properties of the task onto single status attribute does not scale, as task can be for example both waiting and blocked (so what would be the value of the `status` attribute?).

Hence this line of thinking supports the argument for eventual removal of the `waiting` status altogether. That said, it does feel like a change with non-trivial impact, and in some cases the user might not be even empowered to change the filter definition (i.e. if using third party tools that were not updated yet). This makes me want to follow a bit more careful deprecation schedule.

You have already hinted at 3.0.0 being the right place to land the removal of the `waiting` status, and I agree. What I think we could do in 2.6.0 is to:
* mark `status:waiting` as deprecated (suggesting use of `+WAITING` or `wait.after:now` instead)
* implement virtual values for the status:waiting, so that they are derived from the `wait` attribute comparison
* remove the "unwaiting" mechanism in the GC

This would be a bit more work, but would have the advantage of enabling users and third-party a smooth transition path, as in 2.6.0+ both "old-style" `status:waiting` filters and "new-style" `+WAITING` / `wait.after:now` would work.

djmitche:
That seems reasonable!

djmitche:
Hm, the tricky bit is that dom references like `status:pending` apply to the attribute value, not `Task::getStatus`.  Should I special-case it in `Task::get`?

djmitche:
My attempt is [here](https://github.com/djmitche/taskwarrior/pull/new/issue2554-b).  If you add `-WAITING` to the definition of `ls` in `test/wait.t`, it will pass.

tbabej:
> Hm, the tricky bit is that dom references like status:pending apply to the attribute value, not Task::getStatus. Should I special-case it in Task::get?

I think instead of doing that, we should try to have them use `getStatus` where applicable?

> My attempt is here. 

Nice. I also spent some time expanding the prototype about and thinking about this migration. What do you think about pushing that branch to the TW repo so that we could both push commits there?

djmitche:
done!

tbabej:
Alright! Let's open another PR so that we can keep discussion related to those changes separate from this one :slightly_smiling_face:

tbabej:
Want to do the honors? I don't want to steal credit :sweat_smile:

djmitche:
#2563 is the better solution.

PR #2563, opened 2021-08-08
Make status:waiting a "virtual" status

### LINKED ISSUE #1918: [TW-1905] GC un-waits tasks, which bypasses on-modify hook processing
_Paul Beckingham on 2017-04-02T17:34:27Z says:_ 

 When a task un-waits, it occurs in TDB2::gc, and bypasses hooks.  Instead, the on-modify hook should be called, simply because removing the wait date is a modification.

### LINKED ISSUE #2554: Remove the `waiting` state, and consider any task with wait>now to be waiting
This would mean that all waiting task have status:pending, but are treated the same way by the filters.

The advantage to this is that the task need not be modified when its waiting time completes (which appears to happen in `gc`?) -- it just begins being included in filters.

djmitche:
Fixes #2554, see also #2562.

Also closes #1918.


tbabej:
Things to double-check (a comment for myself):
* [ ] The "upgrade" procedure works as expected (status "waiting" is converted to "pending" on the disk)
* [ ] Urgency computation
* [ ] Task is "unwaited" without modifying the data files
* [ ] The `info` command produces expected output (status field, waiting field, virtual tags)

(Irony of using a TODO list for a TODO list application)

src/Task.cpp
```diff
@@ -559,6 +574,23 @@ bool Task::is_overdue () const
 }
 #endif
 
+////////////////////////////////////////////////////////////////////////////////
+bool Task::is_waiting () const
+{
+  // note that is_waiting can return true for tasks in an actual status other
+  // than pending; in this case +WAITING will be set but the status will not be
+  // "waiting"
```
tbabej, line 582: Useful comment! This is indeed a decision that has to be made. No matter which way we go here, I think two things should be consistent: the applicability of `+WAITING` and removal of `wait` attribute on `task done/delete`

In other words, if the +WAITING is going to be allowed to be applied for tasks that are `status:completed`, then the `wait` attribute should not be removed on `task done` or `task delete`. Or conversly we keep removal of the `wait` attribute on `task done/delete`, but then `+WAITING` should only apply to pending tasks.

src/TDB2.cpp
```diff
@@ -355,23 +355,6 @@ void TF2::load_gc (Task& task)
...
-      // Unwaiting pending tasks is the only case not caught by the size()
-      // checks in TDB2::gc(), so we need to signal it here.
-      Context::getContext ().tdb2.pending._dirty = true;
-
-      if (Context::getContext ().verbose ("unwait"))
-        Context::getContext ().footnote (format ("Un-waiting task {1} '{2}'", task.id, task.get ("description")));
-    }
-
-    Context::getContext ().tdb2.pending._tasks.push_back (task);
-  }
```
tbabej, line 374: By completely removing this, the tasks that were already in the `status:waiting` state before the upgrade (expired or not) will be moved to `completed.data`:

```
[warrior@pr-2563 ~]$ rm -rf ~/.task
[warrior@pr-2563 ~]$ task add Actionable task wait:yesterday
Created task 1.
[warrior@pr-2563 ~]$ sed -i 's/pending/waiting/' ~/.task/pending.data 
[warrior@pr-2563 ~]$ task all

ID St UUID     Age  Description    
 1 P  9b38093e 1min Actionable task

1 task
[warrior@pr-2563 ~]$ cat ~/.task/pending.data 
[warrior@pr-2563 ~]$ cat ~/.task/completed.data 
[description:"Actionable task" entry:"1628393350" modified:"1628393350" status:"waiting" uuid:"9b38093e-3eef-42a0-84d3-7c811595ffbe" wait:"1628294400"]
```

We need a slight upgrade procedure here. I'll implement this.

src/Task.cpp
djmitche: That makes sense.  I think minimizing the intertwining of the two (so, tasks with any status can be +WAITING) is the simplest to understand.

src/TDB2.cpp
djmitche: Good catch :)

src/Task.cpp
```diff
@@ -301,12 +303,25 @@ Task::status Task::getStatus () const
...
     return Task::pending;
 
-  return textToStatus (get ("status"));
+  auto status = textToStatus (get ("status"));
+
+  // implement the "virtual" Task::waiting status, which is not stored on-disk
+  // but is defined as a pending task with a `wait` attribute in the future.
+  if (status == Task::pending && is_waiting ()) {
+      return Task::waiting;
+  }
```
djmitche, line 312: Should this have a deprecation comment on iit?  (`2.6.0; remove this support in 3.0.0` or something)

src/TDB2.cpp
```diff
@@ -355,6 +355,14 @@ void TF2::load_gc (Task& task)
   {
     Context::getContext ().tdb2.pending._tasks.push_back (task);
   }
+  // 2.6.0: Waiting status is deprecated. Convert to pending to upgrade status
+  // field value in the data files.
+  else if (status == "waiting")
```
djmitche, line 360: I like this, as it will modify all tasks on the next `gc`, rather than waiting until each individual task is modified.

tbabej: Indeed, eager upgrade procedure :slightly_smiling_face:

src/Task.cpp
tbabej: Agreed. Let's get rid of `wait` removal in the `done/delete` then.

tbabej: Implemented in d999005.

src/Task.cpp
tbabej: Good point. Added in https://github.com/GothenburgBitFactory/taskwarrior/pull/2563/commits/00682326526bb54139472007991931a93313a452

tbabej:
@djmitche Ready for re-review if you could spare another look :slightly_smiling_face: One additional thing to consider is whether we shouldn't also port the `Context.cpp` changes for default report settings that change `status:waiting` to `+WAITING` from #2562 since:
a) the change should be have no effect under current implementation
b) less people will get stuck with the deprecated `status:waiting` in their custom reports (which are often customized by starting with the output of `task show`), which will make it easier to migrate in 3.0

djmitche:
That's a good point -- then new users won't see any subtle behavior changes when we drop `status:pending`.

src/Task.cpp
```diff
@@ -1344,7 +1344,7 @@ bool Task::hasTag (const std::string& tag) const
     if (tag == "PARENT")    return has ("mask") || has ("last");       // 2017-01-07: Deprecated in 2.6.0
     if (tag == "TEMPLATE")  return has ("last") || has ("mask");
     if (tag == "WAITING")   return is_waiting ();
-    if (tag == "PENDING")   return get ("status") == "pending";
+    if (tag == "PENDING")   return getStatus () == Task::pending;
     if (tag == "COMPLETED") return get ("status") == "completed";
     if (tag == "DELETED")   return get ("status") == "deleted";
```
djmitche, line 1350: Should these lines be similarly modified?

djmitche (commented):
Looks good!

src/Task.cpp
tbabej: I think you're correct here. It will be more consistent at the very least. Pushed 48b3ebfa577ac552ea79da246cbcf9db788d036c to address this.

tbabej:
OK, the only outstanding thing is to port the configuration default changes from #2562 and perhaps also the man page adjustments, then we should be ready to merge.

@djmitche feel free to port over if you find some spare time. Alternatively I will be available again in the evening and can push this over the finish line :rocket:

djmitche:
Done!

tbabej:
@djmitche Thanks for pushing through with this! Really happy where this ended up

djmitche:
Me too, thanks!

Issue #2572, opened 2021-08-15
taskDifferences needs to handle `tag_` attributes properly


djmitche:
Longer-term, when we represent task changes as operations, this function can just describe the queued operations.

Until then, it looks at the attributes in tasks, and is currently not doing the right thing for `tag_`.  #2569 will leave the same problem for `dep_`.

djmitche:
`taskInfoDifferences` too

PR #2573, opened 2021-08-15
Refactor to store dependencies as individual attributes

djmitche:
This also drops support for the transitional `json.depends.array` configuration value, which has not been necessary since ~2016.

As with tags, dependencies are stored in both a "combined", comma-separated format (for compatibility) and in an attribute-per-dependency format (for the future).

The first commit in this PR takes pains to prevent direct access to the `depends` attribute, although two places will be handled in #2572.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -1259,6 +1259,15 @@ void Task::removeDependency (int id)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+bool Task::hasDependency (const std::string& uuid) const
+{
+  auto deps = split (get ("depends"), ',');
+
+  auto i = std::find (deps.begin (), deps.end (), uuid);
+  return i != deps.end ();
+}
+
+////////////////////////////////////////////////////////////////////////////////
 std::vector <int> Task::getDependencyIDs () const
 {
   std::vector <int> all;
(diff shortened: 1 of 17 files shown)


src/commands/CmdEdit.cpp
```diff
@@ -667,7 +667,8 @@ void CmdEdit::parseTask (Task& task, const std::string& after, const std::string
   value = findValue (after, "\n  Dependencies:");
   auto dependencies = split (value, ',');
 
-  task.remove ("depends");
+  for (auto& dep : task.getDependencyUUIDs ())
+    task.removeDependency (dep);
```
tbabej, line 671: This works fine for now, but I wonder if we want to put more sophisticated logic in place. Removing all dependencies and then re-adding some of them will potentially generate a lot of operations that cancel each other out (removal and addition of the same `dep_X` attribute).

djmitche: Yeah, I think the "edit" functionality will need a good going-over when that happens.

src/Task.cpp
```diff
@@ -1203,23 +1195,16 @@ void Task::addDependency (const std::string& uuid)
...
-      return;
-    }
+  if (hasDependency (uuid))
+  {
+    Context::getContext ().footnote (format ("Task {1} already depends on task {2}.", get ("uuid"), uuid));
+    return;
   }
-  else
-    set ("depends", uuid);
+#endif
```
tbabej, line 1204: To reflect the previous version exactly, I think only the footnote call should be within the controlled group of the `#ifdef` block. This could cause skipping over the return statement if `PRODUCT_TASKWARRIOR` macro is set to 0 (which is for current implementation of `taskd`).

tbabej (commented):
Done with the review (sorry for stretching it out). Functionally this works well, we just need to adress those minor points above and then it can merge  :)

src/Task.cpp
djmitche: I had assumed that `PRODUCT_TASKWARRIOR` was un-set for timewarrior. Why is `taskd` using this at all?  And, does `taskd` print footnotes?

tbabej: The current implementation of the `taskd` daemon utilizes `Task` objects, for example while performing [on-server merging](https://github.com/GothenburgBitFactory/taskserver/blob/1.2.0/src/daemon.cpp) of updates.

Timewarrior and taskwarrior only share common base in the `libshared` repo.

I think these `#ifdef` controlled groups (I've seen them elsewhere as well) are used exactly to suppress the output on the `taskd` daemon which would stem from merging task updates triggering these code paths.

djmitche: Ah, I had the sense backward.  And I learned something about PRODUCT_TASKWARRIOR :)

djmitche:
OK, I applied all of your recommendations and rebased to fix the ChangeLog conflict.  If this is green, let's click merge :)

tbabej (approved):

PR #2595, opened 2021-09-04
meta: Expand review build target to allow specifying PR reference for libshared

tbabej:


tbabej:
@liskin See here. Neither
* test matrix for the source branch: https://github.com/GothenburgBitFactory/taskwarrior/blob/6d00b5ff919db59f9b604815f79e01a6fa9289d0/.github/workflows/tests.yaml#L7
* test matrix for the target branch: https://github.com/GothenburgBitFactory/taskwarrior/blob/2bd0d8ebd350c13bff1849d99b7a817c8ff1c951/.github/workflows/tests.yaml#L7

contain the two stalled status checks on the bottom
image

My hypotheses are that perhaps `pull-request` takes its definition from the `master` branch, or is somehow frozen at the time of creation, both of which don't make much sense to me (2.6.0 is the default branch for this repository).

tbabej:
This is now green thanks to Tomas' help :rocket:

PR #2598, opened 2021-09-13
fix a typo in the help commnad

Jasha10:
#### Description

Previously the `task help` command gave this output:
```text
Alternately algebraic expressions support:
  and  or  xor            Logical operators
  <  <=  =  !=  >=  >     Relational operators
  (  )                    Precedence

  task due.before:eom priority.not:L   list
  task '(due < eom or priority != L)'  list
```
The last two lines are not equivalent; this PR changes the word `or` in the last line to `and`:
```text
  task due.before:eom priority.not:L   list
  task '(due < eom and priority != L)'  list
```


tbabej (approved):
Not sure if this was a typo, but I guess it does make a bit more sense for those two lines to be functionally equivalent :)

PR #2668, opened 2021-11-06
Fix typo

jwilk:
higlights → hig**h**lights


tbabej (approved):
I'm stumped about how could this have happened three times :astonished:  Thanks for noticing and fixing!

Issue #2507, opened 2021-06-17
Explore an OT-based sync approach


djmitche:
[Operational Transformation](https://en.wikipedia.org/wiki/Operational_transformation) is an approach for synchronizing multiple replicas of the same data in an eventually-consistent model, suitable for cases where there is a central location that all replicas can refer to, and where most modifications do not conflict with one another.  Personal task data fits these conditions perfectly.

This issue is an early exploration of applying OT to synchronize taskwarrior instances, as a wholesale replacement for the current taskd-based synchronization mechanism, which has a number of problems.

TaskChampion has already shown that OT can be successful. The remaining exploration is to determine what changes are required to the TW data model to support this, and to consider approaches for garbage collection, data security, etc.

## Subtasks

* https://github.com/GothenburgBitFactory/tw.org/pull/196
* #2553
* #2554 
* #2569 
* https://github.com/GothenburgBitFactory/tw.org/issues/212
* #2596 (issue was somehow deleted..)
* #2681

djmitche:
I'll be working on this.

tbabej:
According to the linked Wikipedia page, the OT approach could be implemented using various OT control algorithms with slightly different tradeoffs. What is your current thinking on the proper choice there?

djmitche:
(would you mind assigning this to me so that it shows up in my TaskWarrior? :) )

tbabej:
Done!

djmitche:
The [models](https://en.wikipedia.org/wiki/Operational_transformation#Consistency_models) are basically restrictions on the design of the transformation functions.  CC is the most basic, and CCI adds "Intention", which becomes relevant for edge cases like transforming a change which deletes an entire paragraph with a change that inserts a new sentence into that paragraph.  Happily, with a well-defined task model (and this might require some changes to the TW data model), CC is plenty.

First, most operations do not conflict, either because they apply to different tasks, or because they apply to different properties of the same task (e.g., adding a tag and marking a task as started).  So, transposing those operations is simply a matter of swapping them without change.

For operations that do apply to the same property of the same task, we have a timestamp on the operation.  We also make the general presumption that operations are created by one person who intended them to occur in the order they were created.  So we can simply select the later of the two operations and discard the other.  When the timestamps are identical, any total ordering is a good fallback.

[...]

Issue #2683, opened 2021-11-18
Declare `contextTask` in one place


djmitche:
A number of column type implementations have `extern Task& contextTask`.  In general, this seems to be a poorly-guarded global variable that is declared and referred to in a number of places.

Let's clean that up:
 * declare it in one place
 * gate access to it through a function, in some way that it is reset to NULL at an appropriate time to avoid dangling references

djmitche:
This is even a little worse: `contextTask` is set in column modify methods, but those methods are also called for `task add`.

tbabej:
It looks the reason the `contextTask` is being set there is for resolution of DOM references, like this:

```
task add due:2.due This task should be done by the same time as task 2
```

so this is not being set every time when `task add` is called, only when a DOM reference is detected :thinking:

djmitche:
@tbabej mentioned that this value should probably be a Context attribute.  That makes a lot of sense!

And, I think we can move the setting of contextTask out of the column methods and into the commands.  So CmdAdd wouldn't set it, but CmdModify (and anything else that modifies) would.

tbabej:
> And, I think we can move the setting of contextTask out of the column methods and into the commands. So CmdAdd wouldn't set it, but CmdModify (and anything else that modifies) would.

I think there's one potential problem with that, which is that the fact that column methods use context task to resolve DOM references that might refer to other tasks than the one being modified (as I tried to point out in the comment above). Which means a command like this:

```
task add test due:today
task 5 modify priority:3.priority project:4.project scheduled:due
```

will use the task ID 3 when evaluating the `priority` column as its contextTask, and task ID 4 when evaluating `project` column. and task ID 5 otherwise (for the `scheduled` column). This will set:
* priority of task 5 to the same value as task 3
* project of task 5 to the same value as task 4
* scheduled to the same value as due

djmitche:
DOM references are still possible if contextTask is NULL -- they just can't refer to "this" task.  So
```
task add scheduled:due
```
isn't supported, since that makes no sense[1].  But
```
task modify scheduled:due
```
is supported, as is
```
task add scheduled:4.due
```

For the first of those supported options (modify), we'll set contextTask.  For the second, contextTask is NULL but because the DOM reference identifies another task, it still works.

[...]

PR #2684, opened 2021-11-18
Refactor direct access to Task.data

### LINKED ISSUE #2681: Refactor all access to task properties to be made via Task methods
..and make `data` private.

The methods should be semantically meaningful, rather than specific to the key/value format of the stored data.

djmitche:
(Currently work-in-progress, following a fairly mechanical refactoring process with distinct, reviewable commits)

The idea here is #2681 - remove access to the `data` property, as part of a general process of focusing all of the code responsible for the data model into the Task class.  This will be especially important for mutating functions (as they will have to produce operations for #2507) but also for read functions, as it provides a single place where the details of the data model are implemented.

This will be a slow process, over many PRs.  Once I've got `data_removeme ()` deleted, I'll make this PR reviewable.


djmitche:
(this is still failing a few tests but I've run out of time at the moment to debug it)

djmitche:
OK, I think that fixed the last tests.  If CI is green, I'll make this reviewable.

djmitche:
CI is green.  This is best reviewed commit-by-commit.

djmitche:
Oh, I promised I'd get rid of the data_removeme() calls before asking for review.  May I place you on a brief hold?

tbabej:
> May I place you on a brief hold?

Standing by :slightly_smiling_face:

djmitche:
OK, should be good to go!

src/Filter.cpp
```diff
@@ -150,7 +149,7 @@ void Filter::subset (std::vector <Task>& output)
       for (auto& task : completed)
       {
         // Set up context for any DOM references.
-        contextTask = task;
+        contextTask = &task;
```
tbabej, line 152: I couldn't really come up with an example where this would break (and I guess this behaves same as previous version), but I wondered here if we should store and restore the previous value of `contextTask` pointer before and after evaluating the DOM reference.

tbabej: Ah, I see addressing this might be already part of #2683 ?

tbabej (approved):
This refactoring stood its ground against the test suite and my manual testing :slightly_smiling_face: 

The whole issue about the contextTask global variable is unfortunate, but I suppose getting rid of it completely would require a large refactor of the code base. I wonder, however, if it should not become part of the `Context` singleton at some point :thinking: 

Otherwise I agree this is a required step forward, and in particular I appreciate the good eye for refactorings along the way! They all look appropriate to me (and though the change for `getUDAOrphanUUIDs` seems to have been intentional, I agree with you the name makes less sense)

PR #2686, opened 2021-11-20
Task: Correctly handle bulk removal of virtual tags/depends attributes

### LINKED ISSUE #2655: mod depends: does not remove the dependency
task some_project mod depends: (but also with task 100 mod depends:)
To remove the dependency/dependencies
* What actually happened?
First I see a message 'no changes will be made' and then a list of 'modifying', but nothing actually is being modified.

Not sure if this is actually a bug or not implemented. I've tried to search here for a similar report, but couldn't find any.

---
With regards,

Angelo Machils

tbabej:
This edge case would happen if a user issued a following command

    $ task 1 mod depends:
or

    $ task 1 mod tags:

which would not perform as expected for tasks with non-empty depends /
tags attributes.

The problem under the hood is the fact that current synchronization
between 'tags" attribute and its constituent decomposed `tag_X`
attributes is performed in a way where the set of tags obtained from
`tag_X` attributes is taken as the source of truth.

If the legacy 'tags:' attribute was set to empty, the fixTags() method
would then restore its previous value from the `tag_X` attributes,
effectively leading to a no-op.

The same happens with dependencies. The fix here is to detect removal of
depends and tags attributes, and instead of setting the legacy
attributes to empty values, we iteratively remove all tags/dependencies.

Closes #2655.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,8 @@
 
 - TW #2648 xdg-open is not available on Mac OS-X
            Thanks to chapterjson for reporting.
+- TW #2655 The bulk removal of depends: and tags: is ignored
+           Thanks to angelus2014 for reporting.
 - TW #2664 Tag exclusion should be detected as invalid write context
            Thanks to bentwitthold for reporting.
 - TW #2671 The value of soww named date is incorrect
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -2276,6 +2276,19 @@ void Task::modify (modType type, bool text_required /* = false */)
             value == "''"   ||
             value == "\"\"")
         {
+          // Special case: Handle bulk removal of 'tags' and 'depends" virtual
+          // attributes
+          if (name == "depends")
+          {
+            for (auto dep: getDependencyUUIDs ())
+              removeDependency(dep);
+          }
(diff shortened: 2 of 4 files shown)


djmitche (approved):
This looks good, but I do kind of dislike putting the special case into the Task class, rather than the `modify` command.  This seems like a CLI shortcut, whereas a consumer of the Task class's API would just iterate over the deps and delete them.

tbabej:
@djmitche You're right on spot that the place where the fix is implemented is a bit unfortunate. Normally the _right_ place would be handling the value in the `ColDepends::modify` method, where also implementation of #2687 was placed.

The special case here is that the `::modify` for a given column is not called for empty values. The other suggestion for introducing this directly in the `modify` command itself would work for the explicit `modify` column, but other commands can also perform modifications, for example, you can start a task and remove dependencies in one command like this

```
$ task 1 start depends:
```

because the syntax of the `start` command is currently defined as:

```
       task <filter> start <mods>
              Marks the specified tasks as started. Is affected by the context.
```

I'll add a TODO note to investigate if this can be refactored in the future, perhaps by passing empty values to `::modify` column methods.

PR #2690, opened 2021-11-23
Fix typo in XDG news

sionleroux:
#### Description

This had XFG written in it, it's surely a typo of XDG since on the next line it's written correctly. Spotted this while reading the update notes.

I didn't, I only changed a 1-letter typo inside a string. It was the only occurrence I found in the whole repo so I don't think it's referenced by a test.


tbabej:
Spot on! Thanks for catching that :slightly_smiling_face:

PR #2700, opened 2021-12-18
Improve implementation of the "current task" for DOM references

### LINKED ISSUE #2683: Declare `contextTask` in one place

djmitche:
#### Description

This is a series of refactors of how DOM references to the "current task" work.

This can be reviewed commit-by-commit, but the combined diff is quite a bit shorter, as it ends up eliding a few intermediate refactoring steps.

#### Concerns

* I raised one concern in #2683: DOM references in `task add` are ambiguous and a bit buggy. I removed some tests for those, because the behavior the tests were asserting was ambiguous and buggy :)
* The second-to-last commit, adding a new CurrentTask class, might be a big overwrought.  If there's an easier way to do this kind of "revert this change on leaving the calling scope" thing, I'd love to know about it.  I'm no C++ pro.

### DIFF
--- a/src/Context.h
+++ b/src/Context.h
@@ -133,9 +133,9 @@ public:
   ~CurrentTask();
 
 private:
-  CurrentTask(Context *context, const Task *previous);
+  CurrentTask(Context &context, const Task *previous);
 
-  Context *context;
+  Context &context;
   const Task *previous;
 
   friend class Context;
--- a/src/Context.h
+++ b/src/Context.h
@@ -129,6 +129,7 @@ public:
 // CurrentTask resets Context::currentTask to NULL on destruction; this ensures
 // that this context value is not a dangling pointer.
 class CurrentTask {
+public:
   ~CurrentTask();
 
 private:
--- a/src/Context.h
+++ b/src/Context.h
@@ -38,6 +38,8 @@
 #include <Timer.h>
 #include <set>
 
+class CurrentTask;
+
 class Context
 {
 public:
@@ -72,6 +74,9 @@ public:
 
   void decomposeSortField (const std::string&, std::string&, bool&, bool&);
   void debugTiming (const std::string&, const Timer&);
+
+  CurrentTask withCurrentTask (const Task *);
+  friend class CurrentTask;
 
 private:
   void staticInitialization ();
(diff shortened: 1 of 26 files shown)


src/Context.h
```diff
@@ -115,6 +120,25 @@ class Context
...
   long                                time_render_us      {0};
   long                                time_hooks_us       {0};
+
+  // the current task for DOM references, or NULL if there is no task
+  const Task *                        currentTask         {NULL};
+};
+
+////////////////////////////////////////////////////////////////////////////////
+// CurrentTask resets Context::currentTask to NULL on destruction; this ensures
+// that this context value is not a dangling pointer.
```
tbabej: As far as I can tell, technically currently we the implementation does "stack" the contextTask references (similar to nesting Python's contextmanagers), so perhaps we should adjust the comment here to reflect that?

In other words:
```suggestion
// CurrentTask resets Context::currentTask to previous context task on destruction; this ensures
// that this context value is restored when exiting the scope where the context was applied.
```

src/Context.h
```diff
@@ -115,6 +120,25 @@ class Context
...
+class CurrentTask {
+public:
+  ~CurrentTask();
+
+private:
+  CurrentTask(Context *context, const Task *previous);
+
+  Context *context;
```
tbabej: The context should never be `null`, and since it's a singleton we don't need to reassign it either during the lifetime of this object, what do you think about using a reference here instead?

test/project.t
```diff
@@ -500,17 +500,9 @@ class TestBug1900(TestCase):
...
         code, out, err = self.t("_get 1.project")
-        self.assertEqual("due-b\n", out)
-
-        self.t("add foo project:scheduled-home")
-        code, out, err = self.t("_get 2.project")
-        self.assertEqual("scheduled-home\n", out)
-
-        self.t("add foo project:entry-work")
-        code, out, err = self.t("_get 3.project")
-        self.assertEqual("entry-work\n", out)
```
tbabej, line 513: While it's true that these tests test a superset of the condition "can a project name contain a dash", the special case of testing a DOM lookup of `<valid DOM ref>-<invalid DOM ref>` should be ultimately handled correctly (by failing the DOM ref lookup and the parser falling back to the raw string interpretation).

IMHO it's fine to regress on this for now, but we should move these tests into a separate test method marked with `@unittest.expectedFailure` decorator :slightly_smiling_face:

tbabej (changes requested):
Some minor improvements suggested, otherwise looks good. As far as I know this is the idiomatic way of handling context manager equivalents, so :+1: from me for the approach :slightly_smiling_face:

test/project.t
djmitche: I commented on this [here](https://github.com/GothenburgBitFactory/taskwarrior/issues/2683#issuecomment-997121680).  I'm not sure the test is representing expected behavior.  In particular, it tests that the *first* non-description attribute in `task add` does not parse DOM references, but other attributes are parsed.  I think we should consider the expected behavior in an issue, and then write tests for the result of that consideration.  I filed #2705 for that.

tbabej: I knew I saw this discussed somewhere before the holidays!

tbabej (approved):
LGTM, thanks for adding the changes. The test failure is in just one configuration of the matrix, and I believe it's a Github Actions fluke.

PR #2714, opened 2022-01-16
Add special case for taskserver mis-formatting of depends

### LINKED ISSUE #2689: Invalid json in export on some distributions
Running `task xxxx export` on two different installationes (debian and gentoo) with the same version (2.6.1) produced invalide json on gentoo (`…,"depends":["["yyy"]"],…`) while doing correct on debian (`…,"depends":["[\"yyy\"]"],…`).

I checked the patches applied on debian side but they does not show anything that would expect that differences.

djmitche:
#### Description

In #2689, it has been observed that a round-trip to taskserver can, under some circumstances, change
```
depends:"someuuid,someotheruuid"
```
to
```
depends:"&open;\\\"someuuid,someotheruuid\\\"&close;"
```

This PR adds a special-case to reverse this error and fixes #2689.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -657,6 +657,20 @@ void Task::parse (const std::string& input)
               ++annotation_count;
 
             data[name] = decode (json::decode (value));
+
+            // Fix for issue#2689: taskserver sometimes encodes the depends
+            // property as a JSON-encoded one-element array of strings.  So, if
+            // depends has form `[".."]`, then apply another layer of JSON
+            // decoding.
+            if (name == "depends" && data[name].length () >= 4) {
+              auto l = data[name].length ();
+              if (data[name][0] == '[' &&
+                  data[name][1] == '"' &&
+                  data[name][l-2] == '"' &&
+                  data[name][l-1] == ']') {
+                data[name] = data[name].substr(2, l-4);
+              }
+            }
           }
 
           attLine.skip (' ');
--- a/test/t.t.cpp
+++ b/test/t.t.cpp
@@ -119,6 +119,48 @@ int main (int, char**)
   value = ff4.get ("description");
   test.is (value, "Description", "ff4 description");
 
+  // depends as a comma-separated set of strings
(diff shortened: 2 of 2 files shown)


tbabej (approved):

bradyt:
@djmitche I'm not sure this fixes for even a single UUID, as I try to illustrate at https://github.com/GothenburgBitFactory/taskwarrior/issues/2716, but don't you mean the following?

```
depends:["someuuid","someotheruuid"]
```

And,

```
depends:"&open;\\\"someuuid\\\",\\\"someotheruuid\\\"&close;"
```

I don't actually see the format you indicate anywhere at https://github.com/GothenburgBitFactory/taskwarrior/issues/2689.

In other words, I only see the two basic formats `"x,y"` and `["x","y"]`, but your unit tests seem to have the shape of `["x,y"]`, which agrees with the latter only in the case of one UUID. And still, this doesn't seem fixed in that special case of depending on one task.

djmitche:
Hm, you're absolutely right.  I did miss that distinction!

Can you provide a reproduction of the single-dep failure with this patch in place?

bradyt:
@djmitche I think a reproduce is described at https://github.com/GothenburgBitFactory/taskwarrior/issues/2716, and I indicate the commit on develop. Do you want a repo with a Dockerfile to run? Or how else can I clarify?

djmitche:
Ah, I didn't see that, thanks.

bradyt:
Oh, big facepalm moment. I had your patch running in the Docker image, but maybe half the time I was testing with host computer's Taskwarrior. I'll try to add a little more automation to my test so it all runs in the Docker image. Apologies, I'll recheck both your previous patch, for single-dep, as well as new patch, multiple deps, etc, soon.

bradyt:
Okay test is more completely running in Docker image with patched Taskwarrior, still same issue with single UUID and first patch. I'll reply on the other patch regarding second patch.

PR #2728, opened 2022-01-28
TLSClient: fix multiple issues with error handling

jpalus:
- do not check errno on successful function calls (it might not be
  cleared after previous failed one)
- GNUTLS_E_* are not passed through errno but as function return value
- therefore there's more error spectrum than -1
- do not assume whole header is received, check number of bytes received

small additional improvements:
- read as many bytes into buffer as possible before appending to data
- skip writing nul byte at the end of buffer and use string_view instead
- added additional sanity checks

### DIFF
--- a/src/TLSClient.cpp
+++ b/src/TLSClient.cpp
@@ -30,6 +30,7 @@
 
 #include <TLSClient.h>
 #include <iostream>
+#include <sstream>
 #include <unistd.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


jpalus:
Note that these issues start to heavily manifest with latest gnutls 3.7.3 built with ktls support (see https://gitlab.com/gnutls/gnutls/-/issues/1314).

tbabej (approved):
Changes look good and passes the preliminary version of the integration tests :+1:

PR #2742, opened 2022-03-05
Add development docs

### LINKED ISSUE #2741: Move development-focused documentation into this repo
Currently, development documentation is on tw.org.  That's a bit problematic:
 * It's in HTML right now, making it difficult to edit and maintain (and it very much [is not maintained](https://github.com/GothenburgBitFactory/tw.org/issues/61)).  The [project to convert to markdown](https://github.com/GothenburgBitFactory/tw.org/pull/303) is rather bogged down and has quite a ways to go.
 * It's versioned separately from development, so it's difficult to find docs for older versions or correlate docs and code changes.
 * There's a lot of automated GH activity on the tw.org repo which makes it pretty hard to find changes.

I'd like to move the design documents, as well as the coding style, into this repo, either as markdown for viewing in GitHub, or as a simple "GitHub Pages" branch appearing at https://gothenburgbitfactory.github.io/taskwarrior (which would be labeled clearly with links to taskwarrior.org for user docs).

Followups would include
 * writing up the RFC process as it is
 * writing a new RFC for the Sync Overhaul project
 * documenting the current data model

Before I get started: @tbabej @lauft what do you think?

djmitche:
This moves some development-related documentation into the repo, where it can be more easily co-versioned with the source.

It currently uses Jekyll via GitHub pages.  You can see the result [here](https://djmitche.github.io/taskwarrior/).  The theming could certainly be better, but I'm mostly focused on content right now.  And, we could certainly move to use Hugo if y'all would prefer, but again: focused on content.

This just gets the docs in here.  There's lots to be done, and I'll file issues for them if this gets merged:

 * contributing docs are repetitive and outdated and could probably be slimmed down to one or two files
 * we need some kind of RFC process.  I had good luck with [this one](https://github.com/taskcluster/taskcluster-rfcs/blob/main/mechanics.md) so I'd propose something similar.  Nothing too heavy-weight, but some key milestones to know when a proposal can be acted on.  The existing RFCs need also sorting into "abandoned" and "implemented".  And I think "plans" is outdated enough to be removed.
  * documenting the task data model (based on [this](https://taskchampion.github.io/taskchampion/tasks.html))

Fixes #2741.


tbabej:
I think moving the content of the docs closer to the code is definitely the way to go - it will allows us to ensure correctness as the development progresses. I am not sure however, that we need to build and host the docs from this repo - perhaps we could simply check out the `docs/` subfolder of `taskwarrior` in the `tw.org` repo as a submodule?

tags
```diff
@@ -0,0 +1,37 @@
...
+s2.5.0
+s2.5.1
+s2.6.0
+sRecurrence
+stable
+v2.5.0.beta1
+v2.5.0.beta2
+v2.5.0.beta3
+v2.5.1.beta1
+
```
tbabej: Unclear what this file is for, but I suppose we should trash it?

djmitche: Oops, yes, that was a `git add .` mistake!

djmitche:
The submodule idea is an interesting one, but then blocks this project on the markdown rewrite -- which is precisely the blockage I'm trying to avoid.  It also depends on the deployment frequency of tw.org, which would have to be frequent or somehow triggered on a commit to this repository.

tags
djmitche: (now removed)

docs/contrib/build.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+# Cloning the Repo
+
+    $ git clone https://github.com/GothenburgBitFactory/taskwarrior.git taskwarrior.git
+
+
+# Building the Stable Version
+
+The master always represents the more recently released version, and should be
```
lauft, line 33: `master` has become `stable`

docs/contrib/build.md
```diff
@@ -0,0 +1,104 @@
...
+
+    $ cd taskwarrior.git
+    $ git checkout master                # Master is the stable branch.
+    $ cmake -DCMAKE_BUILD_TYPE=release . # 'release' for performance.
+    $ make                               # Just build it.
+
+
+# Building the Dev Branch
+
+The dev branch is always the highest numbered branch, in this example, `2.6.0`.
```
lauft, line 44: There are no more version branches, only `develop` for development

docs/contrib/build.md
```diff
@@ -0,0 +1,104 @@
...
+[vramsteg](https://gothenburgbitfactory.org/projects/vramsteg) utility
+provides blinkenlights for test progress.
+
+    $ cd taskwarrior.git/test
+    $ make VERBOSE=1     # Shows details
+    $ ./run_all          # Runs all tests silently > all.log
+    $ ./problems         # Find errors in all.log
+
+
+# Submitting a Patch
```
lauft, line 67: Should be changed to pull-requests

docs/contrib/coding_style.md
```diff
@@ -0,0 +1,31 @@
...
+To be a little more explicit, the common elements across the languages are:
+
+-   Indent code using two spaces, no tabs
+-   With Python, follow [PEP8](https://www.python.org/dev/peps/pep-0008/) as
+    much as possible
+-   Surround operators and expression terms with a space
+-   No cuddled braces
+-   Class names are capitalized, variable names are not
+
+We target Python 2.7 so that our test suite runs on the broadest set of
```
lauft: Guess that is Python 3 now...
Maybe state even at least Python 3.9 (according to https://python-release-cycle.glitch.me/)

docs/rfcs/_index.md
```diff
@@ -0,0 +1,24 @@
+---
+title: "Taskwarrior - What's next?"
+---
+
+# Design
```
lauft: Add documents present in this directory, remove empty sections

docs/contrib/branching.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Branching Model
+---
+
+Software development typically requires a standardized branching model, to
```
lauft: This document is somehow too general (mentioning 'hidden' topic branches on the developer's machine), also there is some overlap with other documents (`build.md`, `development.md`)

docs/contrib/first_time.md
```diff
@@ -0,0 +1,244 @@
+---
+title: How to become an Open Source Contributor
+---
+
+Welcome, potential new Open Source contributor! This is a guide to show you
```
lauft: In general this document needs to be adapted to the new branching model (only `stable`/`develop` branches)

docs/rfcs/task.md
```diff
@@ -0,0 +1,761 @@
...
+
+Note that not all combinations of and make sense, for example \"3annual\" makes
+no sense, but evaluates to \"3years\".
+
+
+## The Attributes
+
+Here are the standard attributes that may comprise a task:
+
+<table>
```
lauft: Reformat to Markdown table?

lauft (commented):
General comments:
- Keep design/developer documents as simple Markdown files
- Format "one sentence, one line" (leave linebreaks to the Markdown renderer)
- Rename `_index.md` to `README.md` or `<folder-name>.md` or remove
- Move images to common directory?

docs/rfcs/_index.md
djmitche: Oops, I should have deleted this (fixed)

djmitche:
Thanks for the review!

>    Keep design/developer documents as simple Markdown files

I'm not sure what you mean here.  All of the docs files added in this PR are Markdown.

>    Format "one sentence, one line" (leave linebreaks to the Markdown renderer)

Sounds good -- I'll make that change and push it shortly.

>   Rename _index.md to README.md or <folder-name>.md or remove

This doesn't work with Jekyll.  Would you prefer to not use GitHub pages?

>    Move images to common directory?

If you hadn't included `?` I'd just do it on your say-so.  But, these images "belong with" the docs that link to them, so it seems reasonable to put them in the same directory.  I think a big reason to put images in "images" is that we used to have to configure Apache with different settings for those files..

_To the more editorial comments above:_

As I mentioned in the PR description, the plan here was to just get the docs moved -- it needs _lots_ of work after that!  I'd like to avoid doing so in this PR for a few reasons:
 * editing the documentation is necessarily a subjective exercise and I would expect disagreement with my opinions, leading to a lot of discussion;
 * getting these docs landed here is important to me in unblocking further documentation work that is orthogonal to editing what's here; and
 * ideally we can parallelize the followups, across multiple people or at least multiple PRs (so the uncontroversial stuff can land first).

I mentioned a few follow-ups in the PR description, and you've identified a few points included in the first ("contributing docs are repetitive and outdated") which I'd like to defer to that point.

docs/rfcs/task.md
djmitche: I did a few of these, where the Markdown table could reasonably replicate the HTML.

But on some of the never-finished RFCs with complicated tables it just wasn't worth the effort, so I left the tables as-is.

djmitche: (in the end this was just the "plans" RFC, which is from 2016 and maybe should just be rm -rf'd?)

docs/contrib/contribute.md
```diff
@@ -0,0 +1,64 @@
...
+-   Suggest enhancements.
+    We get lots of these, and it's great.
+    Some really good ideas have been suggested and implemented.
+    Sure, some are out of scope, or plain crazy, but the stream of suggestions is fascinating to think about.
+
+-   Participate in the [bug tracking](https://github.com/GothenburgBitFactory/taskwarrior/issues) database, to help others and maybe learn something yourself.
+
+-   Help [triage](/docs/triage) the issues list.
+
+-   Join the IRC channel \#taskwarrior on freenode.net and help answer some questions.
```
lauft, line 27: freenode is dead, long live libera.chat (👉🏻 https://web.libera.chat/#taskwarrior)

docs/contrib/contribute.md
```diff
@@ -0,0 +1,64 @@
...
+
+-   Join either the [developer-mailinglist](https://groups.google.com/forum/#!forum/taskwarrior-dev) or [user-mailinglist](https://groups.google.com/forum/#!forum/taskwarrior-user) (or both) and participate.
```
lauft, line 29: @tbabej I guess we should close down these and direct to our Discord (👉🏻 https://discord.gg/eRXEHk8w62) here...
Also, the [main README](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/README.md) mentiones Reddit, Twitter, and GitHub Discussions. Don't know, whether those are worth mentioning here.

docs/rfcs/task.md
lauft: Deleting `plans.md` if it is outdated is fine with me. But as I am not directly involved in Taskwarrior development, I would delegate the final decision to @tbabej  😃

lauft:
Hi @djmitche 

>>    Keep design/developer documents as simple Markdown files
> I'm not sure what you mean here. All of the docs files added in this PR are Markdown.
>>    Rename _index.md to README.md or .md or remove
> This doesn't work with Jekyll. Would you prefer to not use GitHub pages?

I would prefer to keep the developer documentation as simple and un-fancy as possible, i.e. just Markdown documents _only_, no web hosting. This would keep the Taskwarrior repository free of any 'side use' ( = hosting a web site). The Markdown documents can be linked, and IMHO browsing them on GitHub (like any other code file) is not that bad/cumbersome that it would require more CSS/etc. to increase user experience. However, this is my personal, "what I would do if Taskwarrior was my main project" opinion, so feel free to override it if you see benefits from any other decision... 😉 

>>    Move images to common directory?
> If you hadn't included `?` I'd just do it on your say-so. But, these images "belong with" the docs that link to them, so it seems reasonable to put them in the same directory. I think a big reason to put images in "images" is that we used to have to configure Apache with different settings for those files..

Fair point. There are not may images, so I am fine with this decision.

> As I mentioned in the PR description, the plan here was to just get the docs moved -- it needs lots of work after that! I'd like to avoid doing so in this PR for a few reasons...
I mentioned a few follow-ups in the PR description, and you've identified a few points included in the first ("contributing docs are repetitive and outdated") which I'd like to defer to that point.

I definitely agree with that. Progressing in small steps is better than working on that big one forever.

lauft (approved):

djmitche:
I'll leave the call on GitHub Pages to @tbabej then.  I agree it's not critical, and linking to GitHub /blob/ renderings is adequate.  If we choose to not use GitHub pages then I'll rename to README.md and verify that all of the links work.

docs/contrib/contribute.md
smemsh: I'm usually in the irc channel on libera.chat and I know there are several others I've seen / talked to there and which lurk or send a message, sometimes with days latency but responses occur mostly.  I don't know much about the discord population, but we could keep mention of both irc and discord

tbabej:
IMHO this is a great step forward to what we have right now, which is an absence of the development docs in this repo, hopefully enabling us to keep the docs more in sync with the changes. In the future I'd like to make the website repo use this as an authoritative source, but that's future work.

Thanks @lauft for all the suggestions :slightly_smiling_face:

djmitche:
@tbabej did you want to set up github pages for this?  Or, should I make a PR to rename indexes to README.md?  Or just leave it as-is?

tbabej:
Yep, I'll setup github pages in the interim, right now I am trying to figure out the security audit permissions problem :thinking:

djmitche:
Awesome.  https://stackoverflow.com/questions/70435286/resource-not-accessible-by-integration-on-github-post-repos-owner-repo-ac looks like it may be relevant there.  Don't spend too long on it -- feel free to assign an issue to me instead if you'd prefer.  I had hoped that merging some dependabot PRs would magically "fix" it but that's not realistic :)

Issue #2770, opened 2022-04-23
Replace Task & DB interface with C interface to TaskChampion


djmitche:
TaskChampion now has a [C interface](https://github.com/taskchampion/taskchampion/pull/332) via Rust FFI.

The next step is to integrate this into Taskwarrior:
 * Link with TaskChampion
 * Wrap the C API into C++ (smart pointers, method wrappers)
 * Replace TF2 and TDB2 with this implementation

The result will be a Taskwarrior that uses a solid data backend with a secure sync protocol, and is interoperable with other tools also built on TaskChampion.

djmitche:
I've gotten a start on this in https://github.com/GothenburgBitFactory/taskwarrior/pull/2771

djmitche:
OK, I'm going to break this down a bit and land things one by one.

One major change is this: I'll take the code from TaskChampion and make it a part of TW.  Then we don't have two "TaskThing"s causing confusion.

PR #2771, opened 2022-04-23
Replace Task & DB interface with C interface to TaskChampion

djmitche:
See #2770.


djmitche:
I'm still hacking away at this, if very slowly.  I'm working on getting the test suite passing now, and that's turning up a lot of edge cases that work differently with the new data model.  And the C++ builds are soooo slow so I pretty much make a change and then go do something else while it builds :)

smemsh:
May I ask, after looking a bit at the docs and code, am I understanding right that this is really only replacing the storage and sync as backend for serial records, not using the query engine for any kind of "typed" access? So we will not be able to have any kind of indexed access to the task fields or optimize TW queries or anything (by utilizing the SQL engine), and the client will still have to "parse" tags/dependencies lists for all tasks and search iteratively through all the records to find things? Just curious, thanks.

djmitche:
Yes, and in fact that's how Taskwarrior has always worked.  There's already more indexing available, in that tasks can be accessed directly by uuid.  But operations such as `children(task)` continue to be implemented as linear scans of all tasks (or more precisely for `children(task)`, all pending tasks).

smemsh:
Right, it's always worked that way, but it misses some optimizations and interesting things we could do with the data if we could use the db query engine as well, with engine-native task records.  When I heard about TC and its sqlite backend, I was hopeful we were going to get some "db-like semantics"  or even just be able to use our own SQL on the task data.  But in fact, it's not really storing them as typed data, but just essentially as serialized records (like it does now).

In any case, it's still worthwhile for the transactional semantics, sync server improvements and as you mentioned elsewhere, other language bindings.  Rust is probably a good thing too, but I'm not sure yet ;-)

djmitche:
I think the idea is that a taskdb should be fairly small (100's of tasks), so optimization isn't too critical.  But, there's certainly room to change the backend later, too.  We actually have two kinds of flexibility: storage is already well-abstracted so, for example, TC could work with no changes against other DB engines, if someone took the time to write a storage backend for those engines.  Imagine, for example, a service like in.the.am that wants to store user tasks in a shared MySQL DB rather than per-use sqlite DBs.

The second kind of flexibility is that we can change the storage API itself.  Especially, adding optimized functionality that can be implemented in terms of simpler operations is easy.  For example, we could add a "query" method that takes a filter expression and returns matching results.  This can have a default implementation that just performs a scan of all tasks and returns those that match, with more performance-oriented backends translating the filter expression into a DB query.

This effort is dragging a little bit since I'm the only one working on it.  Would you like to help out?  There's some work in Rust to do if you'd like (and I'm happy to mentor that) and some bits in C++ that I could carve off if you're already familiar with the TW codebase (I'm still learning about the codebase).

smemsh:
Yeah, task db sizes are small, perhaps no need for optimization, but that's just for one person's tasks.  A hosted tw provider might need it to be more clever.  Myself, I'm adding several hundreds of tasks per year, I know some people use it even more finer-grained (many tasks per day), I don't know when this will start showing up as little bits of latency.  I also find that I sometimes have to call taskwarrior multiple times in my wrappers when one result is used for the next.  Maybe I can make more clever filters or work off one single export but having native task columns in SQL sure would be nice.  I suppose I can make some kind of view that unserializes the data into a virtual table or something.

In any case, changing only one layer at a time and letting the interfaces bake a bit is a good thing and I'm not aware of any large scale users at the present time (maybe there are though?)

I only know C (not much C++) and Python, but I do need to learn Rust.  I will study the TC code while I'm doing that, with an eye on possibly helping later if I'm able, but couldn't contribute much presently.  I do see you're pretty much the only one doing the Rust though and am grateful for your efforts and excited about the future possibilities.

djmitche:
This work is in progress, this PR is no longer a useful measure of that progress.

PR #2781, opened 2022-05-07
vim: Add missing config options to taskrc syntax file

drehak:
Sourced from `taskrc(5)` and default config (`Context.cpp`), this should cover all options present in there.


tbabej:
A good catch! Perhaps we should add a note into Context.cpp that the options should be covered here... :thinking:

tbabej:
Added in 0a558ef7c8f3321e7451294512c885661f5b4a84

PR #2799, opened 2022-05-22
Make TDB2.backlog non-public

djmitche:
The stats command gets this information from an API that will also work
for TaskChampion.  The sync command still accesses the field directly,
as the command must be completely rewritten for TaskChampion.

This is one small part of a general effort to make the interface to TDB2 more abstract, so that it can eventually be implemented with a Replica. (#2801)


tbabej (approved):
Worked in my testing last night.

PR #2821, opened 2022-05-27
Update all rust dependencies.

djmitche:
#### Description

Hopefully this will address all of the (comatible) upgrades that dependabot wants to do in one go.


.github/workflows/rust-tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         rust:
           # MSRV; most not be higher than the clippy rust version in checks.yml
```
tbabej, line 15: Unrelated, but `s/most/must`, I'll push a fix.

tbabej (approved):

Issue #1693, opened 2018-02-14
[TW-1669] annotations: escape the escape character


taskwarrior:
_Daniel Shahaf on 2015-08-25T07:51:23Z says:_ 

 In 2.4.5 (2331810), it is not possible to add the literal string `\n` (two bytes) to an annotation using `task annotate`:

```
% task 1 annotate 'Lorem \n ipsum \\n dolor \\\n sit
 amet'
% task _get 1.annotations.1.description
Lorem 
 ipsum 
 dolor \
 sit
 amet
```

This is a regression from 2.4.4, where the output was:
```
Lorem 
 ipsum \n dolor \
 sit
 amet
```

The latter is what I would expect.

taskwarrior:
Migrated metadata:
```
Created: 2015-08-25T07:51:23Z
Modified: 2017-01-16T17:59:07Z
```

DanielMowitz:
I'd like to look into this one, seems interesting.

DanielMowitz:
What I have found out so far:

- The problem seems to be that the input is processed too often. i.e. "\\\n" becomes "\n" but then is processed again to a newline.
- This happens in Lexer::readWord
    - Specifically in the block starting with `// an escaped thing`
    - Lexer::readWord has four definitions, two of which reside in taskwarriors Lexer.cpp, the other two in the shared libraries'.
    - The behavior is correct if the escaped-thing-block is commented out in *all* implementations
   
This raises some questions:
- Why are there four implementations of the same function, none of which call any other but contain a lot of similar code?
- Why is the input to the annotate command processed so often?

edit: Markdown gobbled one of the backslashes in the first bullet point. Escape characters are tricky ;)

DanielMowitz:
After running a small test program through gdb a few times, I found out that the shell already parses "\\\\n" to "\\n" when reading in argv. This can be easily checked by running `echo "\\n"` and `echo "\n"` and observing that they print the same characters. Therefore, both "\\" and "\\\\" are indistinguishable to the c++ code. This explains that "\\\\n" is interpreted as a newline, as the escaped-thing-codeblock only ever sees one backslash.

Sadly I don't see a way for taskwarrior to fix this issue, as we can't controll the users shell or it's escape character handling. 

This leaves people who want to annotate their tasks with "\n" with the option of using `task 42 annotate "a \\\nice annotation"`

smemsh:
[...]

PR #2847, opened 2022-06-12
Fix escape character escaping as described in #1693

DanielMowitz:
#### Description

The fix is achieved by splitting the `Lexer::readWord` function into multiple parts and calling only the necessary one in `CLI2.cpp`.

I was not completely sure about the naming of the new functions and would like to discuss that in this PR.
Since I am also still in the process of writing a new test for #1693, there is no test suite run attached.

### DIFF
--- a/src/CLI2.cpp
+++ b/src/CLI2.cpp
@@ -443,7 +443,7 @@ void CLI2::lexArguments ()
 
       cursor = 0;
       std::string word;
-      if (Lexer::readWord (quote + escaped + quote, quote, cursor, word))
+      if (Lexer::extractWord (quote + escaped + quote, quote, cursor, word))
       {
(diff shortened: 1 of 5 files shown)


DanielMowitz:
A test for the original issue has been created.

All other tests now pass as well. Sadly, part of the code introduced to accomplish this is rather ad-hoc. 
This is because `task add` seemingly relies on `readWord` to escape unicode characters, but not backslash escapes. These seem to get handled elsewhere, with the exception of `\'`and `\"`, which needed to be done manually.
I'm not fairly happy with this, but it works.

djmitche:
Since this is old and not particularly confident, I'm going to close this PR. I've had a tough time working on similar bugs, as the handling of various escaping and un-escaping is kind of all over the place.

PR #2869, opened 2022-07-04
Add a C++ wrapper around TC FFI

djmitche:
This uses CMake to build a simple Rust library (in `src/tc/rust`) that just re-exports everything from the `taskchampion-lib` crate. The C++ wrappers then wrap this into C++ objects with proper lifecycle maintenance, in the `tc` namespace.  More details are in the included docs.

The C++ wrappers are incomplete, and missing methods are tagged with "TODO".  These will be added as needed.

Refs #2770.


djmitche:
(I may need some help getting this to build everywhere... I suspect that none of these builders have Rust!)

djmitche:
I can definitely figure out how to fix the rust tests, and I'll take a swing at the regular tests too.

djmitche:
Hm, I'll poke at that windows error.  It's always windows.

djmitche:
OK, that's better!  I have also renamed `rust/` to `taskchampion/` since that's a better representation of what it contains.

djmitche:
@tbabej I'll rebase this, but -- any chance you could have a look soon?  It is sort of blocking next-steps :)

tbabej:
Looks like I made the rebasing of this one worse today, I'll take a swing at it.

docs/contrib/rust-and-c++.md
```diff
@@ -0,0 +1,29 @@
+# Rust and C++
+
+Taskwarrior has historically been a C++ project, but as part of an [ongoing effort to replace its storage backend](https://github.com/GothenburgBitFactory/taskwarrior/issues/2770) it now includes a Rust library called TaskChampion.
```
tbabej, line 3: Not a blocker for this PR, but IMHO this name dichotomy will be confusing long-term. I'll try to reach out to the owner of TW crate again, or perhaps we should come up with a more descriptive name if that's not possible.

.gitmodules
```diff
@@ -1,3 +1,9 @@
 [submodule "src/libshared"]
 	path = src/libshared
 	url = https://github.com/GothenburgBitFactory/libshared.git
+[submodule "cmake/CMakeRust"]
+	path = cmake/CMakeRust
+	url = https://github.com/Devolutions/CMakeRust
+[submodule "src/taskchampion"]
+	path = src/taskchampion
+	url = https://github.com/taskchampion/taskchampion
```
tbabej: If I'm not mistaken, I don't think we need this submodule anymore?

tbabej: My bad, this is later fixed in 81003bd610f238616b1ed90bd5f711eb6c7b3442.

docs/contrib/rust-and-c++.md
djmitche: Yeah, that will be no harder later than it is now.

djmitche:
Looks all green, but your review is just a comment -- will you be looking again?

tbabej (approved):
Changes look good on my end. I was also able to interact the FFI using Python :+1:

djmitche:
Awesome, I'm excited to hear it worked from Python too :)

Issue #2830, opened 2022-05-30
Address RUSTSEC-2021-0124


djmitche:
This is a vulnerability in tokio, which is required by Actix-web.  For the moment, let's ignore it, and then decide whether to upgrade actix to suit, or switch to a different (simpler) web server package.

djmitche:
It looks like the [audit-check action does not support `--ignore`](https://github.com/actions-rs/audit-check/issues/223), so for the moment it might be best to just disable this check.  We can re-enable it either when this advisory is addressed (by actix-web) or after switching to another web framework.

djmitche:
Hm, audit-check appears unmaintained.  I've pinged the maintainer, but if no response then we might need to fork that action.

djmitche:
@tbabej ok if I fork that to the `GothenburgBitFactory` org or should I do it to `djmitche`?

tbabej:
Forking under the org is fine with me!

pinkforest:
Hi lovelies, feel free to PR link to the fork here: https://github.com/rust-secure-code/projects :purple_heart: 

Or.. maybe we could adopt it under rust-secure-code / rustsec ? - 

I've asked others whether this is something we could do: https://github.com/rust-secure-code/wg/issues/46

djmitche:
I just noted in https://github.com/GothenburgBitFactory/taskwarrior/pull/2903 that the fork doesn't actually add any value over upstream.  But this does seem to be a fairly "important" rust-security-related action, so I'd vote to include it in a collective org like rust-secure-code.

pinkforest:
Cool. There is also cargo-deny alternatively e.g.:
https://github.com/libp2p/rust-libp2p/pull/2803

But I'll see the feasibility of whether we can adopt this in the meantime

djmitche:
This has been resolved in other repos.

PR #2872, opened 2022-07-11
Use actions-rs/audit-check with its new ignore option

djmitche:
#### Description

..to ignore RUSTSEC-2021-0124 as described in #2830.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -15,6 +15,7 @@ jobs:
     name: "Audit Dependencies"
     steps:
       - uses: actions/checkout@v2
-      - uses: actions-rs/audit-check@v1
+      - uses: GothenburgBitFactory/audit-check@master
         with:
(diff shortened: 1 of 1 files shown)


djmitche:
Audits don't typically run on pull requests, so I added a temporary commit here to change that -- and the audit ran successfully.  I then removed that temporary commit.

.github/workflows/security.yml
```diff
@@ -0,0 +1,21 @@
...
+  push:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    name: "Audit Dependencies"
```
tbabej: Perhaps during the transition period we should just be explicit here?

```suggestion
    name: "Audit Rust Dependencies"
```

PR #2903, opened 2022-08-06
use the (existing) .cargo/audit.toml to ignore some rust advisories

djmitche:
I was grepping for `RUSTSEC` and realized we had ignores set up already, but they had ended up in `taskchampion/.cargo` instead of `.cargo`.  Let's see if the GH actions pass for this PR..


djmitche:
It does!  So we don't actually need the forked action.

tbabej:
Nice! I suppose that means we can archive the audit-rs repo?

tbabej (approved):

djmitche:
There's some question about people still using it, or moving it somewhere more official, so let's hold off on that.

pinkforest:
psst we've forked - https://github.com/rustsec/audit-check

psssst it's sometimes helpful to have runtime flag support instead of having to rely on some config file in-repo

djmitche:
True!  Would you like me to propose the same PR to the new "upstream"?

I'll archive the repo in this org.

PR #2904, opened 2022-08-07
Add Replica::num_undo_points and exclude undo points from num_operations

djmitche:
#### Description

Taskwarrior's stats include a number of available undo's.  In the TC data model, these are UndoPoint operations, so make a function to count those.

Also, those aren't real "operations" in the sense that they won't be synchronized to the server, so don't count them in num_operations.


tbabej (approved):
LGTM! My Rust is getting better, I was able to understand this one (or this PR is particularly easier to read :sweat_smile: )

djmitche:
Everything improves with practice!

PR #2907, opened 2022-08-08
Fix repository name in audit check

djmitche:
#### Description

I forgot this doesn't run on PRs, so my confidence that it's using .cargo/audit.toml may have been misplaced.


djmitche:
OK, this ran successfully on a PR (and I subsequently removed the commit to configure it to do so)

PR #2968, opened 2022-10-13
Ignore patch updates for some noisy deps

djmitche:
#### Description

This should help reduce the number of dependabot PRs we get for what are, ultimately, pretty inconsequential dependencies.  We're not using much functionality from either clap or libc, so patch updates are unlikely to be important.


smemsh:
wish there was a way to ignore all automatic rust library update notifications, while remaining subscribed to other commit notifications

djmitche:
We could certainly make other changes to encourage less frequent updates -- a weekly schedule, for example.

smemsh:
Really github should just have a more complete filter mechanism... looks like it cannot filter on title text and has no exclusion mechanism :thinking:

Personally, I avoid automatic updates of dependencies, preferring as-needed strategy and then top-off close to release.  But keeping patch-level updates automatic has advantages also.   Here's the last two whole months, FYI:
```
7 anyhow
6 thiserror
6 libc
5 serde
4 clap
3 chrono
2 futures
2 serde_json
1 env_logger
1 pretty_assertions
1 iana-time-zone
1 strum_macros
```
So 39 or 1.3 every 2 days.  It's not too much I suppose, but it's subjective :-)

djmitche:
A big reason to do automatic updates is so that, a few years later when there's a CVE or some critical bug, you're not left trying to upgrade through a half-dozen major revisions with breaking changes.

I could just change them _all_ to skip patch-level updates.  That shouldn't let things get _too_ out of date.

djmitche:
I'm going to go ahead and merge this to cut down the noise a bit.

PR #2974, opened 2022-10-13
Add support for _not_ creating a DB if one does not exist

djmitche:
#### Description

I noticed some tests that verified Taskwarrior does not create a new DB if one does not exist, so I've added that functionality to TaskChampion.

### DIFF
--- a/taskchampion/taskchampion/src/storage/sqlite.rs
+++ b/taskchampion/taskchampion/src/storage/sqlite.rs
@@ -87,11 +87,11 @@ impl SqliteStorage {
 
         // Open (or create) database
         let db_file = directory.as_ref().join("taskchampion.sqlite3");
-        let mut flags = OpenFlags::SQLITE_OPEN_READ_WRITE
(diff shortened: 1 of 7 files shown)


taskchampion/taskchampion/src/storage/sqlite.rs
```diff
@@ -76,13 +76,24 @@ pub struct SqliteStorage {
...
+            // Ensure parent folder exists
+            std::fs::create_dir_all(&directory)?;
+        }
 
         // Open (or create) database
         let db_file = directory.as_ref().join("taskchampion.sqlite3");
-        let con = Connection::open(db_file)?;
+        let mut flags = OpenFlags::SQLITE_OPEN_READ_WRITE
+            | OpenFlags::SQLITE_OPEN_NO_MUTEX
+            | OpenFlags::SQLITE_OPEN_URI;
```
tbabej: Discussed in person: We could use a `OpenFlags::default()` to make sure we don't go out of sync with the default set of flags for `Connection::open_with_flags`

tbabej (approved):

Issue #174, opened 2018-02-12
[TW-117] Ability to assign urgency coefficients to filter (a.k.a. combinations of attribute values) as well as individual attributes


taskwarrior:
_Max Muller on 2012-11-15T02:25:21Z says:_ 

 I'm really starting to love the urgency attribute, described http://tasktools.org/gitweb/gitweb.cgi?p=rfc.git;a=blob_plain;f=rfc31-urgency.txt;hb=HEAD and elsewhere. It's awesome, As tw user/visionary David Patrick has noted, its well beyond the abilities of most all known task management programs. And it automates task management. Leave the decisions up the computer, where they belong...

However, I've encountered two types of limitations which, if they were addressed, would enhance the value of the urgency attribute much more. The first is described http://taskwarrior.org/issues/1076, and which I'd describe in the terminology of taskwarrior as follows (I hope I get this right): the ability to customize the mapping of attribute values to the numeric scale of the attribute's urgency. This would allow users to not just assign urgency coefficient to user-defined attributes, as users can do now,,, but to have urgency value differ depending on the _value_ of the user-defined attribute.

The above is a restating of a prior feature request. (again, it's this one: http://taskwarrior.org/issues/1076).

Here's the new one: 

It would be even MORE flexible to be able to assign urgency not just to single attributes, but to the results of queries or filters of multiple attributes. This would make assigning urgency to a task as flexible as searching for it using taskwarrior. FOR EXAMPLE: I would like to have unscheduled tasks that are due today or tomorrow have a HIGH urgency (so I can be sure to schedule them--using tasks sched: atttribute--to a certain day or time). I would like to have have UN-scheduled tasks due today or tomorrow be much less urgent (because there's plan to do them at some time already and they don't need my attention right now). 

Here how this could be implemented in a .taskrc file: 

_first define a filter's urgency:__
urgency.user.filter.due-soon-and-unscheduled='(due.today or due:tomorrow) scheduled.none:' 

_then assign an urgency coefficient to the filter:_
urgency.user.filter.due-soon-and-unscheduled=12.0

Having this ability would address the concern expressed here: http://taskwarrior.org/issues/1062 (although less elegantly than proposed in that feature request).It would also allow me to do something I currently can't, which is have tasks due today be very urgent and tasks due tomorrow much less urgent.

taskwarrior:
Migrated metadata:
```
Created: 2012-11-15T02:25:21Z
[...]

PR #2992, opened 2022-11-04
Draft: Custom virtual tag

pianocomposer321:
This PR adds a new feature, custom virtual tags. They can be specified in the config like so:

```
virtualtag.FORTODAY.filter=planned.before:tomorrow
```

Then they can be used like any other virtual tag, and the filter will be applied, e.g. `task +FORTODAY` will do the exact same thing as `task planned.before:tomorrow`. This is especially useful because you can use these tags in urgency settings as well:

```
urgency.user.tag.FORTODAY.coefficient=2.0
```

In this way you can effectively give all tasks that match a filter a boost in urgency.

I believe this would satisfy https://github.com/GothenburgBitFactory/taskwarrior/issues/174, and possibly others.

I have not written documentation yet, I didn't want to put in that work before I knew whether this feature would even be accepted. This is also my first contribution, so I may need a little help in learning the lay of the land ;-).

### DIFF
--- a/src/Eval.cpp
+++ b/src/Eval.cpp
@@ -203,6 +203,12 @@ void Eval::debug (bool value)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+void Eval::ignoreTag(std::string tag)
+{
+  ignoreTags.push_back(tag);
+}
+
+////////////////////////////////////////////////////////////////////////////////
 // Static.
 std::vector <std::string> Eval::getOperators ()
 {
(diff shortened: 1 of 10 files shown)


djmitche:
Were the changes to `tc/rust/Cargo.log` intentional?

djmitche:
(I don't have an opinion on the core of this PR -- I was flagged for review due to the tc/rust changes :) )

pianocomposer321:
> Were the changes to `tc/rust/Cargo.log` intentional?

No...I'm not sure how those changes came about. Is that file edited automatically with make or anything like that?

In any case, I can remove the changes...should I just add a new commit, or somehow edit a previous one. Sorry I'm not a git expert, so I'm not sure what the right thing is to do to revert a mistake like that lol...

djmitche:
You can just add a new commit to undo those changes.

pianocomposer321:
Ok, so I noticed that there was a problem with my implementation, and I'm running into some issues trying to solve it.

Basically the way I'm testing whether a given task passes a filter is by adding the filter to the end of the arguments and then re-parsing the arguments. This was the only way I could see it being done in the rest of the codebase, and the `Filter` object seems to be designed around this method. The problem is that the filter is not removed from the arg list afterwards (and I didn't see an easy way to do this), which means that the argument list is being appended to permanently every time a task is checked for a custom virtual tag, which leads to unexpected results, as it affects the filter for everything.

So my question is this: is there a better way to check if a task is included in an arbitrary filter (the filter being a string of text that would be passed as an argument, like `scheduled.before:today+4d` or similar)?

djmitche (commented):
(deferring review to @tbabej)

Issue #1534, opened 2018-02-14
[TW-1508] The until, wait and scheduled attributes are not shifted by the recurrence span


taskwarrior:
_Tomas Babej on 2015-01-13T10:16:42Z says:_ 

 If you create task like:

```
task add test due:2014-01-12T22:00:00 until:2014-01-12T23:00:00 recur:daily
```

(today is 2014-01-13)

You will see that the child tasks created for this recurrent task do not have shifted until attribute by the reccurence span:

```
Name          Value                                     
ID            3
Description   test                                      
Status        Pending
Recurrence    daily                                     
Parent task   f6470fad-4879-4281-a6b9-75664222d80c
Mask Index    1.000000                                  
Entered       2015-01-13 10:59:13 (32 seconds)
Due           2014-01-13 22:00:00                       
Until         2014-01-12 23:00:00
Last modified 2015-01-13 10:59:10 (35 seconds)          
UUID          d4e99ba5-34d9-4d65-a551-5d589be8086b
Urgency       12                                        
                               due     1  *   12 =    12

```

The same behaviour is present for wait and scheduled attributes.

taskwarrior:
Migrated metadata:
```
Created: 2015-01-13T10:16:42Z
Modified: 2016-05-04T12:54:32Z
```

taskwarrior:
_Tomas Babej on 2015-01-19T18:06:07Z says:_ 

 I wrote a hook to workaround this issue. See:

https://github.com/tbabej/taskwarrior-shift-all-recurrence-hook

taskwarrior:
_slaven ʙanovic on 2016-05-04T12:54:32Z says:_ 

 This bug has a bigger impact than it seems at first. It actually deletes the template and thus all childs that should be created after the first roundabout.

Verify with the following script:
{code}
d0=`date "+%FT%T"`
d1=`date "+%FT%T" --date "1 min"`
d2=`date "+%FT%T" --date "2 min"`
echo $d0 $d1 $d2
task add +twdebug new task with desc 1 due:$d1 wait:$d0 until:$d2 recur:5min
sleep 3
task all +twdebug
sleep 70
task all +twdebug
sleep 60
task all +twdebug
{code}

{code:title=one line paste-able}
[...]

PR #3006, opened 2022-12-09
Scheduled attribute updated in handleRecurrence.

scompo:
Hi, I'm using the scheduled attribute in a recurring task, and noticed it's not updated when creating the next task.

I think it's a known problem, and related to #1534.

#### Description

I've handled the `scheduled` attribute in the same way of `wait` when handling the recurrence.
Just added the difference from the original `due` and `scheduled` to the new `due` date as the `scheduled`.

- I've modified the C++ code but I'm not able to build the code on aarch64, so I haven't tested it.


tbabej:
@scompo This looks good on my end. Do you think you could add support for the `until` attribute as well? That way we'd fully address the issue in #1534.

As far as the build failure on `aarch64`, could you perhaps open up a separate issue about that? We are striving to support multiple architectures, including this one.

scompo:
@tbabej Sure, I'll look into updating the `until` flag too. From the documentation it looks like it should be updated the same as both `wait` and `scheduled`. Anything else I should worry about it when upgrading it?

As for for the compilation problem I'll open up a new issue with the error as soon as I'm able to

djmitche:
@tbabej do you remember the status here? Was this worthwhile merging without the `until` support?

djmitche:
I'm going to close this as stale. It's still here and can be re-opened!

ashprice:
Hmm. In addition to the ongoing interative tasks development, I was wondering if a temporary fix would be desired, maybe just merging the `scheduled`-related commits? (I'd be happy to take a look at them and merge them relative to any changes needed since this PR was closed, though I'm a bit of git noob so I'm not sure of the 'technically correct' way to maintain commit history referring to @scompo.)

For `until` - is it actually desired that, pending iterative tasks landing, that we want an until date to be such that until is `current_task_until = current_task_due + (original_template_until - original_template_due)`? 

Because I actually find the current way where `current_task_until = original_template_until` to be _more_ intuitive, whatever the original documentation suggested. 

The iterative RRULES will handle both cases elegantly I think, but changing `until` for recurring tasks at this point might be undesirable, because it could break the workflows of those who aren't using hooks to address this (and for those who are using hooks to address this, it will break their hooks.) 

It seems people have different ideas about this, cf. discussion at https://github.com/GothenburgBitFactory/taskwarrior/issues/1534#issuecomment-883496041. I don't really mind myself, though the lack of an `until` working how it currently does would make a `runtil` a necessity. 

But debate about what we should do with `until` is little reason to not fix `scheduled`. Of course, it would still break hooks that people are using to address it, but it's already in kind of a weird place, where `wait` is fixed but `scheduled` is not.

djmitche:
Yeah, updating `scheduled` makes sense and is unlikely to break things. I agree that `until` is debatable and in that case it's best to not change existing functionality.

Once we have config in the replica, we could potentially add config for things like this so that people can configure it how they'd like.

PR #3015, opened 2022-12-18
tests: Fail OS-X test workflow on error in the build process

### LINKED ISSUE #3012: Github workflow silently ignores test suite build errors on OS-X
As [discussed](https://github.com/GothenburgBitFactory/taskwarrior/pull/3009#issuecomment-1356385221) in PR #3009, the test suite currently silently ignores failures to build complete test suite. We should make sure to flag that scenario as a failure in the Github workflow itself.

tbabej:
This sets `set -e` to ensure any errors in the test script are propagated upwards and noticed by the Github workflow.

Prior to these changes failures in the build process of the tests would be ignored, and the success / failure would be determined purely by the return code of the `./problems` script.

Closes #3012

### DIFF
--- a/test/scripts/test_osx.sh
+++ b/test/scripts/test_osx.sh
@@ -1,4 +1,4 @@
-set -x
+set -ex
 
 brew install gnutls
 brew install cmake


djmitche (approved):

tbabej:
The tests are failing here, but that's the behaviour we _want_ to see. This should not happen on the `develop` branch, as #3011 fixes the issue.

PR #3019, opened 2022-12-19
use non-deprecated chrono functions

### LINKED ISSUE #3016: Fix warning regarding chrono::TimeZone::timestamp
The following warnings are being emitted on Rust builds:
```
warning: use of deprecated associated function `chrono::TimeZone::timestamp`: use `timestamp_opt()` instead
   --> /home/dustin/p/taskwarrior/taskchampion/taskchampion/src/task/task.rs:230:36
    |
230 |                         entry: Utc.timestamp(ts, 0),
    |                                    ^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `chrono::TimeZone::timestamp`: use `timestamp_opt()` instead
   --> /home/dustin/p/taskwarrior/taskchampion/taskchampion/src/task/task.rs:313:33
    |
313 |                 return Some(Utc.timestamp(ts, 0));
    |                                 ^^^^^^^^^

   Compiling taskchampion-lib v0.1.0 (/home/dustin/p/taskwarrior/taskchampion/lib)
warning: use of deprecated associated function `taskchampion::chrono::TimeZone::timestamp`: use `timestamp_opt()` instead
  --> /home/dustin/p/taskwarrior/taskchampion/lib/src/atomic.rs:26:22
   |
26 |             Some(Utc.timestamp(self as i64, 0))
   |                      ^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `taskchampion::chrono::TimeZone::timestamp`: use `timestamp_opt()` instead
   --> /home/dustin/p/taskwarrior/taskchampion/lib/src/task.rs:751:40
    |
751 |             task.remove_annotation(Utc.timestamp(entry, 0))?;
    |                                        ^^^^^^^^^

warning: `taskchampion` (lib) generated 2 warnings
warning: `taskchampion-lib` (lib) generated 2 warnings
```

djmitche:
Fixes #3016.

### DIFF
--- a/taskchampion/lib/src/task.rs
+++ b/taskchampion/lib/src/task.rs
@@ -5,7 +5,7 @@ use std::convert::TryFrom;
 use std::ops::Deref;
 use std::ptr::NonNull;
 use std::str::FromStr;
-use taskchampion::chrono::{TimeZone, Utc};
+use taskchampion::chrono::{offset::LocalResult, TimeZone, Utc};
 use taskchampion::{Annotation, Tag, Task, TaskMut, Uuid};
 
 /// A task, as publicly exposed by this library.
@@ -760,7 +760,9 @@ pub unsafe extern "C" fn tc_task_remove_annotation(task: *mut TCTask, entry: i64
     wrap_mut(
         task,
         |task| {
-            task.remove_annotation(Utc.timestamp(entry, 0))?;
+            if let LocalResult::Single(ts) = Utc.timestamp_opt(entry, 0) {
+                task.remove_annotation(ts)?;
+            }
             Ok(TCResult::Ok)
         },
         TCResult::Error,
--- a/taskchampion/lib/src/atomic.rs
+++ b/taskchampion/lib/src/atomic.rs
@@ -1,7 +1,7 @@
 //! Trait implementations for a few atomic types
 
 use crate::traits::*;
-use taskchampion::chrono::prelude::*;
+use taskchampion::chrono::{offset::LocalResult, prelude::*};
 
 impl PassByValue for usize {
     type RustType = usize;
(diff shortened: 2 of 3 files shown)


taskchampion/lib/src/task.rs
```diff
@@ -760,7 +760,9 @@ pub unsafe extern "C" fn tc_task_remove_annotation(task: *mut TCTask, entry: i64
     wrap_mut(
         task,
         |task| {
-            task.remove_annotation(Utc.timestamp(entry, 0))?;
+            if let LocalResult::Single(ts) = Utc.timestamp_opt(entry, 0) {
+                task.remove_annotation(ts)?;
+            }
```
ryneeverett, line 765: Does it make sense to return `Ok` if we get a `LocalResult::Ambiguous` and the annotation is not removed?

djmitche: AIUI LocalResult::Ambiguous occurs when a timezone has "savings time" and, for example, there are two different times known as "2am".  Because we're explicitly using Utc here, and Utc doesn't do the "savings time" thing, that should never occur.

ryneeverett: Since that condition should never occur, wouldn't it be better to panic than to return `Ok`?

djmitche: TBH that's what `unreachable!()` is for (which, yes, does panic).  Do you want to make a PR to do that?

ryneeverett: Sure! I hadn't heard of `unreachable!()` so that will be a good learning opportunity.

ryneeverett: So my first thought was to wrap this up into a function:

```rs
pub fn utc_timestamp(secs: i64) -> Option<Timestamp> {
    match Utc.timezone_opt(secs, 0) {
        LocalResult::Single(tz) => tz,
        None => None,
        LocalResult::Ambiguous => unreachable!("We're requesting UTC so there shouldn't be any daylight saving time."),
    }
}
```

Then I realized that there is no pattern for sharing utility functions like that between crates. Should there be a sixth "private-utils" crate for stuff like this? I figure there's probably a good reason you haven't done something like that already. Is it better to just copy the `match` block?

djmitche: `taskchampion-lib` already depends on `taskchampion`, so maybe the best plan is to put this utility function in the `taskchampion` crate, maybe in a `time` sub-module.  Then `taskchampion-lib` can call it from there.

PR #3028, opened 2022-12-22
Only consider tasks dependent if both are pending

djmitche:
Previously, if a pending task depended on _any_ task, that relationship was included in the dependency map.

### DIFF
--- a/taskchampion/taskchampion/src/replica.rs
+++ b/taskchampion/taskchampion/src/replica.rs
@@ -121,27 +121,37 @@ impl Replica {
 
             let mut dm = DependencyMap::new();
             // temporary cache tracking whether tasks are considered Pending or not.
-            let mut is_pending = HashMap::new();
+            let mut is_pending_cache = HashMap::new();
             let ws = self.working_set()?;
+            // for each task in the working set
             for i in 1..=ws.largest_index() {
+                // get the task UUID
                 if let Some(u) = ws.by_index(i) {
+                    // get the task
                     if let Some(taskmap) = self.taskdb.get_task(u)? {
+                        // search the task's keys
                         for p in taskmap.keys() {
+                            // for one matching `dep_..`
                             if let Some(dep_str) = p.strip_prefix("dep_") {
+                                // and extract the UUID from the key
                                 if let Ok(dep) = Uuid::parse_str(dep_str) {
+                                    // the dependency is pending if
                                     let dep_pending = {
-                                        if let Some(dep_pending) = is_pending.get(&dep) {
+                                        // we've determined this before and cached the result
+                                        if let Some(dep_pending) = is_pending_cache.get(&dep) {
                                             *dep_pending
                                         } else if let Some(dep_taskmap) =
+                                            // or if we get the task
                                             self.taskdb.get_task(dep)?
                                         {
+                                            // and its status is "pending"
                                             let dep_pending = matches!(
                                                 dep_taskmap
                                                     .get("status")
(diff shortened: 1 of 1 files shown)


taskchampion/taskchampion/src/replica.rs
```diff
@@ -113,17 +116,40 @@ impl Replica {
...
+                                            let dep_pending = matches!(
+                                                dep_taskmap
+                                                    .get("status")
+                                                    .map(|tm| Status::from_taskmap(tm)),
+                                                Some(Status::Pending)
+                                            );
+                                            is_pending.insert(dep, dep_pending);
+                                            dep_pending
+                                        } else {
+                                            false
```
ryneeverett, line 157: The only way to reach this `else`/`false` branch is if `self.taskdb.get_task(dep)` returns `None`, which means there is a dangling reference to a dependency task that has been deleted, right? Theoretically it could be due to a race condition but I'm assuming that's a rare case.

Do we know how those are normally handled? Does `task delete` search for dependents and delete references, are they garbage collected, or would this be an appropriate place to delete them?

taskchampion/taskchampion/src/replica.rs
```diff
@@ -113,17 +116,40 @@ impl Replica {
     /// although previously-returned dependency maps are not updated.
     pub fn dependency_map(&mut self, force: bool) -> anyhow::Result<Rc<DependencyMap>> {
         if force || self.depmap.is_none() {
+            // note: we can't use self.get_task here, as that depends on a
+            // DependencyMap
+
             let mut dm = DependencyMap::new();
+            // temporary cache tracking whether tasks are considered Pending or not.
+            let mut is_pending = HashMap::new();
```
ryneeverett: I would find this more clear if the fact that it was a cache was part of the name and if the types were explicit:

```rs
let mut is_pending_cache = HashMap::<Uuid, bool>::new();
```

The comment would then be harmless but perhaps superfluous.

taskchampion/taskchampion/src/replica.rs
djmitche: `task delete` does try to delete dependencies, iirc.  However, with `task sync` there are race conditions on the order of hours or days, so it's easy for that to be incorrect.  I think `task diag` will notice that this is incorrect.  But, in general, we want to treat a dangling dependency as if it didn't exist, which is what I (was trying to) do here.  

More broadly, I'd like to get to a place where any key/value map can be treated as a task without raising errors, since the mayhem `task sync` races can wreak is quite substantial.  TC adheres to this principle, but TW will still crash on some forms of task.

taskchampion/taskchampion/src/replica.rs
djmitche: Done :)

taskchampion/taskchampion/src/replica.rs
ryneeverett: > However, with task sync there are race conditions on the order of hours or days...

You lost me here. I understand that databases might go a long time without being synchronized, but wouldn't each replica be internally consistent? Are you talking about ultra-low-bandwidth situations where it might take hours or days? Or can interrupted `task sync` result in partial synchronization?

taskchampion/taskchampion/src/replica.rs
ryneeverett: Can we get the explicit types too?

djmitche: Oh, I missed that.

ryneeverett (approved):

PR #3029, opened 2022-12-23
Do not further update 'modified' if it's explicitly set

djmitche:
This is a minor correctness fix, but helps avoid inconsistencies when things like Taskwarrior call both `set_value` and other, higher-level `set_..` methods like `set_status`.

### DIFF
--- a/taskchampion/taskchampion/src/task/task.rs
+++ b/taskchampion/taskchampion/src/task/task.rs
@@ -1199,7 +1199,7 @@ mod test {
     }
 
     #[test]
-    fn set_value_modiifed() {
+    fn set_value_modified() {
         with_mut_task(|mut task| {
             // set the modified property to something in the past..
             task.set_value("modified", Some("1671820000".into()))
--- a/taskchampion/taskchampion/src/task/task.rs
+++ b/taskchampion/taskchampion/src/task/task.rs
@@ -372,8 +372,17 @@ impl<'r> TaskMut<'r> {
     }
 
     /// Set a tasks's property by name.
+    ///
+    /// This will not automatically update the `modified` timestamp or perform any other
+    /// "automatic" operations -- it simply sets the property. Howerver, if property is
+    /// "modified", then subsequent calls to other `set_..` methods will not update the
+    /// `modified` timestamp.
     pub fn set_value<S: Into<String>>(&mut self, property: S, value: Option<String>) -> Result<()> {
         let property = property.into();
+
+        if &property == "modified" {
+            self.updated_modified = true;
+        }
 
         if let Some(ref v) = value {
             trace!("task {}: set property {}={:?}", self.task.uuid, property, v);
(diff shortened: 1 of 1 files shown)


taskchampion/taskchampion/src/task/task.rs
```diff
@@ -1188,4 +1197,17 @@ mod test {
         assert!(t2.has_tag(&stag(SyntheticTag::Unblocked)));
         assert!(t2.has_tag(&stag(SyntheticTag::Blocking)));
     }
+
+    #[test]
+    fn set_value_modiifed() {
```
ryneeverett: typo

ryneeverett (commented):
I understand the change and why it fixes the test behavior but I don't understand why the test behavior is desirable. Are there places in the code base "modified" is explicitly set? I don't really understand the "inconsistencies" issue here. How do the inconsistencies manifest themselves?

djmitche:
The problem that brought me here is, Taskwarrior _mostly_ doesn't call functions like `set_description` or `add_annotation` -- it interfaces directly with the key/value pairs that make up a task, via `set_value`.  But it does call `set_status` since that updates the working set.  So the issue was that Taskwarrior would call

 * `t.set_value("modified", "12345")`
 * `t.set_status(Status::Pending)`

and the second call would not "see" that the modified timestamp had already been set, and would do so again.  On import, this causes a problem by setting the "modified" timestamp of all tasks to the time the import occurred.

Overall, this is a major "impedance mismatch" between TC and TW: TW wants to handle tasks as key/value maps, while TC was really designed for API consumers to call methods that embody more knowledge about the meaning of particular keys.  These are more than just convenience wrappers, though: they also have this set-the-modified-timestamp thing, as well as automatically updating the working set.  And (referring to #3028) it might be good to cache parent/child relationships as well, to avoid having to load most of the pending tasks one-by-one.  I've been mulling this over for a while but haven't come up with a concrete way to do better.

ryneeverett:
>  On import, this causes a problem by setting the "modified" timestamp of all tasks to the time the import occurred.

Sorry to drag you into the rabbit hole of my ignorance, but why does an import call `set_status`? It seems like an imported task would go straight into the database without ever being a `TaskMut`.

djmitche:
All database modifications are made by applying "operations" and this includes imports -- that way if you import some tasks and then `task sync`, those imported tasks show up on your other replicas.  So there's no "straight to the database".

Taskwarrior calls `TaskMut::set_status` because that function has the side-effect of updating the WorkingSet appropriately.  The WorkingSet maps short integers to task IDs.

djmitche:
(As I suggested above, I don't think this is great, but that's how it is :) )

ryneeverett:
Thanks for the explanation.

It seems like there's a missing abstraction layer and the parts of `TaskMut` which update "modified" could be broken out into a higher level `struct TaskModified { task: TaskMut}` which uses it's `TaskMut` pointer to call the primitive "operations".

ryneeverett (approved):
Regardless, let's not let the perfect be the enemy of the good.

djmitche:
Yeah, I don' think we need to figure that out on this PR, but I like the direction you're thinking!  Something like a high-level vs. low-level interface.

PR #3039, opened 2023-01-05
Add workflow for Taskwarrior Docker image

lauft:
#### Description
Add workflow for Taskwarrior Docker image

The workflow is triggered by a successful run of the test suite and creates a Docker image with a Taskwarrior installation of the current branch (restricted to develop/stable)

The main intention is to have a base image to test the on-modify hook for Timewarrior (see [here](https://github.com/lauft/task-on-modify-hook)), but it may also be useful for other extension – or just to try out Taskwarrior.

  No changes to present code or build infrastructure

  No changes to present code or build infrastructure

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -1,0 +1,53 @@
+name: Taskwarrior Docker image
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: [tests]
+    branches:
+      - develop
+      - stable
+    types:
+      - completed
+
+env:
+  REGISTRY: "ghcr.io"
+
+jobs:
+  build-and-push-docker-image:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v2.8.1
+
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@v2.1.0
+        with:
+          registry: ${{ env.REGISTRY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-image.yaml
```diff
@@ -0,0 +1,53 @@
...
+        uses: sigstore/cosign-installer@v2.8.1
+
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@v2.1.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Timewarrior Docker image
```
heiderich:
```suggestion
      - name: Build and push Taskwarrior Docker image
```

lauft: Whoopsie... must have been late 😅

docker/task.dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:22.04 AS base
+
+FROM base AS builder
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN ( apt-get update && yes | unminimize ) && \
```
djmitche: Was there something in particular that `unminimize` pulls in?  Otherwise I expect this will end up downloading a bunch of unused stuff?

djmitche (approved):
This looks good.  The question below is just a question :)

docker/task.dockerfile
lauft: The usage of `uniminimize` stems from the Timewarrior (test) image. Without it, you get the following message when calling e.g. man pages:
```
This system has been minimized by removing packages and content that are not required on a system that users do not log into.
To restore this content, including manpages, you can run the 'unminimize' command. You will still need to ensure the 'man-db' package is installed.
```
So I guess one can remove this from the builder image.

djmitche: Ah, manpages makes sense. And yeah, not necessary for the temporary builder image :)

PR #3047, opened 2023-01-14
taskchampion: use taskchampion::Error

### LINKED ISSUE #2878: [taskchampion] Return `Result<.., taskchampion::Error>` from Replica API
See https://github.com/taskchampion/taskchampion/issues/226

ryneeverett:
Resolve #2878

Let me know if I completely missed the mark on what you were going for in that issue. I'm just trying to learn Rust at this point so I'm sure there are some anti-patterns here as well.

#### Description

Return the `taskchampion::Error` type from the public API.

### DIFF
--- a/taskchampion/taskchampion/src/errors.rs
+++ b/taskchampion/taskchampion/src/errors.rs
@@ -19,20 +19,25 @@ pub enum Error {
     /// A usage error
     #[error("User Error: {0}")]
     Usage(String),
-
-    /// Error conversions.
-    #[error(transparent)]
-    Http(#[from] ureq::Error),
-    #[error(transparent)]
-    Io(#[from] io::Error),
-    #[error(transparent)]
-    Json(#[from] serde_json::Error),
+    /// A general error.
     #[error(transparent)]
     Other(#[from] anyhow::Error),
-    #[error("Third Party Sqlite Error")]
-    Rusqlite(#[from] rusqlite::Error),
-    #[error(transparent)]
-    Sqlite(#[from] crate::storage::sqlite::SqliteError),
 }
+
+/// Convert private and third party errors into Error::Other.
+macro_rules! convert_error {
+    ( $error:ty ) => {
+        impl From<$error> for Error {
+            fn from(err: $error) -> Self {
+                Self::Other(err.into())
+            }
+        }
+    };
+}
+convert_error!(ureq::Error);
+convert_error!(io::Error);
+convert_error!(serde_json::Error);
+convert_error!(rusqlite::Error);
+convert_error!(crate::storage::sqlite::SqliteError);
 
 pub type Result<T> = std::result::Result<T, Error>;
--- a/taskchampion/taskchampion/src/errors.rs
+++ b/taskchampion/taskchampion/src/errors.rs
@@ -7,7 +7,7 @@ use thiserror::Error;
 pub enum Error {
     /// A crypto-related error
     #[error("Crypto Error: {0}")]
-    Crypto(String),
+    Server(String),
     /// A task-database-related error
     #[error("Task Database Error: {0}")]
     Database(String),
@@ -18,7 +18,7 @@ pub enum Error {
     OutOfSync,
     /// A usage error
     #[error("User Error: {0}")]
-    UserError(String),
+    Usage(String),
 
     /// Error conversions.
     #[error(transparent)]
@@ -35,4 +35,4 @@ pub enum Error {
     Sqlite(#[from] crate::storage::sqlite::SqliteError),
 }
 
-pub type Result<T> = core::result::Result<T, Error>;
+pub type Result<T> = std::result::Result<T, Error>;
--- a/taskchampion/taskchampion/src/errors.rs
+++ b/taskchampion/taskchampion/src/errors.rs
@@ -5,8 +5,8 @@ use thiserror::Error;
 #[non_exhaustive]
 /// Errors returned from taskchampion operations
 pub enum Error {
-    /// A crypto-related error
-    #[error("Crypto Error: {0}")]
+    /// A server-related error
+    #[error("Server Error: {0}")]
     Server(String),
     /// A task-database-related error
     #[error("Task Database Error: {0}")]
@@ -17,7 +17,7 @@ pub enum Error {
     #[error("Local replica is out of sync with the server")]
     OutOfSync,
     /// A usage error
-    #[error("User Error: {0}")]
+    #[error("Usage Error: {0}")]
     Usage(String),
     /// A general error.
     #[error(transparent)]
@@ -25,7 +25,7 @@ pub enum Error {
 }
 
 /// Convert private and third party errors into Error::Other.
-macro_rules! convert_error {
+macro_rules! other_error {
     ( $error:ty ) => {
         impl From<$error> for Error {
             fn from(err: $error) -> Self {
@@ -34,10 +34,10 @@ macro_rules! convert_error {
         }
     };
 }
-convert_error!(ureq::Error);
-convert_error!(io::Error);
-convert_error!(serde_json::Error);
-convert_error!(rusqlite::Error);
-convert_error!(crate::storage::sqlite::SqliteError);
+other_error!(ureq::Error);
+other_error!(io::Error);
+other_error!(serde_json::Error);
+other_error!(rusqlite::Error);
+other_error!(crate::storage::sqlite::SqliteError);
 
 pub type Result<T> = std::result::Result<T, Error>;
--- a/taskchampion/taskchampion/src/errors.rs
+++ b/taskchampion/taskchampion/src/errors.rs
@@ -1,9 +1,13 @@
+use std::io;
 use thiserror::Error;
 
-#[derive(Debug, Error, Eq, PartialEq, Clone)]
+#[derive(Debug, Error)]
 #[non_exhaustive]
 /// Errors returned from taskchampion operations
 pub enum Error {
+    /// A crypto-related error
+    #[error("Crypto Error: {0}")]
+    Crypto(String),
     /// A task-database-related error
     #[error("Task Database Error: {0}")]
     Database(String),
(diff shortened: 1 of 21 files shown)


taskchampion/taskchampion/src/errors.rs
```diff
@@ -1,9 +1,13 @@
...
 use thiserror::Error;
 
-#[derive(Debug, Error, Eq, PartialEq, Clone)]
+#[derive(Debug, Error)]
 #[non_exhaustive]
 /// Errors returned from taskchampion operations
 pub enum Error {
+    /// A crypto-related error
+    #[error("Crypto Error: {0}")]
+    Crypto(String),
```
djmitche: It looks like the places this is used are more generally "server-related", so maybe `Server(..)`?

taskchampion/taskchampion/src/errors.rs
```diff
@@ -12,4 +16,23 @@ pub enum Error {
     /// other irrecoverable error.
     #[error("Local replica is out of sync with the server")]
     OutOfSync,
+    /// A usage error
+    #[error("User Error: {0}")]
+    UserError(String),
```
djmitche: I like the word "Usage" here -- there may not be a "user" involved, but it indicates the usage of some function was incorrect.

taskchampion/taskchampion/src/errors.rs
```diff
@@ -12,4 +16,23 @@ pub enum Error {
...
+    Json(#[from] serde_json::Error),
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+    #[error("Third Party Sqlite Error")]
+    Rusqlite(#[from] rusqlite::Error),
+    #[error(transparent)]
+    Sqlite(#[from] crate::storage::sqlite::SqliteError),
 }
+
+pub type Result<T> = core::result::Result<T, Error>;
```
djmitche: Since this crate uses `std` all over the place, you can refer to this as `std::result::Result`.  (The `std` crate just re-exports everything in `core`, so they're precisely the same type)

djmitche (changes requested):
Nice work!  This is a good refactor!

My concern here is that this puts types like `ureq::Error`, `rusqlite::Error` and `SqliteError` into the public signature.  So for example a library user might start matching on `Http(ureq_err)` and getting the URL and response code from the error for display.  That means that switching to another HTTP library becomes a breaking change.  Similarly for rusqlite, or for changing what had been a non-public error enum, `SqliteError`.

There's been as much variation in advice on how to handle errors in Rust as there has been packaging standards for Python, but one of principles I have found useful is to provide the minimal useful data to a caller in an error object.

In this case, I think that adding `Crypto` (renamed `Server`) and `UserError` (renamed `UsageError`) makes sense, as those indicate useful, broad classes of errors.  For example, a tool embedding TC might want to retry something that failed with a server error, but assume that any other errors indicate corrupted local state and try to re-initialize.

The rest, I think, should be lumped into the `Other` option.  The Rust-y challenge is to do that in an ergonomic way, so that `something_that_returns_ureq_errror()?` in a function returning `crate::errors::Result<..>` will compile without requiring any additional `.map_err(..)` calls.

ryneeverett:
Ok, I was confused by the distinction between thiserror's `#[from]`/`source` style versus `#[error(transparent)]` and I think your concern makes it a bit more clear. Maybe the solution is to create an `IntermediaryError` enum that catches all these third party errors and the anyhow errors and then redirect all the `IntermediaryError`'s to `Error::Other`.

djmitche:
Hm, the `thiserror` docs are a bit confusing on the topic!  I verified by generating docs for this PR (`cargo doc -p taskchampion --open`) and seeing references to `rusqlite::Error` and others in the generated docs.

I think your suggestion is similar to the last paragraph of the `thiserror` docs -- "Another use case is hiding implementation details..".  I tried playing around a little bit with implementing `From<E> for taskchampion::Error` where E can be converted to an `anyhow::Error`, but `taskchampion::Error` itself can make such a conversion, so it conflicts with the global `impl From<T> for T`.  Give it a try!

ryneeverett:
Sorry but I don't understand what you're suggesting I try. I assumed it would be best to keep `anyhow::Error` private and expose `taskchampion::Error` but it seems like you're suggesting the opposite.

To attempt to clarify the problem: what we want is similar to `transparent` -- "to forward the source and Display methods straight through to an underlying error without adding an additional message" -- but we want to forward the Display/message without forwarding the source/type. It isn't clear to me what the pattern is to expose only our single `Error` enum to the public and at the same time satisfy the compiler that our private errors are valid `Result`s.

Here's what I tried:

```rs
use std::io;
use thiserror::Error;

#[derive(Debug, Error)]
/// Contains private errors
pub enum OtherErrorRepr {
    #[error(transparent)]
    Http(#[from] ureq::Error),
    #[error(transparent)]
    Io(#[from] io::Error),
    #[error(transparent)]
    Json(#[from] serde_json::Error),
    #[error(transparent)]
    Other(#[from] anyhow::Error),
    #[error("Third Party Sqlite Error")]
    Rusqlite(#[from] rusqlite::Error),
    #[error(transparent)]
    Sqlite(#[from] crate::storage::sqlite::SqliteError),
}

#[derive(Error, Debug)]
#[error(transparent)]
/// Exposes a single public error
pub struct OtherError(#[from] OtherErrorRepr);

#[derive(Debug, Error)]
#[non_exhaustive]
/// Errors returned from taskchampion operations
pub enum Error {
    /// A crypto-related error
    #[error("Crypto Error: {0}")]
    Server(String),
    /// A task-database-related error
    #[error("Task Database Error: {0}")]
    Database(String),
    /// Private errors.
    #[error(transparent)]
    Other(#[from] OtherError),
    /// An error specifically indicating that the local replica cannot
    /// be synchronized with the sever, due to being out of date or some
    /// other irrecoverable error.
    #[error("Local replica is out of sync with the server")]
    OutOfSync,
    /// A usage error
    #[error("User Error: {0}")]
    Usage(String),
}

pub type Result<T> = std::result::Result<T, Error>;
```

This gives a bunch of compiler errors like "the trait `From<serde_json::Error>` is not implemented for `errors::Error`". I thought this would work because `From<OtherError>` is implemented for `errors:Error`, `From<OtherErrorRepr>` is implemented for `OtherError`, and `From<serde_json::Error>` is implemented for `OtherErrorRepr`. Shouldn't the compiler be able to follow this chain?

djmitche:
I admit I'm not 100% sure how to do this either, but I'll do some experimentation to see how to make it work.

My basic idea is that `anyhow` allows converting any error into `anyhow::Error`.  So, for errors where we don't want to expose their type, we should just convert to an `anyhow::Error` and then wrap that in `Other(anyhow_error)`.

djmitche:
Hm, so this is what I had i mind, implemented just for `ureq::Error`.
```
index fb2bae028..759ede44f 100644
--- a/taskchampion/taskchampion/src/errors.rs
+++ b/taskchampion/taskchampion/src/errors.rs
@@ -17,22 +17,26 @@ pub enum Error {
     #[error("Local replica is out of sync with the server")]
     OutOfSync,
     /// A usage error
     #[error("User Error: {0}")]
     Usage(String),
 
     /// Error conversions.
     #[error(transparent)]
-    Http(#[from] ureq::Error),
-    #[error(transparent)]
     Io(#[from] io::Error),
     #[error(transparent)]
     Json(#[from] serde_json::Error),
     #[error(transparent)]
     Other(#[from] anyhow::Error),
     #[error("Third Party Sqlite Error")]
     Rusqlite(#[from] rusqlite::Error),
     #[error(transparent)]
     Sqlite(#[from] crate::storage::sqlite::SqliteError),
 }
+impl From<ureq::Error> for Error {
+    fn from(err: ureq::Error) -> Self {
+        Self::Other(err.into())
+    }
+}
+
```

I can't figure out a way to make this generic over any error type (that is, anything implementing `std::error::Error`) -- trait bounds don't have a way to say "all types matching the `where` clause _except_ ..".  So, maybe we just need this four-line `impl From` for each type we want to convert to `Other(..)`?

taskchampion/taskchampion/src/errors.rs
```diff
@@ -1,9 +1,13 @@
+use std::io;
 use thiserror::Error;
 
-#[derive(Debug, Error, Eq, PartialEq, Clone)]
```
ryneeverett, line 3: Is it ok that these traits were dropped? I removed them as a quick fix but it also wasn't clear to me what the benefit was. Or is it just best practice to include all possible derivable traits so you're conscientious of losing them?

ryneeverett (commented):
A couple other thoughts upon self-reviewing this:

- Does lumping `SqliteError` into `Other` make sense or should it be passed through transparently? Or should it be converted to `Database`?
- My attack on  `anyhow` was somewhat arbitrary. Should we continue this work (presumably in a separate PR) to replace `anyhow` with `errors`? Are there further places in the public API it is exposed which would make this worthwhile?

taskchampion/taskchampion/src/errors.rs
djmitche: There are best practices for deriving traits when they are practical on public types.  Especially Debug is useful for consumers of a type.

djmitche:
I think `anyhow::Error` is actually a really useful type in general, and for many libraries I'd be happy to just always return `anyhow::Error`.  For TC in particular, there were a few errors we wanted the user to be able to recognize explicitly, leading to the custom type and #2878 to actually return that custom type.

taskchampion/taskchampion/src/errors.rs
djmitche: ..but if the derivation isn't automatic anymore, which I guess was the case here, then it's fine to omit them.

taskchampion/taskchampion/src/errors.rs
```diff
@@ -12,4 +16,28 @@ pub enum Error {
...
+                Self::Other(err.into())
+            }
+        }
+    };
+}
+other_error!(ureq::Error);
+other_error!(io::Error);
+other_error!(serde_json::Error);
+other_error!(rusqlite::Error);
+other_error!(crate::storage::sqlite::SqliteError);
```
djmitche, line 41: nice!

djmitche (approved):

PR #3056, opened 2023-01-28
simpletap: simplify to use parent `run` method

ryneeverett:
#### Description

Per the TODO, use `__init__` to override the stream. There don't seem to be any other substantial changes to the `run` method needed for TAP and I'm assuming the additional code in the upstream `run` has been added since this class was written.

My primary motivation was to set `result.buffer = self.buffer` in the `run` method, which fixes the runner enough to be able to use `breakpoint()`. (But this is taken care of by the parent class when we remove `run`.)

In addition, I updated the shebang in `tw-2189.t` for portability.

### DIFF
--- a/test/simpletap/__init__.py
+++ b/test/simpletap/__init__.py
@@ -30,7 +30,6 @@
 import os
 import sys
 import unittest
-import warnings
 import traceback
 import inspect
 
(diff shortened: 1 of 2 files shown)


djmitche (approved):
Looks good!  I see this code in `unittest.runner` in, at least, Python3.8.  And everything is passing, so this should be fine :)

PR #3061, opened 2023-02-05
taskchampion: Add time.utc_timestamp function.

ryneeverett:
Per discussion in #3019.

Something I hadn't realized in that discussion was that, by my reading of the docs, `None` is also only returned due to daylight saving time issues, so that should also be unreachable.

Additional considerations for potential further improvement:

1. Would it be more appropriate to make our own `Timestamp` than to use it as an alias for `DateTime<Utc>`? Reading https://lurklurk.org/effective-rust/re-export.html:

> As an aside, think carefully before using another crate's types in your API: it intimately ties your crate to that of the dependency.

I think the point may be overstated, but perhaps a consideration. As an additional benefit we could use the `From` trait so we would only have to pass around `Timestamp` and not an additional function.

2. If we're going to keep `Timestamp` as an alias, would it be better to define a new trait (e.g. `FromSecs`) than to pass around this helper function?

#### Description

Add a function that returns a Timestamp from an i64. One advantage is improved readability since this function is guaranteed to return a `LocalResult::Single`. Anther advantage is that it will panic if something other than a LocalResult::Single is returned by chrono, which shouldn't be possible for UTC timestamps which can't have DST ambiguity.

### DIFF
--- a/taskchampion/lib/src/atomic.rs
+++ b/taskchampion/lib/src/atomic.rs
@@ -1,7 +1,8 @@
 //! Trait implementations for a few atomic types
 
 use crate::traits::*;
-use taskchampion::chrono::{offset::LocalResult, prelude::*};
+use taskchampion::chrono::{DateTime, Utc};
+use taskchampion::utc_timestamp;
 
 impl PassByValue for usize {
     type RustType = usize;
@@ -21,9 +22,7 @@ impl PassByValue for libc::time_t {
 
     unsafe fn from_ctype(self) -> Option<DateTime<Utc>> {
         if self != 0 {
-            if let LocalResult::Single(ts) = Utc.timestamp_opt(self, 0) {
-                return Some(ts);
-            }
+            return Some(utc_timestamp(self));
         }
         None
     }
--- a/taskchampion/lib/src/task.rs
+++ b/taskchampion/lib/src/task.rs
@@ -5,8 +5,7 @@ use std::convert::TryFrom;
 use std::ops::Deref;
 use std::ptr::NonNull;
 use std::str::FromStr;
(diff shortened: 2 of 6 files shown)


djmitche (approved):
This looks great!

>     1. Would it be more appropriate to make our own `Timestamp` than to use it as an alias for `DateTime<Utc>`? Reading https://lurklurk.org/effective-rust/re-export.html:
> 
> 
> > As an aside, think carefully before using another crate's types in your API: it intimately ties your crate to that of the dependency.
> 
> I think the point may be overstated, but perhaps a consideration. As an additional benefit we could use the `From` trait so we would only have to pass around `Timestamp` and not an additional function.

I think the next sentence applies here as well:

>> In this case, rand is a semi-standard crate that is high quality and widely used

There were some issues with `chrono` a year or so ago, but really those were the "fault" of a poorly designed libc interface, and were sorted eventually.  If we were to "wrap" the type (say, one that wrapped i64 as seconds since the epoch), then users would need to "unwrap" it using some conversion, and in almost every case they'd be "unwrapping" back to `chrono::DateTime`.

>     2. If we're going to keep `Timestamp` as an alias, would it be better to define a new trait (e.g. `FromSecs`) than to pass around this helper function?

That is definitely an option.  But the functionality would only be internal to the crate and only implement one minor utility function, so it would have minimal benefit.  It's also a "constructor" function rather than a method, and traits are designed around methods.  And finally, traits -- and especially From and Into -- introduce a bit of "magic" that can reduce readability.  Given `ts.into()`, it's hard for the reader to figure out where `into` is defined.

Basically, I think a utility function is fine, even if it's less fancy :)

PR #3089, opened 2023-04-08
Update Actix crates to latest versions

djmitche:
This avoids a vulnerability in tokio (#3085). The major version updates of both actix-web and actix-rt required some signficant changes. Chief among those, it turns out we were relying on actix-rt to run the HttpServer in a different thread from the rest of the test, so that we could talk to it from sync code in the test thread. This no longer works, so the sync code is now run in a dedicated thread with `actix_rt::task::spawn_blocking`.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           # Fixed version for clippy lints.  Bump this as necesary.  It must not
           # be older than the MSRV in tests.yml.
-          toolchain: "1.61"
+          toolchain: "1.63"
           override: true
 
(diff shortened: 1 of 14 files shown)


djmitche:
(mac test failures are unrelated)

tbabej (approved):

tbabej:
@djmitche The Mac OS-X failures now happening on `develop`?

djmitche:
The error is
```
2023-04-08T20:59:43.2953730Z not ok 4 - tw-262.t: project.isnt works within parenthesis after a double leading space
```
which I assume is due to some flakiness? It definitely doesn't involve the sync server..

PR #3092, opened 2023-04-16
[BREAKING] Disable the 'sync' command

djmitche:
Sync will require additional support from TaskChampion. This will be added before the next release.


ryneeverett:
Sorry but it's hard to understand the roadmap here and I question how anyone could really review this. Specifically, how does this deletion facilitate the transition to taskchampion if the `sync` command hasn't been implemented yet?

djmitche:
In general, the removals are my best attempt at not dropping a 100,000-line diff in a PR and asking for review. TaskChampion has sync support, but it will take a little work to hook that up to TaskWarrior, mostly in the form of adding new config parameters and plumbing those to TaskChampion.

Sync was pulled out not so much because it's hard, as because it was easy to separate. And, hopefully, a spot where someone else might be able to step in and take part.

I admit this was also a bit of a trial-balloon to see if _anyone_ was paying attention, so I'm glad to see you are! And also *wave* @dathanb. It's feeling a little lonely developing TW these days, and to be honest a lot of the limited feedback I do get is negative, which makes it hard to keep working on it. So I apologize that things have been slow.

ryneeverett:
Reviewability may be a good reason to make this a separate commit but I don't see the point of making it a separate pull request. Smaller pull requests are generally easier to review, but if they're broken down to the point that they don't stand on their own merits they become impossible to review.

Moreover, shouldn't `develop` be in a working state insofar as possible? I find it hard to justify deleting functionality with a promise to re-implement it before release.

My understanding is that your motivation here is to insert the roadmap into the taskwarrior code base as breadcrumbs for people to help you complete the taskchampion transition. I think a better strategy to improve this discoverability would be to break down the remaining TODO's in a single comprehensive roadmap (either an issue or a project).

The alternative to this PR would then be something like:

- [ ] Transition `sync` to taskchampion.
  - [X] Disable old `sync`. (https://github.com/GothenburgBitFactory/taskwarrior/pull/3092/commits/94cf41c969c0143a0a9ef47ea9929a500d0a2404)
  - [ ] Plumb taskchampion `sync`.

I'm eager to see the taskchampion project succeed. I think part of the roadblock right now is that it is hard to figure out how to help during the transition phase, but I don't see this PR as an effective strategy to overcome that.

djmitche:
There is actually a project for this -- https://github.com/orgs/GothenburgBitFactory/projects/2 -- but it garnered no attention and wasn't particularly useful to me, so I haven't been keeping it updated.

But, your point that this alone does not make a PR is a good one. I'll put up the whole switch-to-TC in a single PR.

Historically, and based on some discussions with @tbabej, there is not an expectation that the `develop` branch is feature-complete and ready to ship. We had agreed that getting something "mostly working" with a list of things to finish was a good way to centralize development, rather than working on a branch. So I'd still like to land that PR with some functionality not complete.

I think a good way to help would be to review that PR, as I don't know who else will.

Once that lands, you might work on adding the sync support. Basically, that's:
 - adding new config parameters for server type (local or remote) and origin, client_key, and encryption_secret
 - supporting a way to create a server using those parameters (already supported in taskchampion-lib, just needs support in `src/tc`) 
 - implementing tc::Replica::sync (currently commented out in `src/tc/Replica.h`)
 - calling that from `task sync`

PR #3060, opened 2023-02-04
[BREAKING] Remove journal.info support

djmitche:
This support will require access to all of the operations ever performed on a task, which is not currently exposed by TaskChampion (but see GothenburgBitFactory/taskchampion#373)

See https://github.com/GothenburgBitFactory/taskwarrior/discussions/3031.

#### Additional information...

- [x] I changed C++ code or build infrastructure.
  Please run the test suite and include the output of `cd test && ./problems`.

- [ ] I changed Rust code or build infrastructure.
  Please run `cargo test` and address any failures before submitting.

PR #3094, opened 2023-04-19
Use taskchampion

### LINKED ISSUE #2770: Replace Task & DB interface with C interface to TaskChampion

### LINKED ISSUE #3102: [using taskchampion] `task start` does not update modified date
```
⸩ task 27 | grep 'Last mod'
Last modified        2023-05-03 19:02:12 (2w)
⸩ task 27 start
Starting task 27 '...'.
Started 1 task.
You have more urgent tasks.
⸩ task 27 | grep 'Last mod'
Last modified        2023-05-03 19:02:12 (2w)
```

I suspect this is because Taskwarrior is directly setting task properties, rather than using the convenience functions.

djmitche:
#### Description

This PR switches Taskwarrior from using the TF2/TDB data storage (text files on disk) to using TaskChampion, as has been discussed for a long time. It's finally here!

This passes all of the tests, and is usable locally. I'm sure there are bugs I've not spotted, but we'll fix those as time allows. There are some breaking changes involved here:
 * `journal.info` support - this is the timetable of modifications to a task shown at the bottom of `task info`. Supporting this requires tracking operations on a per-task basis. This is a lot of work for fairly minimal functionality, so I'd like to defer it from this PR and either tackle it later or just drop the feature. In fact, it's already been removed in #3060.
 * The diff display and consent prompt in `task undo` is similar to `journal.info`, and will also rely on exposing TaskChampion's underlying "operations" to Taskwarrior and translating those to a table-style diff.
 * Sync support - all of the pieces are in place, and this just requires a bit of plumbing to make it work. I'd like to get this PR landed so that people can start looking for bugs in the local functionality while we work on adding the sync support.
   * Update `task sync` to use TC's sync mechanism, along with docs, build rules, and so on
   * Decide what a first release for taskchampion-sync-server looks like, and implement that. It's at "MVP" status right now: it runs, and you can sync against it, but it's got no configuration and will allow new clients without any extra authorization.
   * Update overall documentation for how to set this up.

 * `task purge` - this is no longer supported; tasks are automatically expired after they are deleted. Note that task expiration was added [here](https://github.com/taskchampion/taskchampion/pull/341).
 * the on-exit hook currently gets an empty list of changed tasks, regardless of what changed. We'll need to fix this before release.
 * Upgrade guide / documentation.

There are a number of TODO's left in the code, mostly for debugging functionality and as-yet unimplemented FFI glue between TaskChampion and Taskwarrior. These are by no means the first TODO's in the codebase, and will be addressed or removed in subsequent PRs.

The transition from TF2 to TaskChampion is via an `export` from one and `import` to the other.

The idea is that this will be the beginning of the Taskwarrior 3.0 release, so breaking changes are OK, and we have some time before that release to work out the details on the `develop` branch.

To review:
 * Review the diff below
 * Download, build, and import your current task DB and try it out
 * Consider what must be fixed in this PR, vs. what can be handled in followup issues before the 3.0 release. We can get those followups filed as issues, and track them to completion in a GitHub project.

(/cc @ryneeverett) (Fixes #2770)


djmitche:
Apparently my C++ isn't very good. I'll revise to fix the warnings :)

dathanb (approved):
Admittedly my C++ is weak and I'm not familiar with the nuances of Taskwarrior storage. But I read the whole PR, and it looks good to me.

djmitche:
Thank you @dathanb! The more eyes on this code, the better :D

djmitche:
I'm now using this at work (where I don't need `sync`). No issues so far -- it slotted right in where the old (TDB2-based) version was running, even integrating with bugwarrior!

djmitche:
Since there was some concern that it might be difficult, FYI I've added sync support in [my working branch](https://github.com/GothenburgBitFactory/taskwarrior/compare/develop...djmitche:taskwarrior:working-branch?expand=1).

src/commands/CmdPurge.cpp
```diff
@@ -49,131 +45,11 @@ CmdPurge::CmdPurge ()
...
-        purgeTask (task, count);
-    }
-  }
-
-  if (filtered.size () > 0 and ! matched_deleted)
-    Context::getContext ().footnote ("No deleted tasks specified. Maybe you forgot to delete tasks first?");
-
-  feedback_affected (count == 1 ? "Purged {1} task." : "Purged {1} tasks.", count);
-  return rc;
+  Context::getContext ().footnote ("As of version 3.0, this command has no effect.");
```
ryneeverett, line 51: I find this less clear than the way you put it in the PR description: "this is no longer supported; tasks are automatically expired after they are deleted." or the commit message: "TaskChampion removes deleted tasks automatically when they expire, so
this subcommand is no longer required."

Seems like it should be more clear (a) it's deprecated and (b) why it's now irrelevant.

test/undo.t
```diff
@@ -74,13 +74,15 @@ class TestUndoStyle(TestCase):
         self.t("add one project:foo priority:H")
         self.t("1 modify +tag project:bar priority:")
 
+    @unittest.expectedFailure  # undo diffs are not supported
```
ryneeverett, line 77: Would it be helpful to make it clear that this functionality is no longer supported by taskchampion? Also I'd never seen `expectedFailure` before and don't know why one would choose that over `skip`. Suggestion: `@unittest.skip('undo diffs not supported by taskchampion')`.

test/undo.t
```diff
@@ -74,13 +74,15 @@ class TestUndoStyle(TestCase):
...
     def test_undo_side_style(self):
         """Test that 'rc.undo.style:side' generates the right output"""
         self.t.config("undo.style", "side")
         code, out, err = self.t("undo", input="n\n")
         self.assertNotRegex(out, "-tags:\s*\n\+tags:\s+tag")
         self.assertRegex(out, "tags\s+tag\s*")
 
+    @unittest.expectedFailure  # undo diffs are not supported
```
ryneeverett, line 85: Would it be helpful to make it clear that this functionality is no longer supported by taskchampion? Also I'd never seen `expectedFailure` before and don't know why one would choose that over `skip`. Suggestion: `@unittest.skip('undo diffs not supported by taskchampion')`.

src/TDB2.cpp
```diff
@@ -780,8 +780,7 @@ void TDB2::revert ()
   std::string prior;
   revert_undo (u, uuid, when, current, prior);
 
-  // Display diff and confirm.
-  show_diff (current, prior, when);
```
ryneeverett, line 791: Should there be a TODO here? My impression is that that's how you're planning to keep track of functionality that needs to be re-implemented in the near term.

src/recur.cpp
```diff
@@ -90,7 +90,6 @@ void handleRecurrence ()
 
           Task rec (t);                          // Clone the parent.
           rec.setStatus (Task::pending);         // Change the status.
-          rec.id = Context::getContext ().tdb2.next_id ();      // New ID.
```
ryneeverett, line 93: Will recurring tasks now have the same id?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-  _U2I[task.get ("uuid")] = task.id;
-
-  _dirty = true;
+  replica = tc::Replica(location, create_if_missing);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-bool TF2::modify_task (const Task& task)
+// Add the new task to the replica.
+void TDB2::add (Task& task, bool/* = true */)
```
ryneeverett: Am I right in understanding that the second argument is now unnamed and left in to maintain the signature API? Should there be a TODO to remove this argument?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-      return true;
-    }
-  }
-
-  return false;
-}
+  // Ensure the task is consistent, and provide defaults if necessary.
+  // bool argument to validate() is "applyDefault". Pass add_to_backlog through
+  // in order to not apply defaults to synchronized tasks.
+  // This also ensures that the `uuid` attribute is set.
```
ryneeverett: Seems like this comment ought to be updated if the add_to_backlog argument is going to be dropped.

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-    Task original;
-    get (uuid, original);
-    Context::getContext ().hooks.onModify (original, task);
+  if (id.has_value ()) {
+    task.id = id.value();
   }
 
-  update (task, add_to_backlog);
+  // run hooks for this new task
+  Context::getContext ().hooks.onAdd (task);
```
ryneeverett, line 116: Comes down to add_to_backlog again -- is it no longer going to be necessary to exclude from running hooks on synced tasks?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-void TDB2::update (
-  Task& task,
-  const bool add_to_backlog,
-  const bool addition /* = false */)
+// Modify the task in storage to match the given task.
+//
+// This exhibits a bit of a race condition: if the stored task has changed since
+// it was loaded, this will revert those changes.  In practice, this is not an
+// issue.
+void TDB2::modify (Task& task, bool/* = true */)
```
ryneeverett: Am I right in understanding that the second argument is now unnamed and left in to maintain the signature API? Should there be a TODO to remove this argument?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-    if (task == original)
-      return;
-
-    if (add_to_backlog)
-    {
-      // All locally modified tasks are timestamped, implicitly overwriting any
-      // changes the user or hooks tried to apply to the "modified" attribute.
-      task.setAsNow ("modified");
+  get (uuid, original);
+	Context::getContext ().hooks.onModify (original, task);
```
ryneeverett, line 145: Comes down to add_to_backlog again -- is it no longer going to be necessary to exclude from running hooks on synced tasks?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
 ////////////////////////////////////////////////////////////////////////////////
-void TDB2::update (
-  Task& task,
-  const bool add_to_backlog,
-  const bool addition /* = false */)
+// Modify the task in storage to match the given task.
+//
+// This exhibits a bit of a race condition: if the stored task has changed since
+// it was loaded, this will revert those changes.  In practice, this is not an
+// issue.
```
ryneeverett: Could you elaborate on how this race condition would occur? What other mechanisms besides this method can modify the stored task?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-      // All locally modified tasks are timestamped, implicitly overwriting any
-      // changes the user or hooks tried to apply to the "modified" attribute.
-      task.setAsNow ("modified");
+  get (uuid, original);
+	Context::getContext ().hooks.onModify (original, task);
+
+  auto maybe_tctask = replica.get_task (uuid);
+  if (!maybe_tctask.has_value ()) {
+    throw std::string ("task no longer exists");
+  }
```
ryneeverett, line 150: How could the task not exist? Is it going to be possible for multiple processes to access the database? (Sorry if you already explained this to me in a previous PR review but I've forgotten the answer.)

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
-  // Restore signal handling.
-  signal (SIGHUP,    SIG_DFL);
-  signal (SIGINT,    SIG_DFL);
-  signal (SIGPIPE,   SIG_DFL);
-  signal (SIGTERM,   SIG_DFL);
-  signal (SIGUSR1,   SIG_DFL);
-  signal (SIGUSR2,   SIG_DFL);
-
-  Context::getContext ().time_commit_us += timer.total_us ();
+  // does nothing; changes are committed as they are made
```
ryneeverett: Should eliminating `TDB2::commit` be a TODO?

src/TDB2.cpp
```diff
@@ -42,967 +42,161 @@
...
+  }
+  return _working_set.value ();
 }
 
 ////////////////////////////////////////////////////////////////////////////////
 void TDB2::get_changes (std::vector <Task>& changes)
 {
-  changes = _changes;
+  // Modifications are not supported, therefore there are no changes
+  changes.clear();
```
ryneeverett, line 203: Should eliminating `TDB2::get_changes` be a TODO?

src/TDB2.cpp
```diff
@@ -1041,124 +235,128 @@ void TDB2::show_diff (
...
+  std::vector <Task> result;
+  for (size_t i = 0; i <= largest_index; i++) {
+    auto maybe_uuid = ws.by_index (i);
+    if (maybe_uuid.has_value ()) {
+      auto maybe_task = replica.get_task (maybe_uuid.value ());
+      if (maybe_task.has_value ()) {
+        result.push_back (Task (std::move (maybe_task.value ())));
+      }
+    }
+  }
```
ryneeverett, line 296: This seems like a relatively expensive operation compared to simply reading a file as we were doing before. Has any effort been made to minimize how often this gets called? Should there be a TODO for that?

src/TDB2.cpp
```diff
@@ -1041,124 +235,128 @@ void TDB2::show_diff (
...
+  auto all_tctasks = replica.all_tasks();
+  const tc::WorkingSet &ws = working_set ();
+
+  std::vector <Task> result;
+  for (auto& tctask : all_tctasks) {
+    // if this task is _not_ in the working set, return it.
+    if (!ws.by_uuid (tctask.get_uuid ())) {
+      result.push_back (Task (std::move (tctask)));
+    }
+  }
```
ryneeverett, line 315: This seems like a relatively expensive operation compared to simply reading a file as we were doing before. Has any effort been made to minimize how often this gets called? Should there be a TODO for that?

ryneeverett (commented):
I'm pleased with the scope and direction of this PR, thanks!

Note: I haven't tested it out yet and I'm not competent to give more than a superficial review of C++. Feel free to ignore my duplicate comments if the answers are the same.

Question: Do you see the current TDB2 implementation as a long-term solution or is it more of a scaffolding to be gradually phased out as taskchampion takes more responsibility?

tbabej:
I'm going to test drive this branch locally as well.

djmitche:
Awesome! I set up an hourly `task export` as a backstop against any disasters, but so far it's working fine for me.

I have `task sync` mostly put together locally. There are some remaining TODO's on that which I may find time to work on this weekend. I'll also open some issues to make the sync server easier to use and deploy - it's pretty bare-bones right now (like, it should probably have a config file, and maybe some access-control options).

I'll also work through the review comments here this weekend.

test/undo.t
djmitche: I expect that we'll re-introduce this functionality. The advantage of `expectedFailure` is that it fails if the test succeeds. So when someone does re-introduce this behavior, the test failures will alert them that there are tests they should re-enable.

src/TDB2.cpp
djmitche: I plan to file issues. There are 121 TODO's in the codebase already (according to `git grep TODO | wc -l`), so they don't make a great way to track work :(

src/recur.cpp
djmitche: No, the id's are set based on the task's position in the working set, which comes from TaskChampion. Next time this task is read from the DB it will have its correct ID.

src/TDB2.cpp
djmitche: Good spot - there's no need to maintain the signature, so I'll remove the argument.

src/TDB2.cpp
djmitche: Right, sync'd tasks don't enter the task DB through `TDB2::add`, so there's no need to exclude them.

src/TDB2.cpp
djmitche: It's mostly defensive programming. The DB uses sequential transactions, but the Taskwarrior code itself is not executed within that transaction. So it's possible for a task to be modified by some other process (such as if the user is running multiple `task` processes at the same time) between being read and the call to TDB2::modify. In practice, even if that occurs the task is unlikely to no longer exist, since we don't delete tasks until they expire.

src/TDB2.cpp
djmitche: Yep, thanks!

src/TDB2.cpp
djmitche: Yes, I'll remove it.

src/TDB2.cpp
djmitche: Correct.

src/TDB2.cpp
djmitche: See my comment on "How could this task not exist".

src/TDB2.cpp
djmitche: Indeed, I can just remove it.

src/TDB2.cpp
djmitche: Hm, this is used in hooks, and I didn't notice that. We'll need to bring this back. I'll add it to the list in the PR description (and update the comment here).

src/TDB2.cpp
djmitche: This worried me a little bit, too -- this function is basically loading every pending task from the DB, using a distinct transaction for each one. In general that's not how you're supposed to do things with a DB!

We can definitely optimize things later as we identify performance issues.

src/TDB2.cpp
djmitche: This one is even more worrisome, since it's loading _all_ tasks! But, it's less frequently used.

djmitche:
(branch updated to rebase on `main` and to adopt changes suggested by @ryneeverett)

src/TDB2.cpp
ryneeverett: Thanks. Maybe you could incorporate the more concrete application in that comment into this code comment? Something like:

> This exhibits a bit of a race condition: if the stored task has changed since it was loaded, for example by two simultaneous `task` invocations, this will revert those changes. In practice, this is not an issue because we don't delete tasks until they expire.

djmitche: I'll add to the comment. Note that there are two different race conditions here, although they are related.
 * Multiple changes are made simultaneously.
 * A task being deleted simultaneously with a modification.

In the first case, whichever change commits last, wins. In the second case, the deletion wins.

djmitche: Well, the deletion _could_ win, but instead we throw an error.

src/commands/CmdPurge.cpp
ryneeverett: I believe this is the only comment you haven't addressed. A more helpful error message would be nice. (Github's latest UI change seems to have removed the "Resolved" button which makes it painful to keep track.)

djmitche: I added another line. This is something we'll need to document, too. I'm not sure if/why people were using `task purge`, but we'll need to suggest alternatives for whatever their use-case was.

Huh, I do see the resolved buttons. I'll click them :)

djmitche:
@tbabej how is this working for you? What do you think about landing?

I did encounter #3102 just now, but that's not going to be too hard to fix.

djmitche:
fa5d448ef6943f4dafa7f82851cd0eb8f60f647f fixes #3102.

djmitche:
9bf2c2e4c08e2d4ac2ea17f092127543e8911009 fixes #3109.

djmitche:
I think further progress on this overall project is blocked on landing this PR. Once that's done, we can get a bunch of issues filed, organized in a project, and start gathering more as reports come in.  It's getting a little awkward filing issues against this PR. Is there anything that would _prevent_ this PR from landing?

The one wrinkle is, we're currently planning to ask users to `task export` on 2.6 and then `task import` on 3.0. But, `task export` omits the task history (which is part of what led me to think it was low-priority). So we might need to release a new 2.6 that can export both tasks and their history.  But, I think we could do that on a release branch. Also, design for that functionality is still not established, so I wouldn't want to block landing this PR for the time it takes to finish that.

djmitche:
In an effort to unstick this:

**I will merge this PR tomorrow, Sunday, in the absence of a reason or request not to**

(of course, resets and reverts are always an option)

Issue #2870, opened 2022-07-10
Support a way to get `taskchampion.h` in a build script


djmitche:
Currently, to link against the taskchampion-lib C functions, you need to copy `taskchampion.h` out of the source directory (or, in Taskwarrior's case, put that directory in the include path).  That's going to lead to sadness when the version of taskchampion-lib and of taskchampion.h inevitably fall out of sync.

Let's provide a function or constant on taskchampion-lib that contains the header file, so that consumers of the library can put the header file wherever they want using a simple build script (`build.rs`).

djmitche:
ffizz might help with this

djmitche:
Working on this in my [use-ffizz-header branch](https://github.com/djmitche/taskwarrior/pull/new/use-ffizz-header).

djmitche:
Fixed in #3114.

PR #3114, opened 2023-06-11
Use ffizz-header

djmitche:
#### Description

This switches the generation of `taskchampion.h` to use `ffizz-header` (which I wrote). This drops use of `cbindgen`, which did not allow sufficient control over the generated header.


.github/workflows/publish-docs.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          # if this changes, change it in cli/Cargo.toml and .github/workflows/publish-docs.yml as well
+          # if this changes, change it in .github/workflows/publish-docs.yml as well
```
ryneeverett: This comment created an infinite loop in my brain. Should the referenced file be changed to `
.github/workflows/checks.yml`?

djmitche: Thanks - fixed!

djmitche:
This also fixes #2870, by adding a `taskwarrior_lib::generate_header` function that returns the header. Embedders can use this in their build process.

djmitche:
I'm merging this on the assumption that (a) someone has looked at it (thanks @ryneeverett!) and (b) it has no user-visible impact.

PR #3130, opened 2023-07-08
Change "client key" to "client id"

djmitche:
In #3118 @ryneeverett mentioned that "key" suggests that this is a secret, when in truth it's just a user identifier. So "ID" is a better word for it than "key".

This would be a breaking change (including making a breaking change to the server's DB schema), except we have never released anything with `client_key`.


ryneeverett (approved):
Thanks!

Issue #3131, opened 2023-07-08
on-exit hook called many times when using tab completion


sdondley:
* What command(s) did you run?

task pr&lt;tab>

* What did you expect to happen?

tab completion to happen quickly

* What actually happened?

tab completion happened slowly

AFAICT, an `on-exit.sh` script I have is run several times after hitting the <tab> key. 

Here's my `on-exit.sh` script I'm using to troubleshoot this problem:

```
if [ $# -gt 1 ]; then
    echo $2 >&2
    task rc.verbose=nothing rc.hooks=0 sync
    cd /root/.config/ext_volume/orgs
    git add  . > /dev/null
    git commit -m "committing tasks" > /dev/null
fi
exit 0;

```

When I hit tab during completion, I see something like this on the command line:

```
task pr<tab>args:task _commands
args:task _aliases
args:task _zshcommands
args:task _aliases
args:task _projects
args:task _zshids
args:task _udas
args:task _get rc.uda.priority.label
args:task _get rc.uda.reviewed.label
args:task _tags
args:task _config
```

Either this is the bug or I'm using the hook wrong. But I didn't see any documentation that explains how to use it properly.

sdondley:
As a workaround, I add the following code to the top of my on-exit script

```
if [[ $command == "args:task _"* ]]; then
    exit 0
fi
```

PR #3133, opened 2023-07-14
prevent task completion commands from triggering hooks #3131

sdondley:
#### Description

Fixed bug mentioned at #3131. tab completion was triggering hooks. `rc.hooks=0` added to suppress the hooks calls for zsh

### DIFF
--- a/scripts/zsh/_task
+++ b/scripts/zsh/_task
@@ -31,7 +31,7 @@ _task_filter() {
   local word=$'[^\0]#\0'
 
   # projects
-  local _task_projects=($(task _projects))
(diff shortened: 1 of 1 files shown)


djmitche (commented):
Could you add a bit of abstraction here, maybe
```
local task='task rc.hooks=0'
```
and later use that as `$($task _tags)` and so on?

sdondley:
That's an interesting suggestion and topic of conversation. Seems like it would be a good idea as the command grows more complex. But for this simple fix, it might seem like overkill. 

Not sure myself. But my knee jerk reaction is to think that adding another layer of abstraction is overkill.

Issue #3142, opened 2023-07-16
Automate synchronization of MSRV


djmitche:
We've already got a number of places we specify the minimum supported Rust version (MSRV), and will probably add a few more.

Let's add a function to `cargo xtask` to update the MSRV everywhere it appears, and run that in CI with a `diff` to ensure it's up-to-date. We already do something similar to generate and diff `taskwarrior.h`, so follow that pattern.

iwyatt:
Hi Dustin! I saw you are working on integrating TaskChampion into TaskWarrior and its also developed in Rust! I'm interested in all those things and as I'm learning Rust, might want to lend a hand if able (I'm not a professional programmer but eager to cut my teeth on "intern-level" type work).

I hope to replace my current task manager someday and I see a lot of potential in the TaskWarrior/Champion ecosystem. 

I wouldn't want to be a burden to anyone (I realize I might need some guidance), but am curious if you think this would be a good issue for me to lend a hand?

ryneeverett:
FWIW I also took a look at this and determined it's not a sufficiently entry-level task for me. Seems like either it's mislabeled (as "good first issue") or more elaboration is needed.

djmitche:
Sure, and I'm happy to provide guidance :)

There's a nice readme at [cargo-xtask](https://github.com/matklad/cargo-xtask) that describes how this fits together. Briefly:
 * The `cargo` command looks in [`.cargo/config`](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/.cargo/config) for aliases for subcommands. In this repo, `cargo xtask ...` is translated to `cargo run --package xtask -- ...`
 * This runs the crate in this repo called `xtask` (https://github.com/GothenburgBitFactory/taskwarrior/tree/develop/taskchampion/xtask).
 * The `xtask` crate just has a `main.rs` which is a regular command-line program. It's very short and hopefully pretty easy to understand. Right now, it has a single subcommand `codegen` (so, `cargo xtask codegen`) which results in a call to the `codegen` method.

The goal for this issue would be to add code to that `main.rs` to support another subcommand, `cargo xtask msrv X.Y` which updates all of the places in the repo where the MSRV occurs (`git grep MSRV` will show you where those are) to contain `X.Y`. This will require
 * A little bit more command-line parsing. What exists right now is done by hand, but `clap` is a popular CLI crate that could be used for this purpose.
[...]

PR #3189, opened 2023-10-10
Add xtask msrv

iwyatt:
#### Description

Adds a placeholder function for xtask called msrv with arguments (x,y) that starts work on [taskwarrior issue 3142](https://github.com/GothenburgBitFactory/taskwarrior/issues/3142): Automate synchronization of MSRV

- [x] I added placeholder function with 'hello world' and x,y parameter output.

- [x] I ran `cargo run msrv arg1 arg2` to confirm function is called and outputs as expected

- [x] I ran cargo test to confirm no warnings or errors

### DIFF
--- a/taskchampion/xtask/src/main.rs
+++ b/taskchampion/xtask/src/main.rs
@@ -3,11 +3,6 @@
 //! At the moment it is very simple, but if this grows more subcommands then
 //! it will be sensible to use `clap` or another similar library.
 
-//! The const MSRV_FILE_PATHS_REGEX in /xtask/main.rs is an array of tuples
-//!  of the form (PATH, REGEX) where PATH and REGEX are literals where
-//!  PATH is a file that updates the Minimum Supported Rust Version
-//!  and REGEX is the pattern to find the appropriate line in the file
-
 use regex::Regex;
 use std::env;
 use std::fs::File;
@@ -15,29 +10,34 @@ use std::io::{BufRead, BufReader};
 use std::io::{Seek, Write};
 use std::path::{Path, PathBuf};
 
+/// The const MSRV_PATHS_REGEX in /xtask/main.rs is an array of tuples
+///  of the form (PATH, REGEX) where PATH and REGEX are literals where
+///  PATH is a file that updates the Minimum Supported Rust Version
+///  and REGEX is the pattern to find the appropriate line in the file
+
+// MSRV - "2.0"
+
 // Increment length of array when adding tuples of (PATH, REGEX).
 const MSRV_PATH_REGEX: [(&str, &str); 2] = [
+    ("xtask/src/main.rs", r#"- "[0-9]+("|\.|[0-9])+"#),
     (
-        "../../.github/workflows/rust-tests.yml",
-        r#"- "[0-9]+("|\.|[0-9])+"#,
-    ),
-    (
-        "../../.github/workflows/checks.yml",
+        ".github/workflows/checks.yml",
         r#"toolchain: "[0-9]+("|\.|[0-9])+"#,
     ),
 ];
 
 pub fn main() -> anyhow::Result<()> {
+    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let workspace_dir = manifest_dir.parent().unwrap();
     let arguments: Vec<String> = env::args().collect();
+
     if arguments.len() < 2 {
         anyhow::bail!("xtask: Valid arguments are: `codegen`, `msrv <version x.y>`");
     }
 
-    println!("{:#?}", env::current_dir());
-
     match arguments[1].as_str() {
-        "codegen" => codegen(),
-        "msrv" => msrv(arguments),
+        "codegen" => codegen(workspace_dir),
+        "msrv" => msrv(arguments, workspace_dir),
         _ => anyhow::bail!("xtask: unknown xtask"),
     }
 }
@@ -45,9 +45,7 @@ pub fn main() -> anyhow::Result<()> {
 /// `cargo xtask codegen`
 ///
 /// This uses ffizz-header to generate `lib/taskchampion.h`.
-fn codegen() -> anyhow::Result<()> {
-    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
-    let workspace_dir = manifest_dir.parent().unwrap();
+fn codegen(workspace_dir: &Path) -> anyhow::Result<()> {
     let lib_crate_dir = workspace_dir.join("lib");
     let mut file = File::create(lib_crate_dir.join("taskchampion.h")).unwrap();
     write!(&mut file, "{}", ::taskchampion_lib::generate_header()).unwrap();
@@ -59,104 +57,77 @@ fn codegen() -> anyhow::Result<()> {
 ///
 /// This checks and updates the Minimum Supported Rust Version for all files specified in MSRV_PATH_REGEX`.
 /// Each line where the regex matches will have all values of the form `#.##` replaced with the given MSRV.
-fn msrv(args: Vec<String>) -> anyhow::Result<()> {
-    if args.len() != 2 {
-        // check that (X.Y) argument is (mostly) valid:
-        if !args[2].chars().all(|c| c.is_numeric() || c == '.') {
-            anyhow::bail!("xtask: Invalid argument format. Xtask MSRV argument takes the form \"X.Y(y)\", where XYy are numbers. eg: `cargo run xtask MSRV 1.68`");
-        }
+fn msrv(args: Vec<String>, workspace_dir: &Path) -> anyhow::Result<()> {
+    // check that (X.Y) argument is (mostly) valid:
+    if !args[2].chars().all(|c| c.is_numeric() || c == '.') {
+        anyhow::bail!("xtask: Invalid argument format. Xtask MSRV argument takes the form \"X.Y(y)\", where XYy are numbers. eg: `cargo run xtask MSRV 1.68`");
+    }
 
-        // set regex for replacing version number only within the pattern found within a line
-        let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+").unwrap();
+    // set regex for replacing version number only within the pattern found within a line
+    let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+").unwrap();
 
-        // for each file in const paths vector
-        for msrv_file in MSRV_PATH_REGEX {
-            let path: &Path = Path::new(&msrv_file.0);
-            let mut is_pattern_in_file = false;
+    // for each file in const paths tuple
+    for msrv_file in MSRV_PATH_REGEX {
+        let mut is_pattern_in_file = false;
 
-            // check if path exists, if not, skip then continue
-            if !path.exists() {
-                println!(
-                    "xtask: Warning: xtask: Path \"{}\" not found. Skipping.",
-                    path.display()
-                );
-                continue;
-            }
+        let file_path = workspace_dir.join(msrv_file.0);
+        let path: &Path = Path::new(&file_path);
+        if !&path.exists() {
+            anyhow::bail!("xtask: path does not exist {}", &path.display());
+        };
 
-            let mut file: File = File::options().read(true).write(true).open(path)?;
-            let reader = BufReader::new(&file);
+        let mut file: File = File::options().read(true).write(true).open(path)?;
+        let reader = BufReader::new(&file);
 
-            // set search string and the replacement string for version number content
-            let re_msrv_pattern = Regex::new(msrv_file.1).unwrap();
-            let version_replacement_string = &args[2];
+        // set search string and the replacement string for version number content
+        let re_msrv_pattern = Regex::new(msrv_file.1).unwrap();
+        let version_replacement_string = &args[2];
 
-            // for each line in file
-            let mut file_string = String::new();
-            for line in reader.lines() {
-                let line_ref = &line.as_ref().unwrap();
+        // for each line in file
+        let mut file_string = String::new();
+        for line in reader.lines() {
+            let line_ref = &line.unwrap();
 
-                // if rust version pattern is found and is different, update it
-                if let Some(pattern_offset) = re_msrv_pattern.find(line_ref) {
-                    if pattern_offset.as_str() != version_replacement_string {
-                        println!(
-                            "{}",
-                            format!(
-                                "{}\n",
-                                re_msrv_version
-                                    .replace(line_ref, version_replacement_string)
-                                    .to_string()
-                                    .as_str()
-                            )
+            // if rust version pattern is found and is different, update it
+            if let Some(pattern_offset) = re_msrv_pattern.find(line_ref) {
+                if pattern_offset.as_str() != version_replacement_string {
+                    file_string += format!(
+                        "{}\n",
+                        re_msrv_version
+                            .replace(line_ref, version_replacement_string)
+                            .to_string()
                             .as_str()
-                        );
-
-                        file_string += format!(
-                            "{}\n",
-                            re_msrv_version
-                                .replace(line_ref, version_replacement_string)
-                                .to_string()
-                                .as_str()
-                        )
-                        .as_str();
-
-                        is_pattern_in_file = true;
-                        println!(
-                            "xtask: MSRV pattern found in {:#?}; updating MSRV to {}",
-                            path.as_os_str(),
-                            args[2]
-                        );
-
-                        continue;
-                    } else {
-                        println!("xtask: MSRV pattern found in {:#?} and is same as specified version in xtask argument.", path.as_os_str());
-                    }
+                    )
+                    .as_str();  
+                    is_pattern_in_file = true;
+                    continue;
                 }
-
-                file_string += line_ref;
-                file_string += "\n";
             }
 
-            // if pattern was found and updated, write to disk
-            if is_pattern_in_file {
-                // Prepare to write the file to disk
-
-                //  Set the file length to the file_string length
-                let _ = file.set_len(file_string.len() as u64);
-
-                //  set the cursor to the beginning of the file and write
-                let _ = file.seek(std::io::SeekFrom::Start(0));
-                let file_write_result = file.write(file_string.as_bytes());
-
-                // if error, print error messege and exit
-                if file_write_result.is_err() {
-                    anyhow::bail!("xtask: unable to write file to disk: {}", &path.display());
-                }
-            } else {
-                println!("xtask: No changes to file: \"{}\"", &path.display());
-            }
+            file_string += line_ref;
+            file_string += "\n";
         }
-    } else {
-        anyhow::bail!("xtask: Wrong number of arguments.");
+
+        // if pattern was found and updated, write to disk
+        if is_pattern_in_file {
+            //  Set the file length to the file_string length
+            let _ = file.set_len(file_string.len() as u64);
+
+            //  set the cursor to the beginning of the file and write
+            let _ = file.seek(std::io::SeekFrom::Start(0));
+            let file_write_result = file.write(file_string.as_bytes());
+
+            // if error, print error messege and exit
+            if file_write_result.is_err() {
+                anyhow::bail!("xtask: unable to write file to disk: {}", &path.display());
+            }
+
+            // notify user this file was updated
+            println!(
+                "xtask: Updated MSRV in {}",
+                re_msrv_version.replace(msrv_file.0, version_replacement_string)
+            );
+        }
     }
 
     Ok(())
--- a/taskchampion/xtask/src/main.rs
+++ b/taskchampion/xtask/src/main.rs
@@ -6,20 +6,23 @@
 use regex::Regex;
 use std::env;
 use std::fs::File;
-use std::io::{BufRead, BufReader};
-use std::io::{Seek, Write};
+use std::io::{BufRead, BufReader, Seek, Write};
 use std::path::{Path, PathBuf};
 
-/// The const MSRV_PATHS_REGEX in /xtask/main.rs is an array of tuples
-///  of the form (PATH, REGEX) where PATH and REGEX are literals where
-///  PATH is a file that updates the Minimum Supported Rust Version
-///  and REGEX is the pattern to find the appropriate line in the file
-
-// Increment length of array when adding tuples of (PATH, REGEX).
-const MSRV_PATH_REGEX: [(&str, &str); 1] = [(
-    "xtask/Cargo.toml",
-    r#"rust-version = "[0-9]+("|\.|[0-9])+""#,
-)];
+/// Tuples of the form (PATH, REGEX) where PATH and REGEX are literals where PATH is a file that
+/// conains the Minimum Supported Rust Version and REGEX is the pattern to find the appropriate
+/// line in the file. PATH is relative to the `taskchampion/` directory in the repo.
+const MSRV_PATH_REGEX: &[(&str, &str)] = &[
+    (
+        "../.github/workflows/checks.yml",
+        r#"toolchain: "[0-9.]+*" # MSRV"#,
+    ),
+    ("../.github/workflows/rust-tests.yml", r#""[0-9.]+" # MSRV"#),
+    (
+        "taskchampion/src/lib.rs",
+        r#"Rust version [0-9.]* and higher"#,
+    ),
+];
 
 pub fn main() -> anyhow::Result<()> {
     let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
@@ -54,9 +57,10 @@ fn codegen(workspace_dir: &Path) -> anyhow::Result<()> {
 /// Each line where the regex matches will have all values of the form `#.##` replaced with the given MSRV.
 fn msrv(args: Vec<String>, workspace_dir: &Path) -> anyhow::Result<()> {
     // check that (X.Y) argument is (mostly) valid:
-    if !args[2].chars().all(|c| c.is_numeric() || c == '.') {
+    if args.len() < 3 || !args[2].chars().all(|c| c.is_numeric() || c == '.') {
         anyhow::bail!("xtask: Invalid argument format. Xtask MSRV argument takes the form \"X.Y(y)\", where XYy are numbers. eg: `cargo run xtask MSRV 1.68`");
     }
+    let version_replacement_string = &args[2];
 
     // set regex for replacing version number only within the pattern found within a line
     let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+")?;
@@ -65,28 +69,27 @@ fn msrv(args: Vec<String>, workspace_dir: &Path) -> anyhow::Result<()> {
     for msrv_file in MSRV_PATH_REGEX {
         let mut is_pattern_in_file = false;
 
-        let file_path = workspace_dir.join(msrv_file.0);
-        let path: &Path = Path::new(&file_path);
-        if !&path.exists() {
+        let path = workspace_dir.join(msrv_file.0);
+        let path = Path::new(&path);
+        if !path.exists() {
             anyhow::bail!("xtask: path does not exist {}", &path.display());
         };
 
-        let mut file: File = File::options().read(true).write(true).open(path)?;
+        let mut file = File::options().read(true).write(true).open(path)?;
         let reader = BufReader::new(&file);
 
         // set search string and the replacement string for version number content
         let re_msrv_pattern = Regex::new(msrv_file.1)?;
-        let version_replacement_string = &args[2];
 
         // for each line in file
         let mut file_string = String::new();
         for line in reader.lines() {
-            let line_ref = &line?;
+            let line = &line?;
 
             // if rust version pattern is found and is different, update it
-            if let Some(pattern_offset) = re_msrv_pattern.find(line_ref) {
+            if let Some(pattern_offset) = re_msrv_pattern.find(line) {
                 if !pattern_offset.as_str().contains(version_replacement_string) {
-                    file_string += &re_msrv_version.replace(line_ref, version_replacement_string);
+                    file_string += &re_msrv_version.replace(line, version_replacement_string);
 
                     file_string += "\n";
 
@@ -95,23 +98,18 @@ fn msrv(args: Vec<String>, workspace_dir: &Path) -> anyhow::Result<()> {
                 }
             }
 
-            file_string += line_ref;
+            file_string += line;
             file_string += "\n";
         }
 
         // if pattern was found and updated, write to disk
         if is_pattern_in_file {
             //  Set the file length to the file_string length
-            let _ = file.set_len(file_string.len() as u64);
+            file.set_len(file_string.len() as u64)?;
 
             //  set the cursor to the beginning of the file and write
-            let _ = file.seek(std::io::SeekFrom::Start(0));
-            let file_write_result = file.write(file_string.as_bytes());
-
-            // if error, print error messege and exit
-            if file_write_result.is_err() {
-                anyhow::bail!("xtask: unable to write file to disk: {}", &path.display());
-            }
+            file.seek(std::io::SeekFrom::Start(0))?;
+            file.write_all(file_string.as_bytes())?;
 
             // notify user this file was updated
             println!(
--- a/taskchampion/xtask/src/main.rs
+++ b/taskchampion/xtask/src/main.rs
@@ -15,19 +15,14 @@ use std::path::{Path, PathBuf};
 ///  PATH is a file that updates the Minimum Supported Rust Version
 ///  and REGEX is the pattern to find the appropriate line in the file
 
-// MSRV - "2.0"
-
 // Increment length of array when adding tuples of (PATH, REGEX).
-const MSRV_PATH_REGEX: [(&str, &str); 2] = [
-    ("xtask/src/main.rs", r#"- "[0-9]+("|\.|[0-9])+"#),
-    (
-        ".github/workflows/checks.yml",
-        r#"toolchain: "[0-9]+("|\.|[0-9])+"#,
-    ),
-];
+const MSRV_PATH_REGEX: [(&str, &str); 1] = [(
+    "xtask/Cargo.toml",
+    r#"rust-version = "[0-9]+("|\.|[0-9])+""#,
+)];
 
 pub fn main() -> anyhow::Result<()> {
-    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
     let workspace_dir = manifest_dir.parent().unwrap();
     let arguments: Vec<String> = env::args().collect();
 
@@ -64,7 +59,7 @@ fn msrv(args: Vec<String>, workspace_dir: &Path) -> anyhow::Result<()> {
     }
 
     // set regex for replacing version number only within the pattern found within a line
-    let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+").unwrap();
+    let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+")?;
 
     // for each file in const paths tuple
     for msrv_file in MSRV_PATH_REGEX {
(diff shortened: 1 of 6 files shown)


taskchampion/xtask/src/main.rs
```diff
@@ -9,10 +9,16 @@ use std::io::Write;
...
+    if arguments.len() < 2 {
+        println!("Valid arguments are `codegen`, `msrv <arg1> <arg2>`");
+        return Ok(());
+    }
+
+    match arguments[1].as_str() {
+        "codegen" => codegen(),
+        "msrv" => msrv(arguments),
+        arg => anyhow::bail!("unknown xtask {}", arg),
         _ => anyhow::bail!("unknown xtask"),
```
djmitche: This `_` was to catch the `None` variant (`arg.as_deref()` is an `Option<..>`). Since there's no such variant anymore, the first three arms of this `match` capture all of the possibilities and this last one is not required.

djmitche (commented):
This looks good so far! Let's continue working on this branch.

I think the next step will be to take the new MSRV given on the command-line and substitute it into the necessary files. I think the trick will be to identify the precise spot in those files to write the MSRV, without accidentally changing anything else. It's fine to add some comments to the files that the `cargo xtask msrv` command can use to find it. For example, in the `.github/workflow` files, that could be

```yaml
        # MSRV - update with `cargo xtask msrv`
        - "1.64"
```

It's probably easiest to do this by reading each file into memory with [`std::fs::read_to_string`](https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html), then manipulating it with regular expressions (using the `regex` crate). But, you could also do this by iterating over lines in the file and using some basic string operations like [`s.ends_with("# MSRV")`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.ends_with).

Give that a try, and let me know if you get stuck!

taskchampion/xtask/src/main.rs
iwyatt: I changed my code a bit and now I believe these paths are reachable. Interestingly, the last two arms are helpful for providing feedback for two different types of failures. I left them in there for the time being but if the text output isn't useful, then I have no problem removing them.

iwyatt:
Hi @djmitche  - happy Friday!

Ok, so based on your feedback and some of my own assumptions and updated the code accordingly:
- implemented upsert functionality which looks for a known pattern in a comment block (if the MSRV comment block isn't present, it will be added)
- added partial validation check for the version X.Y argument (noted in my testing that emojis can really mess up the rest of document)
- added checks for valid paths, successful write to disk, and placing comment block AFTER library-doc comments `//!` as req'd by cargo
- left lots of comments that explain my thinking in attempting to keep formatting nice

I see that the CICD test `rust / rust stable on windows-latest (pull_request)` has failed, which appears to be due to `use std::os::unix::prelude::FileExt;` not being available on windows platforms. I think I see what is going on, which will require changing some of my code to not depend on the write_at() function, so will fix in the next pull request. 

Q: Is there a way to run all the CICD tests locally before push my pull request?

djmitche:
> Q: Is there a way to run all the CICD tests locally before push my pull request?

Other than on a platform you have locally, I don't think so. It _may_ be possible to run a codespace in other platforms (like Windows), but I don't have any experience doing so. Still, if tests run locally then it's usually fine to use the PR for testing other platforms.

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
 }
+
+/// `cargo xtask msrv (X.Y)`
+///
+/// This updates all of the places in the repo where the MSRV occurs to <arg2> for the list of file paths defined as const in Xtask main.rs.
+/// If no pre-existing Xtask-formatted MSRV is found, inserts the MSRV as first comment block after the last `//!` library-doc comments.
+fn msrv(args: Vec<String>) -> anyhow::Result<()> {
+    match args.len(){
+        0 ..= 2 => println!("Error: xtask: Command `caro run xtask msrv` expects at least one argument, none detected."),
+        3 => {
```
djmitche: It may help indentation to just use `if` here (`if args.len != 3 { ..error; return; }`)

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
+
+                // check that (X.Y) argument is (mostly) valid:
+                for xy in & mut args[2].chars() {
```
djmitche: This is fine, but somewhat more idiomatic to use an iterator expression: `if !args[2].chars().all(|c| c.is_numeric() || c == '.') { .. }`

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
+
+                        // if a pre-existing MSRV pattern is found, update it.
+                        if pattern_offset.is_some() {
+                            msrv_pattern_found = true;
+                            file_string += & replacement_string;
+                            println!("xtask: MSRV pattern found in {:#?}; updating MSRV to {}", path.as_os_str(), args[2]);
+                            continue;
+                        }
+
+                        file_string += & line.as_ref().unwrap();
```
djmitche: It looks like this also needs a `rustfmt` - there should be no space between `&` and the thing it references.

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
     Ok(())
 }
+
+/// `cargo xtask msrv (X.Y)`
+///
+/// This updates all of the places in the repo where the MSRV occurs to <arg2> for the list of file paths defined as const in Xtask main.rs.
+/// If no pre-existing Xtask-formatted MSRV is found, inserts the MSRV as first comment block after the last `//!` library-doc comments.
+fn msrv(args: Vec<String>) -> anyhow::Result<()> {
+    match args.len(){
+        0 ..= 2 => println!("Error: xtask: Command `caro run xtask msrv` expects at least one argument, none detected."),
```
djmitche: This should probably be `anyhow::bail!`

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
+                    // set pattern to search for and the replacement string
+                    let msrv_pattern = "///  MSRV = \"";
+                    let replacement_string = format!("///  MSRV = \"{}\"\n", args[2]);
+
+                    // for each line in file
+                    let mut file_string = String::new();
+                    let mut msrv_pattern_found = false; //should change this to an int and update with the line# when found
+                    let mut comment_pattern_last_offset : usize = 0;
+                    for line in reader.lines() {
+                        let pattern_offset = & line.as_ref().unwrap().find(msrv_pattern);
```
djmitche: A more common idiom here, with line of type `Result<String, Error>`, is to use `?`.  If `line` was just used once, that might be:
```rust
                        let pattern_offset = line?.find(msrv_pattern);
```
but in this case `line` is used a few more times, so just check for errors once:
```rust
                        let line = line?;
                        let pattern_offset = line.find(msrv_pattern);
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,17 +3,33 @@
...
 use std::env;
 use std::fs::File;
 use std::io::Write;
-use std::path::PathBuf;
+use std::io::{BufRead, BufReader};
+use std::os::unix::prelude::FileExt;
+use std::path::{Path, PathBuf};
+
+// Increment length of array when adding more paths.
+const MSRV_FILE_PATHS: [&str; 2] = [r"main.rs", r"src/main.rs"];
```
djmitche: I can't quite tell where these paths are relative to?

Cargo sets `$CARGO_MANIFEST_DIR` to the `taskchampion/xtask` directory, so you might be able to build paths relative to there, and then use that environment variable to make `xtask msrv` independent of the current directory.

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
+
+                        // if the line starts with a library-doc comments (which must precede regular comments), update index
+                        // used iff MSRV isn't found
+                        if line.as_ref().unwrap().starts_with("//!") {
+                            comment_pattern_last_offset = file_string.len();
+                        }
+
+                    }
+
+                    // insert MSRV block if no pre-existing block is found
```
djmitche: This is probably unnecessary -- If we add a new file to MSRV_FILE_PATHS, we can add the comment at the same time.

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +45,99 @@ fn codegen() -> anyhow::Result<()> {
...
+                        }
+
+                        file_string.insert_str(comment_pattern_last_offset, & insert_string);
+                    }
+
+                    // write the updated file to disk
+                    let file_write_result = file.write_at(file_string.as_bytes(),0);
+
+                    // if error, print error messege and exit
+                    if file_write_result.is_err() {
```
djmitche: Could just use `?` here, too.

djmitche (commented):
Looking good so far! Hopefully these comments help with next steps.

Worth considering, too: `///` is not a comment in YAML, but `#` is. Can you abstract that away? Or maybe have different functions for `.rs` and .`yml` files? We'll also need to modify a `.toml` file (`Cargo.toml`) when the rust version is added there.

iwyatt:
Thank you for taking the time to provide thorough and articulate comments! Will get started on these changes this upcoming week.

taskchampion/xtask/src/main.rs
iwyatt: Oh, I'm sorry. These are left over test paths. One fails, one succeeds.

taskchampion/xtask/src/main.rs
iwyatt: Done

taskchampion/xtask/src/main.rs
iwyatt: Done.

taskchampion/xtask/src/main.rs
iwyatt: Done.

taskchampion/xtask/src/main.rs
iwyatt: Done.

taskchampion/xtask/src/main.rs
iwyatt: Done. Ah. Ok. Thank you. This makes sense. I had a different workflow envisioned (devs would just put the path in the CONST and not bother specifying MSRV manually) but it makes sense and now (devs will add MSRV themselves) and I understand why you suggested regex in the start.

iwyatt:
Next push to this pr will include three updates:
- fix the rust version dependency for the regex crate. Downgrading from 1.10 to 1.9x should fix
- fix some trailing bytes written to file when updating MSRV from a longer string to a shorter string and
- removing test paths from CONST

iwyatt:
Hi @djmitche - I think this is ready for another review. Thanks for your patience. 

- I implemented regex per your earlier suggestion. Now that I understand that you envision developers to add MSRV to a file themselves and also add the path to xtask (rather than just the latter), the use of regex masks sense to me. 

- I implemented a variety of checks for unwrapping Results - particularly for file IO. I wasn't sure whether or not it is a bit unnecessary for the size/scope of the project, though.

- I wasn't sure if your [comment about environment paths](https://github.com/GothenburgBitFactory/taskwarrior/pull/3189#discussion_r1359398928) was a suggestion for my local testing or for something to implement, so I didn't change the behavior of the script. I did remove my own tests paths from this commit, however. 

- I made the changes you suggested in the last review. 

- I had to try a few times to figure out that `cargo.lock` was why my regex version was causing the automated tests were failing, but I think I have that sorted now and it appears the latest version is passing. 

There are two limitations I am aware of that I didn't address: 
1. the regex pattern is case-sensitive and assumes 'MSRV' will always appear in caps. I felt like this was a reasonable assumption.
2. the [regex pattern](https://regex101.com/r/FAhCa5/1) assumes a general pattern of `MSRV = X.Y` but does not search for `Minimum Supported Rust Version` . I think this is reasonable and reduces chances of false-positive matches and over-writing parts of files that shouldn't be.

Please let me know what else can I clean up before it is ready to be merged and used!

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +41,106 @@ fn codegen() -> anyhow::Result<()> {
...
+                    path.display()
+                );
+                continue;
+            }
+
+            let mut file: File = File::options().read(true).write(true).open(path)?;
+            let reader = BufReader::new(&file);
+
+            // set the file_extension if it exists and is valid UTF-8
+            let mut file_extension = String::new();
```
djmitche: The clippy lint here is pointing out that `file_extension` is overwritten before it is read in every codepath, so it doesn't need to be initialized here. Same thing below for `comment_glyph`.

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +41,106 @@ fn codegen() -> anyhow::Result<()> {
...
+
+            // for each line in file
+            let mut file_string = String::new();
+            for line in reader.lines() {
+                let line_ref = &line.as_ref().unwrap();
+                let pattern_offset = re.find(line_ref);
+
+                // if a pre-existing MSRV pattern is found and is different, update it.
+                if pattern_offset.is_some() {
+                    if pattern_offset.unwrap().as_str() != replacement_string {
```
djmitche: This lint is pointing out that `if foo.is_some() { .. foo.unwrap() .. }` is better expressed as `if let Some(bar) = foo { .. bar .. }`.

djmitche (commented):
There are a few clippy checks below that I've highlighted. But, I'm a bit confused about what this is doing right now. It looks like, in a Rust file for example, it would look for a line like

```rust
    let x = y; // MSRV = 1.59
```
and if found turn that into
```rust
// MSRV = 1.60
    let x = y; // MSRV = 1.59
```

Maybe I can give a clearer description of how this typically works in other projects. Somewhere in the middle of ``, there appears

```yaml
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: "1.64" # MSRV - update with 'cargo xtask msrv' 
          override: true
```
(I've added the comment; you can do the same in the PR)

So, to update that MSRV:
 * look for a line ending with `# MSRV update with 'cargo xtask msrv'`
 * if not found, output the existing line
 * if found, replace `[0-9]\.[0-9]{2,}` with the new version within the line and output it.

Does that make more sense?

This might be easier to test if you factor out the "replace the MSRV" operation into a function that operates on a string. Then the `msrv` function can read the file content to a string, call the replacement function, and write the result back out. And tests can just provide a test string to the replacement function.

iwyatt:
Thanks for the feedback @djmitche . I had incorrectly understood and assumed that the intent was to update **comments** in files that specify the MSRV to be used, therefore my approach was to look for comments in files that contained `MSRV` and replace those comments with the X.Y argument.

Instead, it sounds like the intent is to update any given configuration file's specifications. So, similar to what you wrote above, in `checks.yml` where the following appears...

```yaml
      - uses: actions-rs/toolchain@v1
        with:
          # Same as the MSRV in rust-tests.yml
          toolchain: "1.64"
```

...the version in the line ```toolchain: "1.64"``` should be updated to whatever X.Y is passed as an argument to xtask msrv.

If that is the case, then my approach would be to look for a different X.Y and update it with the new X.Y where `MSRV` (and similar phrases) appears on the same line, or the preceding line.

Do I have that correct?

iwyatt:
Pushed a commit to fix clippy issues, but still working on functionality to update file contents (instead of just file comments). (Had much less time this week).

djmitche:
I think you have that correct! 

[Here](https://github.com/taskcluster/taskcluster/blob/fa5cc1ae21600f63c74fb0a99fc6da2f5667ea99/infrastructure/tooling/src/release/tasks.js#L135-L192) is an example of doing something similar, but in JS. This code writes out a new version number in a bunch of places. The `modifyRepoFile` method reads a file, calls a function with the content, then writes the result back to the file. As you can see, each file has a special way of replacing the content, that _only_ matches the place to change the value, and not anywhere else. Those have needed a bit of a tweak now and then, but have been pretty reliable. In my memory, some of them used a comment to help find the right place to modify, but that appears not to be the case!

Anyway, within Taskwarrior, as long as we can be pretty confident that only the correct value will be changed, and other parts of the file won't be accidentally modified, any kind of replacement strategy is fine.

iwyatt:
Good afternoon @djmitche - I changed out the script action to update the actual version number content of a file instead of managing the comments.

I examined the example you linked and found that the onus for configuration is put on the developer to 1) add their file to the path list and 2) specify the regex that is used to properly find the line containing the MSRV.

Taking a similar approach in this script, a developer would add the appropriate file path and regex to the CONST tuple array. For which they would need to be responsible for assuring that their provided regex does not result in inadvertent matches within the file they specify. 

- I'm still a little unsure if how I specified relative paths in the `MSRV_PATH_REGEX` for `check.yml` and `rust-tests.yml` is how these would work for all developers. Guidance there would be welcome.

If there are any further updates needed, I should have more availability next week to get a quicker turn, so thanks for the patience this week as I had less capacity.

taskchampion/xtask/src/main.rs
```diff
@@ -3,17 +3,42 @@
 //! At the moment it is very simple, but if this grows more subcommands then
 //! it will be sensible to use `clap` or another similar library.
 
+//! The const MSRV_FILE_PATHS_REGEX in /xtask/main.rs is an array of tuples
```
djmitche:
```suggestion
//! The const MSRV_PATHS_REGEX in /xtask/main.rs is an array of tuples
```
you could also move this rustdoc comment down a few lines to document the `MSRV_PATH_REGEX` itself.

taskchampion/xtask/src/main.rs
```diff
@@ -3,17 +3,42 @@
...
-    match arg.as_deref() {
-        Some("codegen") => codegen(),
-        Some(arg) => anyhow::bail!("unknown xtask {}", arg),
-        _ => anyhow::bail!("unknown xtask"),
+    let arguments: Vec<String> = env::args().collect();
+    if arguments.len() < 2 {
+        anyhow::bail!("xtask: Valid arguments are: `codegen`, `msrv <version x.y>`");
+    }
+
+    println!("{:#?}", env::current_dir());
```
djmitche: I ran this script in the root of the repository, and as you suggested the paths didn't work:
```
xtask: Warning: xtask: Path "../../.github/workflows/rust-tests.yml" not found. Skipping.
xtask: Warning: xtask: Path "../../.github/workflows/checks.yml" not found. Skipping.
```
However, I updated this line to print all env vars, and saw
```
"CARGO_MANIFEST_DIR", "/home/dustin/p/taskwarrior/taskchampion/xtask"
```
which is the directory your paths are relative to. Perhaps you can use that environment variable to make this work regardless of directory?

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +54,110 @@ fn codegen() -> anyhow::Result<()> {
...
+                            "{}",
+                            format!(
+                                "{}\n",
+                                re_msrv_version
+                                    .replace(line_ref, version_replacement_string)
+                                    .to_string()
+                                    .as_str()
+                            )
+                            .as_str()
+                        );
```
djmitche: There's an impressive amount of unnecessary code here!
```suggestion
                        println!(
                            "{}",
                                re_msrv_version
                                    .replace(line_ref, version_replacement_string)
                          );
```

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +54,110 @@ fn codegen() -> anyhow::Result<()> {
...
+
+                        file_string += format!(
+                            "{}\n",
+                            re_msrv_version
+                                .replace(line_ref, version_replacement_string)
+                                .to_string()
+                                .as_str()
+                        )
+                        .as_str();
```
djmitche: It seems like the `Cow` type returned from `Regex::replace` caused some trouble :)
```suggestion
                        file_string += 
                            &re_msrv_version
                                .replace(line_ref, version_replacement_string);
```

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +54,110 @@ fn codegen() -> anyhow::Result<()> {
...
+            let reader = BufReader::new(&file);
+
+            // set search string and the replacement string for version number content
+            let re_msrv_pattern = Regex::new(msrv_file.1).unwrap();
+            let version_replacement_string = &args[2];
+
+            // for each line in file
+            let mut file_string = String::new();
+            for line in reader.lines() {
+                let line_ref = &line.as_ref().unwrap();
```
djmitche: The auto-deref functionality of the language will automatically make references to an object when calling a method on it, so this is the same as:
```suggestion
                let line_ref = &line.unwrap();
```

taskchampion/xtask/src/main.rs
```diff
@@ -29,3 +54,110 @@ fn codegen() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// `cargo xtask msrv (X.Y)`
+///
+/// This checks and updates the Minimum Supported Rust Version for all files specified in the tuple list `MSRV_PATH_REGEX` const in `xtask/src/main.rs` where the regex pattern specified is found in the file.
+/// The argument "X.Y" will replace existing X.Y any place the regex pattern specified in the tuple list `MSRV_PATH_REGEX` is found.
```
djmitche: Clarifying a little:
```suggestion
/// This checks and updates the Minimum Supported Rust Version for all files specified in MSRV_PATH_REGEX`.
/// Each line where the regex matches will have all values of the form `#.##` replaced with the given MSRV.
```

djmitche (commented):
This looks good and does the job!

I've left a few comments to suggest more concise code. I suspect there are more ways to "tighten up" this function. Removing some of the debugging prints, once you're happy with the functionality, would be good too.

taskchampion/xtask/src/main.rs
iwyatt: Ah makes sense! I should have considered refactoring that as I basically updated the previous code in-place. Thanks for providing that solution.

taskchampion/xtask/src/main.rs
iwyatt: Oh man! This was a big "DERP!" on my part. Not only did you mention environment variables before, but the `codegen()` function literally uses the code already. /facepalm Sorry you had to spell it out for me.

iwyatt:
Hi @djmitche - I think I've addressed your concerns in your feedback.

I assume that a developer would use this manually, infrequently, and specifically within the context of a particular project, so they would only add paths to xtask's `main.rs` that exist within the context of their current project or child projects.

So the `.yml` files that I had been using in testing, which are part of a cousin(uncle?) folder's project. As such, xtask would have to be run in origin/trunk project directory and add paths relative to that. 

Because I'm not 100% familiar with how this type of tool gets used and I'm trying not to bother with too many questions, the assumption above may be incorrect. (But I am very grateful for the guidance and patience you've provided!).

Please let me know if there is anything else I can address for you!

djmitche:
> Hi @djmitche - I think I've addressed your concerns in your feedback.
> 
> I assume that a developer would use this manually, infrequently, and specifically within the context of a particular project, so they would only add paths to xtask's `main.rs` that exist within the context of their current project or child projects.

This tool would only be used by this repository, so it can definitely be specific to taskwarrior. I'm not sure if that's what you meant by "project" here -- but anything in the repo is fair game.

> So the `.yml` files that I had been using in testing, which are part of a cousin(uncle?) folder's project. As such, xtask would have to be run in origin/trunk project directory and add paths relative to that.

Yep, that's the directory indicated by CARGO_MANIFEST_DIR.

> Because I'm not 100% familiar with how this type of tool gets used and I'm trying not to bother with too many questions, the assumption above may be incorrect. (But I am very grateful for the guidance and patience you've provided!).

To be clear, this isn't a general MSRV-management tool. It's meant specifically for this repository, which is why it's OK to hard-code the filenames in the source flie.

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+use std::io::{BufRead, BufReader};
+use std::io::{Seek, Write};
+use std::path::{Path, PathBuf};
+
+/// The const MSRV_PATHS_REGEX in /xtask/main.rs is an array of tuples
+///  of the form (PATH, REGEX) where PATH and REGEX are literals where
+///  PATH is a file that updates the Minimum Supported Rust Version
+///  and REGEX is the pattern to find the appropriate line in the file
+
+// Increment length of array when adding tuples of (PATH, REGEX).
```
djmitche: Let's make this all a single docstring, instead of referring to the identifier by name and filename.

```suggestion
/// Tuples of the form (PATH, REGEX) where PATH and REGEX are literals where PATH is a file that
/// conains the Minimum Supported Rust Version and REGEX is the pattern to find the appropriate
/// line in the file. PATH is relative to the `taskchampion/` directory in the repo.
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+const MSRV_PATH_REGEX: [(&str, &str); 1] = [(
+    "xtask/Cargo.toml",
+    r#"rust-version = "[0-9]+("|\.|[0-9])+""#,
+)];
```
djmitche: Using a slice here, instead of an array, removes the need to specify the length. Let's also use the actual files we'd like to modify.
```suggestion
const MSRV_PATH_REGEX: &[(&str, &str)] = &[                                                                                                                                                                                                                                                                                  
    (       
        "../.github/workflows/checks.yml",
        r#"toolchain: "[0-9.]+*" # MSRV"#,                                                                                                                                                                                                                                                                                     
    ),      
    ("../.github/workflows/rust-tests.yml", r#""[0-9.]+" # MSRV"#),                                                                                            
    (         
        "taskchampion/src/lib.rs",                                                                                                                             
        r#"Rust version [0-9.]* and higher"#,                                                                                                                  
    ),  
];
```

I also made the following changes in the github workflow files, so that the regexes above will work:
```patch
diff --git .github/workflows/checks.yml .github/workflows/checks.yml
index b1370c905..64de42ce6 100644
--- .github/workflows/checks.yml
+++ .github/workflows/checks.yml
@@ -24,19 +24,17 @@ jobs:
       - name: Cache cargo build
         uses: actions/cache@v3
         with:
           path: target
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
 
       - uses: actions-rs/toolchain@v1
         with:
-          # Fixed version for clippy lints.  Bump this as necesary.  It must not
-          # be older than the MSRV in tests.yml.
-          toolchain: "1.64"
+          toolchain: "1.64" # MSRV
           override: true
 
       - uses: actions-rs/cargo@v1.0.3
         with:
           command: check
 
       - run: rustup component add clippy
 
diff --git .github/workflows/rust-tests.yml .github/workflows/rust-tests.yml
index 776642ee6..be2b7c711 100644
--- .github/workflows/rust-tests.yml
+++ .github/workflows/rust-tests.yml
@@ -10,18 +10,17 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
 jobs:
   test:
     strategy:
       matrix:
         rust:
-          # MSRV; must not be higher than the clippy rust version in checks.yml
-          - "1.64"
+          - "1.64" # MSRV
           - "stable"
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
 
     name: "rust ${{ matrix.rust }} on ${{ matrix.os }}"
     runs-on: ${{ matrix.os }}
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+
+    // set regex for replacing version number only within the pattern found within a line
+    let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+")?;
+
+    // for each file in const paths tuple
+    for msrv_file in MSRV_PATH_REGEX {
+        let mut is_pattern_in_file = false;
+
+        let file_path = workspace_dir.join(msrv_file.0);
+        let path: &Path = Path::new(&file_path);
```
djmitche: First, Rust infers types, so there's no need for the `: &Path`. Second, a common idiom in Rust when changing the type of something without really changing what it is (here PathBuf -> Path), is to redeclare the variable with a new name, rather than invent two different names:
```suggestion
        let path = workspace_dir.join(msrv_file.0);
        let path = Path::new(&file_path);
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+        if !&path.exists() {
+            anyhow::bail!("xtask: path does not exist {}", &path.display());
+        };
+
+        let mut file: File = File::options().read(true).write(true).open(path)?;
```
djmitche:
```suggestion
        let mut file= File::options().read(true).write(true).open(path)?;
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+    // set regex for replacing version number only within the pattern found within a line
+    let re_msrv_version = Regex::new(r"([0-9]+(\.|[0-9]+|))+")?;
+
+    // for each file in const paths tuple
+    for msrv_file in MSRV_PATH_REGEX {
+        let mut is_pattern_in_file = false;
+
+        let file_path = workspace_dir.join(msrv_file.0);
+        let path: &Path = Path::new(&file_path);
+        if !&path.exists() {
```
djmitche: Method calls automatically add `&` as necessary, so this is unnecessary:
```suggestion
        if !path.exists() {
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+
+                    file_string += "\n";
+
+                    is_pattern_in_file = true;
+                    continue;
+                }
+            }
+
+            file_string += line_ref;
+            file_string += "\n";
```
djmitche: Similar idiom here of replacing `line` (of type `Result<String, ..>`) with `line` of type `&String`.

I'd be tempted to write `let line = line?;` and use `&line` within the remainder, but that's probably personal preference.
```suggestion
            let line = &line?;

            // if rust version pattern is found and is different, update it
            if let Some(pattern_offset) = re_msrv_pattern.find(line) {
                if !pattern_offset.as_str().contains(version_replacement_string) {
                    file_string += &re_msrv_version.replace(line, version_replacement_string);

                    file_string += "\n";

                    is_pattern_in_file = true;
                    continue;
                }
            }

            file_string += line;
            file_string += "\n";
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+        }
+
+        // if pattern was found and updated, write to disk
+        if is_pattern_in_file {
+            //  Set the file length to the file_string length
+            let _ = file.set_len(file_string.len() as u64);
```
djmitche: I suspect you added `let _ = ..` because Rust complained that you weren't using the `Result` value.  This silences the warning, but will miss any errors from the operation. Instead, use `?` to handle the result.  The type of the `...?` expression will be `()` which doesn't need to be assigned anywhere.
```suggestion
            file.set_len(file_string.len() as u64)?;
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+
+            //  set the cursor to the beginning of the file and write
+            let _ = file.seek(std::io::SeekFrom::Start(0));
```
djmitche: Same here.

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+            let file_write_result = file.write(file_string.as_bytes());
+
+            // if error, print error messege and exit
+            if file_write_result.is_err() {
+                anyhow::bail!("xtask: unable to write file to disk: {}", &path.display());
+            }
```
djmitche: `?` can handle this, too:
```suggestion
            file.write(file_string.as_bytes())?;
```

taskchampion/xtask/src/main.rs
```diff
@@ -3,29 +3,123 @@
...
+        if !&path.exists() {
+            anyhow::bail!("xtask: path does not exist {}", &path.display());
+        };
+
+        let mut file: File = File::options().read(true).write(true).open(path)?;
+        let reader = BufReader::new(&file);
+
+        // set search string and the replacement string for version number content
+        let re_msrv_pattern = Regex::new(msrv_file.1)?;
+        let version_replacement_string = &args[2];
```
djmitche: This doesn't change in each iteration of the loop, so it probably makes sense to move this variable up to the top where you check the content of this argument.

djmitche (changes requested):
Some comments to make this a bit more idiomatic, but this is definitely looking more concise already!

taskchampion/xtask/src/main.rs
iwyatt: Ok got it. I like your style, though I had to think through it a bit and ask others to clarify my thinking about what was happening with the memory and pointers - and I had forgotten about shadowing in Rust.

taskchampion/xtask/src/main.rs
iwyatt: Clippy complained that `file.write()?` goes unhangled, and suggested I change it to `file.write_all()` which worked great. I had used `write_all()` before, but apparently I accidentally imported the unix-specific flavor, which caused the CI checks to fail for windows. Interesting learning!

taskchampion/xtask/src/main.rs
iwyatt: I'm assuming I don't need to include the changes to these configuration files as part of my own PR, but please let me know if that is incorrect. I did change the array to a slice per your recommendation, which makes sense and hadn't considered that as an option.

djmitche: Yes, please do include these changes in `.github` as well -- without them, `cargo xtask msrv` won't update these files, so arguably the job isn't finished :)

djmitche:
This is looking great! I think the `msrv` function is a lot clearer now. Just get those `.github/` changes in, and we can get this merged!

Thanks for sticking with this through all the revisions..

iwyatt:
Hi @djmitche I updated the two .yml files . I also noticed in `checks.yml` has the rust version specified in two places, so I added ` # MSRV` for the other one in that file in addition to the one you mentioned. 

I also see that `checks.yml` has a version specification for `mdbook-version` with a comment to update it in `publish-docs.yml`. This would appear to follow the same use case as the MSRV project I'm working on. Is that something we want to do in a new project after we close out MSRV, or maybe save for another 'good first issue' ?

djmitche (approved):
Nice work! And good job spotting that other place in `checks.yml`.

Bumping the mdbook version is much less common, and less critical that it be done completely, so I don't think we need to add an `xtask` command for it.

I'll get this merged. Thanks again!

djmitche:
Did you have ideas what you'd like to work on next?

iwyatt:
I haven't quite figured it out yet.

Here are some unvetted / initial ideas:

## Unvetted Project Ideas
- Website work to update tw.org:
	- Not ideal since there is little to no Rust involved. Helps task warrior ecosystem.
- Nushell Shell Completer:
	- Little/no actual Rust involved, but helpful to my use case and interesting. Seems moderately straight forward.
- Build a TaskWarrior PWA:
	- Most could be done in Rust. Big benefit to Task Warrior Ecosystem. Solve some personal requirements for a task manager. 
- Build a Firefox Extension to capture tasks from browser context (eg email / web pages):
	- No Rust involved. Something similar using Python and Bookmarks already exists. Would solve personal need for a task manager.

## Objectives
The ideal project is an intersection of the following:
1) is appropriate for my rust skills
2) would be helpful to Taskwarrior
3) would be helpful for me to learn given my goals:

**Primary Goal:**
- learn rust
	- would be nice to have enough skill to be employable as a Rust developer
	- or have enough Rust skill to build something myself that I can make a dollar from

**Secondary Goals:**
- Help make taskwarrior something I can use for my primary/only task manager todo app
	- requires browser extension, mobile app, sync

## Advice?
I've been embarking on a journey to build, hone, and develop enough coding skill to call myself and actual 'programmer' ([context](https://www.isaacwyatt.com/coding/)). If I were an intern on your team in a professional setting, what feedback / advice would you give me?

If you don't mind, I'll ping you on Discord for a personal, non-public, response.

## FYI
I'll be writing a short blog post about what I learned building xtask MSRV.

Issue #3058, opened 2023-02-04
Reconsider "level" of the Task/TaskMut interface


djmitche:
The problem that brought me [to #3029] is, Taskwarrior _mostly_ doesn't call functions like `set_description` or `add_annotation` -- it interfaces directly with the key/value pairs that make up a task, via `set_value`.  But it does call `set_status` since that updates the working set.  So the issue was that Taskwarrior would call

 * `t.set_value("modified", "12345")`
 * `t.set_status(Status::Pending)`

and the second call would not "see" that the modified timestamp had already been set, and would do so again.  On import, this causes a problem by setting the "modified" timestamp of all tasks to the time the import occurred.

Overall, this is a major "impedance mismatch" between TC and TW: TW wants to handle tasks as key/value maps, while TC was really designed for API consumers to call methods that embody more knowledge about the meaning of particular keys.  These are more than just convenience wrappers, though: they also have this set-the-modified-timestamp thing, as well as automatically updating the working set.  And (referring to #3028) it might be good to cache parent/child relationships as well, to avoid having to load most of the pending tasks one-by-one.  I've been mulling this over for a while but haven't come up with a concrete way to do better.

_Originally posted by @djmitche in https://github.com/GothenburgBitFactory/taskwarrior/issues/3029#issuecomment-1414569579_

djmitche:
cc @ryneeverett continuing discussion from #3029.

ryneeverett:
A better understanding of why we're dissatisfied with the current interface would be helpful for developing and evaluating solutions. When a refactor doesn't solve a concrete problem or enable a concrete use case, it can be difficult to tell whether it is an actual improvement. In the case of #3029, I would suggest that we were uncomfortable with that solution because it was not encapsulated in `TaskMut` but rather dependent upon the consumer manually setting the "modified' attribute before calling a method which automatically calls `TaskMut::update_modified()`.

[...]

Issue #2928, opened 2022-08-28
Expose operations from TaskChampion


djmitche:
We'll need this for a few reasons:
 * To generate a diff for undo (get the list of operations since the last undo point, apply them in reverse, and generate the diff)
 * To generate the "history" of a task (which means we'll need to store operations per-task, until a task is deleted)

djmitche:
@masaeedu is considering working on this, so I'll un-assign from myself.

djmitche:
@ryneeverett I think this is done -- is there anything more to do?

ryneeverett:
Yes, #3213 only addressed the "Operations for Undo" section of your detailed breakdown. "Operations for Info" has not yet been addressed.

djmitche:
Right, thank you!

Issue #2887, opened 2022-07-17
src/tc/rust doesn't correctly depend on taskchampion/*


djmitche:
The `src/tc/rust` CMake target (tc_rust) doesn't automatically get rebuilt if you change something in `taskchampion`, despite depending on the crates in that directory.  This is because the dependency is represented in Cargo.toml, and CMake is unaware of it.

However, I don't know how to fix this with CMake.

Any change in `taskchampion/` should trigger a rebuild of `tc_rust`.

djmitche:
As a note-to-self you can work around this by running `make -C src/tc/rust clean` to force-rebuild taskchampion.

Shadow53:
Did a search on this, found [this pdf](https://webdevolutions.blob.core.windows.net/blog/pdf/insider-series-using-rust-code-in-a-cc-project-with-cmake.pdf), which links to [this CMake module](https://github.com/Devolutions/CMakeRust).

dathanb:
I think this might be resolved already?

```bash
$ make
[ 15%] Built target task
[ 29%] Built target libshared
[ 52%] Built target columns
[ 92%] Built target commands
[ 96%] Built target tc
[ 96%] Built target tc-rust_target
[ 98%] Built target task_executable
[ 98%] Built target calc_executable
[100%] Built target lex_executable
$ echo "pub fn foo(){}" >> src/tc/rust/src/lib.rs
$ make
[ 15%] Built target task
[ 29%] Built target libshared
[ 52%] Built target columns
[ 92%] Built target commands
[ 96%] Built target tc
[ 96%] running cargo
   Compiling tc-rust v0.1.0 (/Users/dathanb/dev/dathanb/taskwarrior/src/tc/rust)
    Finished dev [unoptimized + debuginfo] target(s) in 0.24s
[ 96%] Built target tc-rust_target
[ 98%] Linking CXX executable task
[ 98%] Built target task_executable
[ 98%] Linking CXX executable calc
[ 98%] Built target calc_executable
[ 98%] Linking CXX executable lex
[100%] Built target lex_executable
```

djmitche:
We are using CMakeRust, and that does work. However, @dathanb's reproduction doesn't quite hit the issue.  Instead, try modifying [`taskchampion/taskchampion/src/lib.rs`](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/taskchampion/taskchampion/src/lib.rs), and noting that CMake doesn't rebuild anything.

[...]

PR #3213, opened 2023-12-21
show undo diff

ryneeverett:
#### Description

A start at phase 1 of GothenburgBitFactory/taskchampion#373. The rust compiles but the C++ does not yet. I was hoping to solicit some high-level guidance before proceeding, namely:

1. Does the callback approach make sense?
  a. I originally started looking at this approach when i realized that the use of dynamic storage backends (`TaskDb.storage`) seems to mean that no transaction could outlive a method (without cloning).
  b. I don't know how Storage is generally guarded against race conditions but the callback method seems like it will take care of this concern.
2. What is the best way to pass the data (`UndoDiff`) through ffi? For example, should I pass the values separately or can I pass the whole struct? Should I serialize it deep in taskchampion (undo.rs) or at the interface (taskchampion/lib/src/replica.rs)? I'm under the vague impression that as long as the data is typed with the right traits the necessary transformations to pass through ffi happen automatically...at least it seems like it could/should work that way.

### DIFF
--- a/src/tc/Replica.h
+++ b/src/tc/Replica.h
@@ -95,6 +95,7 @@ namespace tc {
     void sync(Server server, bool avoid_snapshots);
     tc::ffi::TCReplicaOpList get_undo_ops ();
     void commit_undo_ops (tc::ffi::TCReplicaOpList tc_undo_ops, int32_t *undone_out);
+    void free_replica_ops (tc::ffi::TCReplicaOpList tc_undo_ops);
     int64_t num_local_operations ();
     int64_t num_undo_points ();
 // TODO: TCResult tc_replica_add_undo_point(struct TCReplica *rep, bool force);
--- a/src/tc/Replica.h
+++ b/src/tc/Replica.h
@@ -93,7 +93,8 @@ namespace tc {
     tc::Task import_task_with_uuid (const std::string &uuid);
 // TODO: struct TCTask *tc_replica_import_task_with_uuid(struct TCReplica *rep, struct TCUuid tcuuid);
     void sync(Server server, bool avoid_snapshots);
-    void undo (int32_t *undone_out, bool(*condition)(struct tc::ffi::TCUndoDiff));
+    tc::ffi::TCReplicaOpList get_undo_ops ();
+    void commit_undo_ops (tc::ffi::TCReplicaOpList tc_undo_ops, int32_t *undone_out);
     int64_t num_local_operations ();
     int64_t num_undo_points ();
 // TODO: TCResult tc_replica_add_undo_point(struct TCReplica *rep, bool force);
--- a/src/tc/Replica.h
+++ b/src/tc/Replica.h
@@ -93,7 +93,7 @@ namespace tc {
     tc::Task import_task_with_uuid (const std::string &uuid);
 // TODO: struct TCTask *tc_replica_import_task_with_uuid(struct TCReplica *rep, struct TCUuid tcuuid);
     void sync(Server server, bool avoid_snapshots);
-    void undo (int32_t *undone_out);
+    void undo (int32_t *undone_out, bool(*condition)(struct tc::ffi::TCUndoDiff));
     int64_t num_local_operations ();
     int64_t num_undo_points ();
 // TODO: TCResult tc_replica_add_undo_point(struct TCReplica *rep, bool force);
--- a/src/tc/Replica.h
+++ b/src/tc/Replica.h
@@ -96,6 +96,12 @@ namespace tc {
     tc::ffi::TCReplicaOpList get_undo_ops ();
     void commit_undo_ops (tc::ffi::TCReplicaOpList tc_undo_ops, int32_t *undone_out);
     void free_replica_ops (tc::ffi::TCReplicaOpList tc_undo_ops);
+    std::string get_op_uuid(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_property(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_value(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_old_value(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_timestamp(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_old_task_description(tc::ffi::TCReplicaOp &tc_replica_op) const;
     int64_t num_local_operations ();
     int64_t num_undo_points ();
 // TODO: TCResult tc_replica_add_undo_point(struct TCReplica *rep, bool force);
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -1,6 +1,6 @@
 use crate::traits::*;
 use crate::types::*;
-use std::ptr::null_mut;
+use crate::util::vec_into_raw_parts;
 use taskchampion::storage::TaskMap;
 
 #[ffizz_header::item]
@@ -122,10 +122,11 @@ impl CList for TCKVList {
 
 impl Default for TCKVList {
     fn default() -> Self {
+        let (ptr, len, capacity) = vec_into_raw_parts(Vec::default());
         Self {
-            len: libc::size_t::default(),
-            _capacity: libc::size_t::default(),
-            items: null_mut(),
+            items: ptr,
+            len,
+            _capacity: capacity,
         }
     }
 }
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -1,7 +1,5 @@
 use crate::traits::*;
 use crate::types::*;
-use crate::util::vec_into_raw_parts;
-use taskchampion::storage::TaskMap;
 
 #[ffizz_header::item]
 #[ffizz(order = 600)]
@@ -77,22 +75,6 @@ pub struct TCKVList {
     /// total size of items (internal use only)
     pub _capacity: libc::size_t,
     pub items: *mut TCKV,
-}
-
-impl From<TaskMap> for TCKVList {
-    fn from(taskmap: TaskMap) -> TCKVList {
-        let vec = taskmap
-            .iter()
-            .map(|(k, v)| {
-                let key = RustString::from(k.as_ref());
-                let value = RustString::from(v.as_ref());
-                TCKV::as_ctype((key, value))
-            })
-            .collect();
-        // SAFETY:
-        //  - caller will free this list
-        unsafe { TCKVList::return_val(vec) }
-    }
 }
 
 impl CList for TCKVList {
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -19,6 +19,7 @@ use taskchampion::storage::TaskMap;
 /// } TCKV;
 /// ```
 #[repr(C)]
+#[derive(Debug)]
 pub struct TCKV {
     pub key: TCString,
     pub value: TCString,
@@ -70,6 +71,7 @@ impl PassByValue for TCKV {
 /// } TCKVList;
 /// ```
 #[repr(C)]
+#[derive(Debug)]
 pub struct TCKVList {
     pub len: libc::size_t,
     /// total size of items (internal use only)
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -1,5 +1,6 @@
 use crate::traits::*;
 use crate::types::*;
+use taskchampion::storage::TaskMap;
 
 #[ffizz_header::item]
 #[ffizz(order = 600)]
@@ -69,10 +70,26 @@ impl PassByValue for TCKV {
 /// ```
 #[repr(C)]
 pub struct TCKVList {
-    len: libc::size_t,
+    pub len: libc::size_t,
     /// total size of items (internal use only)
-    _capacity: libc::size_t,
-    items: *mut TCKV,
+    pub _capacity: libc::size_t,
+    pub items: *mut TCKV,
+}
+
+impl From<TaskMap> for TCKVList {
+    fn from(taskmap: TaskMap) -> TCKVList {
+        let vec = taskmap
+            .iter()
+            .map(|(k, v)| {
+                let key = RustString::from(k.as_ref());
+                let value = RustString::from(v.as_ref());
+                TCKV::as_ctype((key, value))
+            })
+            .collect();
+        // SAFETY:
+        //  - caller will free this list
+        unsafe { TCKVList::return_val(vec) }
+    }
 }
 
 impl CList for TCKVList {
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -122,12 +122,9 @@ impl CList for TCKVList {
 
 impl Default for TCKVList {
     fn default() -> Self {
-        let (ptr, len, capacity) = vec_into_raw_parts(Vec::default());
-        Self {
-            items: ptr,
-            len,
-            _capacity: capacity,
-        }
+        // SAFETY:
+        //  - caller will free this list
+        unsafe { TCKVList::return_val(Vec::new()) }
     }
 }
 
--- a/taskchampion/lib/src/kv.rs
+++ b/taskchampion/lib/src/kv.rs
@@ -1,5 +1,6 @@
 use crate::traits::*;
 use crate::types::*;
+use std::ptr::null_mut;
 use taskchampion::storage::TaskMap;
 
 #[ffizz_header::item]
@@ -114,6 +115,16 @@ impl CList for TCKVList {
 
     fn into_raw_parts(self) -> (*mut Self::Element, usize, usize) {
         (self.items, self.len, self._capacity)
+    }
+}
+
+impl Default for TCKVList {
+    fn default() -> Self {
+        Self {
+            len: libc::size_t::default(),
+            _capacity: libc::size_t::default(),
+            items: null_mut(),
+        }
     }
 }
 
--- a/taskchampion/lib/src/replica.rs
+++ b/taskchampion/lib/src/replica.rs
@@ -133,6 +133,30 @@ where
 }
 
 #[ffizz_header::item]
+#[ffizz(order = 900)]
+/// ***** TCReplicaOpType *****
+///
+/// ```c
+/// typedef enum TCReplicaOpType {
+///     Create,
+///     Delete,
+///     Update,
+///     Undopoint,
+/// } TCReplicaOpType;
+/// ```
+#[derive(Default)]
+pub enum TCReplicaOpType {
+    Create,
+    Delete,
+    Update,
+    UndoPoint,
+    #[default]
+    Error,
+}
+
+impl PassByPointer for TCReplicaOpType {}
+
+#[ffizz_header::item]
 #[ffizz(order = 901)]
 /// Create a new TCReplica with an in-memory database.  The contents of the database will be
 /// lost when it is freed with tc_replica_free.
@@ -218,37 +242,27 @@ impl From<TCKVList> for TaskMap {
 /// ***** TCReplicaOp *****
 ///
 /// ```c
-/// enum TCReplicaOp {
-///    Create { struct TCUuid uuid },
-///    Delete { struct TCUuid uuid, TCKVList old_task },
-///    Update {
-///        struct TCUuid uuid,
-///        struct TCString property,
-///        struct TCString old_value,
-///        struct TCString value,
-///        struct TCString timestamp,
-///    },
-///    UndoPoint,
-/// }
+/// struct TCReplicaOp {
+///     TCReplicaOpType operation_type;
+///     TCUuid uuid;
+///     TCKVList old_task;
+///     TCString property;
+///     TCString old_value;
+///     TCString value;
+///     TCString timestamp;
+/// };
 ///
 /// typedef struct TCReplicaOp TCReplicaOp;
 /// ```
-pub enum TCReplicaOp {
-    Create {
-        uuid: TCUuid,
-    },
-    Delete {
-        uuid: TCUuid,
-        old_task: TCKVList,
-    },
-    Update {
-        uuid: TCUuid,
-        property: TCString,
-        old_value: TCString,
-        value: TCString,
-        timestamp: TCString,
-    },
-    UndoPoint,
+#[derive(Default)]
+pub struct TCReplicaOp {
+    operation_type: TCReplicaOpType,
+    uuid: TCUuid,
+    old_task: TCKVList,
+    property: TCString,
+    old_value: TCString,
+    value: TCString,
+    timestamp: TCString,
 }
 
 impl PassByPointer for TCReplicaOp {}
@@ -256,16 +270,20 @@ impl PassByPointer for TCReplicaOp {}
 impl From<ReplicaOp> for TCReplicaOp {
     fn from(replica_op: ReplicaOp) -> TCReplicaOp {
         match replica_op {
-            ReplicaOp::Create { uuid } => TCReplicaOp::Create {
+            ReplicaOp::Create { uuid } => TCReplicaOp {
+                operation_type: TCReplicaOpType::Create,
                 // SAFETY:
                 //  - caller promises to free this value.
                 uuid: unsafe { TCUuid::return_val(uuid) },
+                ..Default::default()
             },
-            ReplicaOp::Delete { uuid, old_task } => TCReplicaOp::Delete {
+            ReplicaOp::Delete { uuid, old_task } => TCReplicaOp {
+                operation_type: TCReplicaOpType::Delete,
                 // SAFETY:
                 //  - caller promises to free this value.
                 uuid: unsafe { TCUuid::return_val(uuid) },
                 old_task: TCKVList::from(old_task),
+                ..Default::default()
             },
             ReplicaOp::Update {
                 uuid,
@@ -279,7 +297,8 @@ impl From<ReplicaOp> for TCReplicaOp {
                 let value_ruststring = RustString::String(value.unwrap_or_default());
                 let timestamp_ruststring = RustString::String(timestamp.to_string());
 
-                TCReplicaOp::Update {
+                TCReplicaOp {
+                    operation_type: TCReplicaOpType::Update,
                     // SAFETY:
                     //  - caller promises to free this value.
                     uuid: unsafe { TCUuid::return_val(uuid) },
@@ -295,9 +314,13 @@ impl From<ReplicaOp> for TCReplicaOp {
                     // SAFETY:
                     //  - caller promises to free this value.
                     timestamp: unsafe { TCString::return_val(timestamp_ruststring) },
+                    ..Default::default()
                 }
             }
-            ReplicaOp::UndoPoint => TCReplicaOp::UndoPoint,
+            ReplicaOp::UndoPoint => TCReplicaOp {
+                operation_type: TCReplicaOpType::UndoPoint,
+                ..Default::default()
+            },
         }
     }
 }
@@ -305,23 +328,34 @@ impl From<ReplicaOp> for TCReplicaOp {
 impl From<TCReplicaOp> for ReplicaOp {
     fn from(tc_replica_op: TCReplicaOp) -> ReplicaOp {
         match tc_replica_op {
-            TCReplicaOp::Create { uuid } => ReplicaOp::Create {
+            TCReplicaOp {
+                operation_type: TCReplicaOpType::Create,
+                uuid,
+                ..
+            } => ReplicaOp::Create {
                 // SAFETY:
                 //  - uuid is a valid TCUuid (all byte patterns are valid)
                 uuid: unsafe { TCUuid::val_from_arg(uuid) },
             },
-            TCReplicaOp::Delete { uuid, old_task } => ReplicaOp::Delete {
+            TCReplicaOp {
+                operation_type: TCReplicaOpType::Delete,
+                uuid,
+                old_task,
+                ..
+            } => ReplicaOp::Delete {
                 // SAFETY:
                 //  - uuid is a valid TCUuid (all byte patterns are valid)
                 uuid: unsafe { TCUuid::val_from_arg(uuid) },
                 old_task: TaskMap::from(old_task),
             },
-            TCReplicaOp::Update {
+            TCReplicaOp {
+                operation_type: TCReplicaOpType::Update,
                 uuid,
                 property,
                 old_value,
                 value,
                 timestamp,
+                ..
             } => {
                 // SAFETY:
                 //  - uuid is a valid TCUuid (all byte patterns are valid)
@@ -355,7 +389,14 @@ impl From<TCReplicaOp> for ReplicaOp {
                         .unwrap(),
                 }
             }
-            TCReplicaOp::UndoPoint => ReplicaOp::UndoPoint,
+            TCReplicaOp {
+                operation_type: TCReplicaOpType::UndoPoint,
+                ..
+            } => ReplicaOp::UndoPoint,
+            TCReplicaOp {
+                operation_type: TCReplicaOpType::Error,
+                ..
+            } => panic!("This shouldn't be possible."),
         }
     }
 }
@@ -366,10 +407,10 @@ impl From<TCReplicaOp> for ReplicaOp {
 ///
 /// ```c
 /// struct TCReplicaOpList {
-///     void* ptr
-///     size_t len
-///     size_t capacity
-/// }
+///     void* ptr;
+///     size_t len;
+///     size_t capacity;
+/// };
 ///
 /// typedef struct TCReplicaOpList TCReplicaOpList;
 /// ```
--- a/taskchampion/lib/src/replica.rs
+++ b/taskchampion/lib/src/replica.rs
@@ -190,15 +190,11 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
 
 impl From<TCKVList> for TaskMap {
     fn from(kvlist: TCKVList) -> TaskMap {
-        // SAFETY:
-        //  - because we have &mut self, we have read/write access to items[0..len]
-        //  - all items are properly initialized Element's
-        //  - return value lifetime is equal to &mmut self's, so access is exclusive
-        //  - items and len came from Vec, so total size is < isize::MAX
-        let slice = unsafe { std::slice::from_raw_parts(kvlist.items, kvlist.len) };
+        // TODO SAFETY:
+        let vec = unsafe { Vec::from_raw_parts(kvlist.items, kvlist.len, kvlist._capacity) };
 
         let mut taskmap = TaskMap::new();
-        slice.into_iter().for_each(|kv| {
+        vec.into_iter().for_each(|kv| {
             // SAFETY:
             //  - key is valid (promised by caller)
             //  - caller will not use key after this call (convention)
@@ -397,17 +393,18 @@ impl From<Vec<ReplicaOp>> for TCReplicaOpList {
 
 impl From<TCReplicaOpList> for Vec<ReplicaOp> {
     fn from(tc_replica_op_list: TCReplicaOpList) -> Vec<ReplicaOp> {
-        // SAFETY:
-        //  - because we have &mut self, we have read/write access to items[0..len]
-        //  - all items are properly initialized Element's
-        //  - return value lifetime is equal to &mmut self's, so access is exclusive
-        //  - items and len came from Vec, so total size is < isize::MAX
-        let tc_replica_op_slice =
-            unsafe { std::slice::from_raw_parts(tc_replica_op_list.ptr, tc_replica_op_list.len) };
+        // TODO SAFETY:
+        let tc_replica_op_vec = unsafe {
+            Vec::from_raw_parts(
+                tc_replica_op_list.ptr,
+                tc_replica_op_list.len,
+                tc_replica_op_list.capacity,
+            )
+        };
 
-        tc_replica_op_slice
-            .iter()
-            .map(|&tc_op| ReplicaOp::from(tc_op))
+        tc_replica_op_vec
+            .into_iter()
+            .map(|tc_op| ReplicaOp::from(tc_op))
             .collect()
     }
 }
--- a/taskchampion/lib/src/replica.rs
+++ b/taskchampion/lib/src/replica.rs
@@ -141,7 +141,7 @@ where
 ///     Create,
 ///     Delete,
 ///     Update,
-///     Undopoint,
+///     UndoPoint,
 /// } TCReplicaOpType;
 /// ```
 #[derive(Default)]
@@ -407,7 +407,7 @@ impl From<TCReplicaOp> for ReplicaOp {
 ///
 /// ```c
 /// struct TCReplicaOpList {
-///     void* items;
+///     struct TCReplicaOp *items;
 ///     size_t len;
 ///     size_t capacity;
 /// };
--- a/taskchampion/lib/src/replica.rs
+++ b/taskchampion/lib/src/replica.rs
@@ -616,8 +616,23 @@ pub unsafe extern "C" fn tc_replica_sync(
 pub unsafe extern "C" fn tc_replica_get_undo_ops(rep: *mut TCReplica) -> TCReplicaOpList {
     wrap(
         rep,
-        |rep| Ok(TCReplicaOpList::from(rep.get_undo_ops()?)),
-        TCReplicaOpList::from(Vec::new()),
+        |rep| {
+            // SAFETY:
+            //  - caller will free this value, either with tc_replica_commit_undo_ops or
+            //  tc_replica_op_list_free.
+            Ok(unsafe {
+                TCReplicaOpList::return_val(
+                    rep.get_undo_ops()?
+                        .into_iter()
+                        .map(TCReplicaOp::from)
+                        .collect(),
+                )
+            })
+        },
+        // SAFETY:
+        //  - caller will free this value, either with tc_replica_commit_undo_ops or
+        //  tc_replica_op_list_free.
+        unsafe { TCReplicaOpList::return_val(Vec::new()) },
     )
 }
 
@@ -640,7 +655,12 @@ pub unsafe extern "C" fn tc_replica_commit_undo_ops(
     wrap(
         rep,
         |rep| {
-            let undo_ops: Vec<ReplicaOp> = Vec::from(tc_undo_ops);
+            // SAFETY:
+            // - `tc_undo_ops` is a valid value, as it was acquired from `tc_replica_get_undo_ops`.
+            let undo_ops: Vec<ReplicaOp> = unsafe { TCReplicaOpList::val_from_arg(tc_undo_ops) }
+                .into_iter()
+                .map(|op| *op.inner)
+                .collect();
             let undone = i32::from(rep.commit_undo_ops(undo_ops)?);
             if !undone_out.is_null() {
                 // SAFETY:
@@ -796,22 +816,18 @@ pub unsafe extern "C" fn tc_replica_free(rep: *mut TCReplica) {
 /// ```
 #[no_mangle]
 pub unsafe extern "C" fn tc_replica_op_list_free(oplist: *mut TCReplicaOpList) {
+    debug_assert!(!oplist.is_null());
     // SAFETY:
-    //  - items is not NULL
-    //  - items is properly aligned (promised by caller)
-    let items = unsafe { (*oplist).items };
-    // SAFETY:
-    //  - len is not NULL
-    //  - len is properly aligned (promised by caller)
-    let len = unsafe { (*oplist).len };
-    // SAFETY:
-    //  - capacity is not NULL
-    //  - capacity is properly aligned (promised by caller)
-    let capacity = unsafe { (*oplist).capacity };
-    // SAFETY:
-    //  - items, len, and capacity are the unmodified values returned by vec_into_raw_parts
-    //  (promised by caller)
-    unsafe { Vec::from_raw_parts(items, len, capacity) };
+    // - arg is not NULL (just checked)
+    // - `*oplist` is valid (guaranteed by caller not double-freeing this value)
+    unsafe {
+        TCReplicaOpList::take_val_from_arg(
+            oplist,
+            // SAFETY:
+            //  - value is empty, so the caller need not free it.
+            TCReplicaOpList::return_val(Vec::new()),
+        )
+    };
 }
 
 #[ffizz_header::item]
--- a/taskchampion/lib/src/replica.rs
+++ b/taskchampion/lib/src/replica.rs
@@ -237,7 +237,9 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
 
 impl From<TCKVList> for TaskMap {
     fn from(kvlist: TCKVList) -> TaskMap {
-        // TODO SAFETY:
+        // SAFETY:
+        //  - items, len, and capacity are the unmodified values returned by vec_into_raw_parts
+        //  (promised by caller)
         let vec = unsafe { Vec::from_raw_parts(kvlist.items, kvlist.len, kvlist._capacity) };
 
         let mut taskmap = TaskMap::new();
@@ -365,7 +367,9 @@ impl From<Vec<ReplicaOp>> for TCReplicaOpList {
 
 impl From<TCReplicaOpList> for Vec<ReplicaOp> {
     fn from(tc_replica_op_list: TCReplicaOpList) -> Vec<ReplicaOp> {
-        // TODO SAFETY:
+        // SAFETY:
+        //  - items, len, and capacity are the unmodified values returned by vec_into_raw_parts
+        //  (promised by caller)
         let tc_replica_op_vec = unsafe {
             Vec::from_raw_parts(
                 tc_replica_op_list.items,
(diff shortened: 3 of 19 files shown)


taskchampion/taskchampion/src/taskdb/undo.rs
```diff
@@ -1,31 +1,147 @@
...
+        }
+
+        UndoDiff {
+            current,
+            prior,
+            when: when.unwrap(),
+        }
+    }
+
+    fn apply_op(task: &mut TaskMap, op: &SyncOp) -> Option<DateTime<Utc>> {
```
djmitche: A TaskMap represents a single task, while operations can apply to any task. For example, the `..` in `Delete { .. }` below may refer to a different task than the one represented by `task`. Similarly, `Update` may be an update to a different task.

djmitche (commented):
## Callbacks and Soundness

The callback approach is, in principle, a good way to hang on to the StorageTxn, and within _safe_ Rust will prevent race conditions. The race here would be that the callback invokes some other function which also acquires a StorageTxn for the same Storage instance. To do so would require an exclusive reference to the Replica, but one such exclusive reference already exists (in order to get the existing StorageTxn), so the borrow checker won't let that happen. (Really, Rust is preventing UB, which is a bit broader than race conditions)

However, C++ knows nothing of the borrow checker, so that callback could blithely call `replica->new_task(..)` or whatever and the borrow checker would be none the wiser. We would need to either represent this requirement to C++ in a way that it can enforce, or enforce it at runtime, or just write some suitably stern warnings in comments and hope everyone reads them.

As an example, [`Task::into_mut`](https://github.com/GothenburgBitFactory/taskwarrior/blob/1c11f8b650cdd15526c6694d9032b645485660fa/taskchampion/taskchampion/src/task/task.rs#L118) requires a mutable reference to the Replica, and holds that reference until the TaskMut is dropped. The [`TCTask` wrapper](https://github.com/GothenburgBitFactory/taskwarrior/blob/1c11f8b650cdd15526c6694d9032b645485660fa/taskchampion/lib/src/task.rs#L55) wraps this into a single class, and the [C++ `ReplicaGuard`](https://github.com/GothenburgBitFactory/taskwarrior/blob/1c11f8b650cdd15526c6694d9032b645485660fa/src/tc/Replica.h#L51) class sets the replica pointer to `nullptr` for the duration if that mutable borrow, thereby enforcing the soundness requirement at runtime.

tl;dr: it's a lot easier to not have Rust references with a lifetime that includes any C++ execution.

## Concurrency and Operations

AIUI the reason to want to hang onto the transaction is so that nothing changes between the time the confirmation prompt is shown and the time the user replies. I don't actually think that's necessary, due to the nature of the data we're handling: most operations commute with no conflicts.

Instead, I think you could pass a new StorageTxn to `commit`, and it can simply commit the reversed operations. If those happen to conflict with something that happened in the interim, that's OK -- the whole operational-transform system is built to handle that. But, it probably won't conflict. For example, maybe I accidentally modified the description on the wrong task T1, and while I'm confirming the undo, a bugwarrior sync adds T2 and modifies the description for T3. Those operations will trivially commute without conflict. By analogy to Git, this would be `git revert` -- applying a new patch to undo the effects of an existing patch, even if other patches have been committed in the interim.

There's a tricky bit, though: the existing undo support just "pops" some operations from the local operations vector, and updates the state of the tasks to match (`apply::apply_op`), with no possibility of conflict. This is more like `git reset --hard HEAD^`, in that the history of those operations is deleted. And another `task undo` can then pop some more operations.

A further wrinkle is that if a `task sync` occurs while the transaction is being undone, that sync "commits" all of the existing operations, and they can no longer be popped. The analogy would be `git push` to a repo that doesn't allow force-pushes -- once you've pushed a commit, you can't edit it. So after sync, changes would have to be undone with the `git revert` approach, pushing the reversed operations into the list of local ops and eventually syncing those reversed ops to the server.

I believe the 2.6.0 `sync` and `undo` have some unsound interactions, and while I don't want to do unsound things, I think it's OK to fail with an error in cases that used to be unsound. Further, I think `task undo` is mostly used for those "oops" moments of typing a command wrong -- that is, almost immediately. So, I think a solution to the tricky bits and wrinkles above that constrains the generality of `task undo` is OK, as long as it hits the major use-cases. Maybe that means only one level of undo? Maybe it means that a `task undo` which can't simply pop the local operations _works_, but that afterward another `task undo` is not supported? Maybe something else?

## Generalizing Operations

#2928 is a little more general than this PR, too, and it may make sense to think forward to other uses of operations. In particular,
 - `task info` used to display the changes to that particular task. That information as basically a human-readable translation of a set of operations for that task, so it probably makes sense to have some kind of `get_all_operations_for_task(..)` method. That would mean that operations are a type represented to C++ via FFI.
 - Depending on config, `task modify` and friends will show a friendly diff of what they are about to change before making that change. We had talked in #3058 about using operations for that support as well: `task modify` could parse its command line and generate a list of operations. Then, if configured, it can show a human-readable version of that list and ask for confirmation, and finally "commit" that list of operations. This is very similar to undo: prepare  a list of operations, confirm them, and then commit them.

So maybe that's a better API to expose from taskchampion: lists of operations, and a way to "commit" such a list.

## FFI

The FFI situation isn't the best -- it's very much _not_ automated. I've learned a little bit more since I did this, and might solve it differently now, but for purposes of getting 3.0 shipped let's stick to what we have. So, an operation might be modeled on [`TCUda`](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/taskchampion/lib/src/uda.rs), and a sequence of operations (i.e., UndoDiff) on `TCUdaList`. That serialization should happen in `taskchampion-lib` (under `taskchampion/lib/src`) and not in `taskchampion` (under `taskchampion/taskchampion`).

---

Sorry, that's a lot! I think this is a great start and I'm interested to see where it develops. I like the approach of starting with the Rust implementation, and testing it in Rust, before moving on to wrap it in C++ and integrate with Taskwarrior.

ryneeverett:
> There's a tricky bit, though: the existing undo support just "pops" some operations from the local operations vector

How would you suggest to resolve this? It seems like there are three possibilities:

1. Persistent transaction. (My current approach which we agree is not ideal.)
2. Don't remove operations, just append reversal operations. This would require changing `Replica.reverse_ops` from returning `Vec<SyncOp>` to returning `Vec<ReplicaOp>` and calling `op.into_sync()` before applying. I think `ReplicaOp` would have to gain some consciousness of whether it is itself an undo operation or otherwise subsequent undos would actually be redos ad infinitum. Maybe just add `ReplicaOp::RedoPoint`? This would also open the door to the possibility of `task redo` some day.
3. Remove operations. Add `StorageTxn.remove_operation(&mut self, op: ReplicaOp)` and then try to remove by value? Or add a `primary_key` field to each `ReplicaOp` variant?

Now that I bothered to articulate the options (2) seems like the less messy choice but maybe you had something else in mind entirely. It also occurs to me that there may be an established solution to this sort of problem in OT theory.

djmitche:
From a user perspective, I think the best approach is to try to support undo as much as possible, and no more. I think that means trying to "pop" operations when that's possible, and falling back to applying reverse operations when that's not possible.

Operations can be popped when they have not been sync'd, and when no further local operations have been added, such as while waiting for user confirmation. Otherwise, reversed operations can be applied with similar user-visible effect. After application of reversed operations, or after a sync, I think it's fine to prohibit further undo operations: from a user perspective, it's a little ambiguous if they are undoing the undo, or what it means to undo a sync. We could probably relax the last sentence a little in the future, but for the moment I think it's an acceptable compromise.

In practical terms, my suggestion is to "commit" an undo by first trying to "pop" operations from the local operation list, provided they match precisely. If that fails because the operations do not match, then fall back to applying those in reverse.

That's pretty complicated, though! I would also be OK with only supporting pop, and just failing with a nice error message otherwise. I think that would cover the "oops" use-case well, especially for the majority of users who do not have a lot of automation modifying their tasks. We could also implement the apply-in-reverse part in the future.

---

As a side-note, apply-in-reverse will sometimes do potentially confusing things. Consider, for example:

 * Begin with T1.description = "blueberries"
 * Update(T1, "description", "carrots")
 * Update(T1, "description", "cabbage")

Now the user wants to undo the first update. As a ReplicaOp, that Update stored its previous value, too: Update(T1, "description", "carrots", old_value: "blueberries"). So, when it is reversed and applied as:

 * Update(T1, "description", "blueberries")

while the user might reasonably expect the final description to be "cabbage". This doesn't affect the soundness of the OT model, but it diminishes the attractiveness of the apply-in-reverse approach.

---

Having written all that, I think we should opt for the simplest approach right now: provide a method that will "undo" a sequence of operations if and only if those are precisely the most recent local operations in the replica. If that proves insufficient, we can later implement the apply-in-reverse functionality.

ryneeverett:
I'm a little conflicted about your position on this. If we're not dedicated to supporting "concurrent undo" by reversing operations then why isn't a single persistent transaction via callback a viable option? That does seem like "the simplest approach right now" which is probably why you did it that way in the first place. I'm not sure the creation of a more general operations api justifies a more complicated undo implementation.

---

> As a side-note, apply-in-reverse will sometimes do potentially confusing things.

Following your git analogy, I think we could have a notion of "conflicts" in which if any operation prior to the last UndoPoint deletes a task which has been concurrently updated or updates a property which had concurrently been updated the undo fails but otherwise succeeds. Further development could even introduce "interactive rebase" via "picking"/"dropping"/"editing" certain operations.

djmitche:
Holding the transaction over a callback is technically risky, but we could work around that -- maybe it's simple enough to just include some really restrictive requirements in a comment: don't reference any other taskchampion values from within this callback. This particular use-case is just to pretty-print the callback argument and get a y/n answer from the user, so that should be easy.

I think we'd still want some pretty generic FFI that will cover the `task info` case as well as `task undo`, just because it's so much work to build an FFI type.

ryneeverett:
@djmitche I think this is fairly close to what you were suggesting. I'd be interested in more guidance when you get a chance.

As far as the implementation of `task undo` goes:

1. I'm currently stuck on "Can't undo non-operation." (An error thrown if an invalid operation type is passed through ffi.)
2. If I don't throw an error but instead assign `opstr = "Non-Operation";` it turns out that *none* of the operations match the TCReplicaOpType variants.
3. If I confirm the undo I get a segfault. I attempted to debug this with gdb and rust-gdb but debug symbols don't seem to be available (`0x0000000000000000 in ?? ()`) despite `cmake -DCMAKE_BUILD_TYPE=debug` and `cargo build`ing without the `--release` flag.

Any advice on overcoming any of these obstacles would be appreciated.

taskchampion/lib/src/replica.rs
```diff
@@ -130,6 +132,30 @@ where
...
+/// ```c
+/// typedef enum TCReplicaOpType {
+///     Create,
+///     Delete,
+///     Update,
+///     UndoPoint,
+/// } TCReplicaOpType;
+/// ```
+#[derive(Default)]
+pub enum TCReplicaOpType {
```
djmitche: I think this is the issue -- without any instruction, Rust is free to handle this enum however it would like, which is probably not the same way the C compiler will interpret the typedef above. Take a look at [result.rs](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/taskchampion/lib/src/result.rs), and in particular the `repr(i32)` (in this case you could choose `repr(u32)`), explicit values for the variants, and a bunch of #ifdef stuff so that C++ chooses the same type.

taskchampion/lib/src/replica.rs
```diff
@@ -130,6 +132,30 @@ where
...
+    Create,
+    Delete,
+    Update,
+    UndoPoint,
+    #[default]
+    Error,
+}
+
+impl PassByPointer for TCReplicaOpType {}
```
djmitche: I suspect you'll want PassByValue here, since this is a simple copy-able type.

taskchampion/lib/src/replica.rs
```diff
@@ -186,6 +212,268 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
...
+///     TCString property;
+///     TCString old_value;
+///     TCString value;
+///     TCString timestamp;
+/// };
+///
+/// typedef struct TCReplicaOp TCReplicaOp;
+/// ```
+#[derive(Default)]
+pub struct TCReplicaOp {
```
djmitche, line 234: This is a similar issue -- this needs to be `#[repr(C)]` so that Rust will lay the struct out like C would. Otherwise, Rust gets to arrange it however it chooses, which is why reading `operation_type` from C++ is returning garbage.

djmitche (commented):
I don't have any good advice about debugging or debug symbols -- I suspect there's a way to do that via CMake, but I don't know what it is.

Sending all of TCReplicaOp on a round-trip to C structures and back is a lot of work -- although you have done most of it now! It might be easier to treat TCReplicaOp as an opaque struct from the C side, which just contains a pointer to the ReplicaOp (so `Box<ReplicaOp>` on the Rust side). You can add functions to get the necessary pieces of data as needed (`tc_replica_op_get_type`, `tc_replica_op_get_property`, etc.), leaving the Rust ReplicaOp struct intact. Then when the TCReplicaOp gets passed back to `tc_replica_commit_undo_ops` all that method has to do is unbox it to get a ReplicaOp -- instead of converting all of those fields back from the C/C++ types.

ryneeverett:
I'm very close to having the first leg of the trip done now! The first operation in the undo batch is still garbage, but the others are coming through. Here's some output I'm currently getting:

<details>

```
$ ./src/task undo
[]
[]
[
    Update {
        uuid: a9218e0f-3d2f-4cd0-bd8b-a18d6828e0b7,
        property: "status",
        old_value: None,
        value: Some(
            "pending",
        ),
        timestamp: 2024-01-05T01:23:26.886007341Z,
    },
    Update {
        uuid: a9218e0f-3d2f-4cd0-bd8b-a18d6828e0b7,
        property: "modified",
        old_value: None,
        value: Some(
            "1704417806",
        ),
        timestamp: 2024-01-05T01:23:26.867950428Z,
    },
    Update {
        uuid: a9218e0f-3d2f-4cd0-bd8b-a18d6828e0b7,
        property: "entry",
        old_value: None,
        value: Some(
            "1704417806",
        ),
        timestamp: 2024-01-05T01:23:26.858236832Z,
    },
    Update {
        uuid: a9218e0f-3d2f-4cd0-bd8b-a18d6828e0b7,
        property: "description",
        old_value: None,
        value: Some(
            "foo",
        ),
        timestamp: 2024-01-05T01:23:26.848818290Z,
    },
    Create {
        uuid: a9218e0f-3d2f-4cd0-bd8b-a18d6828e0b7,
    },
]
[
    TCReplicaOp {
        operation_type: Update,
        uuid: TCUuid(
            [
                169,
                33,
                142,
                15,
                61,
                47,
                76,
                208,
                189,
                139,
                161,
                141,
                104,
                40,
                224,
                183,
            ],
        ),
        old_task: TCKVList {
            len: 0,
            _capacity: 0,
            items: 0x0000000000000000,
        },
        property: TCString {
            ptr: 0x000000002dc38e80,
            len: 6,
            cap: 6,
            ty: 3,
        },
        old_value: TCString {
            ptr: 0x0000000000000001,
            len: 0,
            cap: 0,
            ty: 3,
        },
        value: TCString {
            ptr: 0x000000002dc390e0,
            len: 7,
            cap: 7,
            ty: 3,
        },
        timestamp: TCString {
            ptr: 0x000000002dc3ed50,
            len: 33,
            cap: 64,
            ty: 3,
        },
    },
    TCReplicaOp {
        operation_type: Update,
        uuid: TCUuid(
            [
                169,
                33,
                142,
                15,
                61,
                47,
                76,
                208,
                189,
                139,
                161,
                141,
                104,
                40,
                224,
                183,
            ],
        ),
        old_task: TCKVList {
            len: 0,
            _capacity: 0,
            items: 0x0000000000000000,
        },
        property: TCString {
            ptr: 0x000000002dc38ce0,
            len: 8,
            cap: 8,
            ty: 3,
        },
        old_value: TCString {
            ptr: 0x0000000000000001,
            len: 0,
            cap: 0,
            ty: 3,
        },
        value: TCString {
            ptr: 0x000000002dc38d70,
            len: 10,
            cap: 10,
            ty: 3,
        },
        timestamp: TCString {
            ptr: 0x000000002dc3eda0,
            len: 33,
            cap: 64,
            ty: 3,
        },
    },
    TCReplicaOp {
        operation_type: Update,
        uuid: TCUuid(
            [
                169,
                33,
                142,
                15,
                61,
                47,
                76,
                208,
                189,
                139,
                161,
                141,
                104,
                40,
                224,
                183,
            ],
        ),
        old_task: TCKVList {
            len: 0,
            _capacity: 0,
            items: 0x0000000000000000,
        },
        property: TCString {
            ptr: 0x000000002dc38bc0,
            len: 5,
            cap: 5,
            ty: 3,
        },
        old_value: TCString {
            ptr: 0x0000000000000001,
            len: 0,
            cap: 0,
            ty: 3,
        },
        value: TCString {
            ptr: 0x000000002dc38c50,
            len: 10,
            cap: 10,
            ty: 3,
        },
        timestamp: TCString {
            ptr: 0x000000002dc3edf0,
            len: 33,
            cap: 64,
            ty: 3,
        },
    },
    TCReplicaOp {
        operation_type: Update,
        uuid: TCUuid(
            [
                169,
                33,
                142,
                15,
                61,
                47,
                76,
                208,
                189,
                139,
                161,
                141,
                104,
                40,
                224,
                183,
            ],
        ),
        old_task: TCKVList {
            len: 0,
            _capacity: 0,
            items: 0x0000000000000000,
        },
        property: TCString {
            ptr: 0x000000002dc3a7a0,
            len: 11,
            cap: 11,
            ty: 3,
        },
        old_value: TCString {
            ptr: 0x0000000000000001,
            len: 0,
            cap: 0,
            ty: 3,
        },
        value: TCString {
            ptr: 0x000000002dc3ab10,
            len: 3,
            cap: 3,
            ty: 3,
        },
        timestamp: TCString {
            ptr: 0x000000002dc3ee40,
            len: 33,
            cap: 64,
            ty: 3,
        },
    },
    TCReplicaOp {
        operation_type: Create,
        uuid: TCUuid(
            [
                169,
                33,
                142,
                15,
                61,
                47,
                76,
                208,
                189,
                139,
                161,
                141,
                104,
                40,
                224,
                183,
            ],
        ),
        old_task: TCKVList {
            len: 0,
            _capacity: 0,
            items: 0x0000000000000000,
        },
        property: TCString {
            ptr: 0x0000000000000000,
            len: 0,
            cap: 0,
            ty: 0,
        },
        old_value: TCString {
            ptr: 0x0000000000000000,
            len: 0,
            cap: 0,
            ty: 0,
        },
        value: TCString {
            ptr: 0x0000000000000000,
            len: 0,
            cap: 0,
            ty: 0,
        },
        timestamp: TCString {
            ptr: 0x0000000000000000,
            len: 0,
            cap: 0,
            ty: 0,
        },
    },
]
The following 5 operations would be reverted:
- NON-OPERATION: 187454
- Update
- Update
- Update
- Create
The undo command is not reversible.  Are you sure you want to revert to the previous state? (yes/no)
```

</details>

---

> Sending all of TCReplicaOp on a round-trip to C structures and back is a lot of work -- although you have done most of it now! It might be easier to treat TCReplicaOp as an opaque struct from the C side, which just contains a pointer to the ReplicaOp (so Box<ReplicaOp> on the Rust side). 

My thinking is that while the roundtrip is wasteful, the roundtrip in undo is the exception and will only be very small batches of operations. Whereas any other use of `TCReplicaOp` will probably be larger batches of operations only sent one way (e.g. `task info`). Maybe a compromise would be to forget the return trip and send both the structs and the pointers for undo and free the structs before returning the pointer on the C-side:

```rs
pub struct TCUndoOpsList {
    ptr: Box<Vec<ReplicaOp>>;
    oplist: TCReplicaOpList;
}
```

My premise is that C structs seem easier to work with on the C-side and potentially less of a messy ffi implementation than 7 functions for each field. Maybe you disagree.

---

I'm having a bad time getting taskwarrior to link against the newest taskchampion build when I make a modification. The only thing I've figured out so far is to move the entire taskwarrior directory every time I modify taskchampion and `rm CMakeCache.txt && cargo build && cargo xtask codegen && cmake -DCMAKE_BUILD_TYPE=debug . && make VERBOSE=1 -j4`. Any ideas?

djmitche:
> My premise is that C structs seem easier to work with on the C-side and potentially less of a messy ffi implementation than 7 functions for each field. Maybe you disagree.

I think this is true when they are simple C structs, but when they are strings then you already need the `tc_string_*` methods to access the data on the C side, and there's all the unsafe conversions to do on the Rust side. But I definitely don't know what's "best" or "common" in this area!

> I'm having a bad time getting taskwarrior to link against the newest taskchampion build when I make a modification.

Yeah, this is GothenburgBitFactory/taskwarrior#2887. I have this in my editor's config to force a run of `cargo build`:
```
rm -rf src/tc/rust/x86_64-unknown-linux-gnu/debug/libtc_rust.a                                                                                                                                                                                                                                                       
```

taskchampion/lib/src/replica.rs
```diff
@@ -186,6 +238,272 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
...
+    fn from(replica_op_list: Vec<ReplicaOp>) -> TCReplicaOpList {
+        // XXX Remove prints
+        println!("{:#?}", replica_op_list);
+        let tc_replica_op_list: Vec<TCReplicaOp> = replica_op_list.into_iter().map(|op| TCReplicaOp::from(op)).collect();
+        println!("{:#?}", tc_replica_op_list);
+        TCReplicaOpList {
+            items: tc_replica_op_list.as_ptr() as *mut TCReplicaOp,
+            len: tc_replica_op_list.len(),
+            capacity: tc_replica_op_list.capacity(),
+        }
```
djmitche: I think the reason you're getting garbage is that this is creating a new Vec, then embedding a pointer to it in the return value, then freeing the Vec on return from this method. So the resulting TCReplicaOpList is pointing to unallocated memory!

To achieve this, I think you'll need to make the TCReplicaOpList "own" the Vec. You can do that with [`Vec::into_raw_parts`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_raw_parts), which both returns the raw parts of the Vec and also "forgets" the original Vec so that it won't be freed on return. You'd need to pair that with `Vec::from_raw_parts` in `tc_replica_op_list_free`, to ensure that memory was freed.

djmitche:
This is kind of reinforcing for me that this isn't a great way to implement FFI. I think we should continue to muddle through this, but hopefully we can find or invent a better (and safer) way eventually.

taskchampion/lib/src/replica.rs
ryneeverett: That was it!

However, can we really use `Vec::into_raw_parts` when it is a nightly/experimental API? I tried using `Box::leak` instead but I'm not confident that's right. I got a `free(): invalid pointer` error when attempting to free it.

djmitche: Sorry, I forgot about that! There is an equivalent utility function in taskchampion-lib somewhere (vec_into_raw_parts?)

ryneeverett:
Some preliminary thoughts on getting this ready to merge:

1. A lot of stuff is getting crammed in replica.rs. Should I pull the ReplicaOp-related stuff into a separate file? Or should replica.rs be broken up into a folder?
2. The tests appear to be failing only on older distributions. Any ideas?
3. The (numerous duplicate) clippy warnings in the github diff appear to be accurate, but I can't reproduce them locally and the clippy CI check is green. What's going on here?

djmitche:
> Some preliminary thoughts on getting this ready to merge:
> 
>     1. A lot of stuff is getting crammed in replica.rs. Should I pull the ReplicaOp-related stuff into a separate file? Or should replica.rs be broken up into a folder?

I think it's OK to leave that as-is and fix in a followup. I'd suggest keeping the hierarchy flat and moving replica-op stuff to another file. The resulting C API is flat anyway, so splitting into folders doesn't do much.

>     2. The tests appear to be failing only on older distributions. Any ideas?

```
#        free(): invalid size
```
that looks like some error with allocations. I wonder if the older Linuxes just have a different libc that's somehow not masking the error?

>     3. The (numerous duplicate) clippy warnings in the github diff appear to be accurate, but I can't reproduce them locally and the clippy CI check is green. What's going on here?

I wonder if the duplicates are just piling up from multiple pushes?

There are weird things with `cargo check` and `cargo clippy` that I haven't entirely figured out. I think incremental compilation leads to checks "disappearing" if that particular code unit doesn't get recompiled on the next run of `cargo check`. And `cargo clippy` seems to want to check all dependencies, not just the crate you're building. But in this case, I can reproduce locally:

```
⸩ cargo check -p taskchampion-lib 2>&1 | head -n10
warning: unused import: `atomic::*`
   --> taskchampion/lib/src/lib.rs:133:9
    |
133 | pub use atomic::*;
    |         ^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

warning: dereference of raw pointer is unsafe and requires unsafe block (error E0133)
   --> taskchampion/lib/src/replica.rs:795:25
```
and re-running that still shows the warnings. The GitHub action does run `cargo clippy` from the MSRV (1.65), but I don't think that matters in this case.

taskchampion/lib/src/kv.rs
```diff
@@ -100,6 +120,16 @@ impl CList for TCKVList {
     }
 }
 
+impl Default for TCKVList {
+    fn default() -> Self {
+        Self {
+            len: libc::size_t::default(),
+            _capacity: libc::size_t::default(),
+            items: null_mut(),
+        }
```
djmitche: This is one of the really ugly corner-cases for Rust / C/C++ interop: Rust doesn't necessarily represent an empty vector as (0, 0, NULL), and uses a different semaphore value in the pointer field. I think a better approach here would be something like `vec_into_raw_parts(Vec::default())`. This _might_ be the cause of the free errors you're seeing.

djmitche: https://github.com/rust-lang/rust/issues/39625 seems to be the best info I can find on this topic (it's also been discussed quite a lot internally at work, which is why it comes to mind)

djmitche:
BTW, here's a longer explanation of the empty-slice problem - https://davidben.net/2024/01/15/empty-slices.html

taskchampion/lib/src/kv.rs
ryneeverett: Well that wasn't it. Turns out this implementation wasn't even being used. I deleted it.

ryneeverett:
After rebasing so that CI will run on the latest distros, I'm observing the following:

- Fedora38/Fedora39: `test_undo_diff_operations` is failing because 5 operations are found rather than the expected 4. The operation set is including a changed `modified` property. I could change the test to accommodate this but any idea why it's happening? Curiously, when I run the tests in a fedora39 container locally I do not get this error. But why would the kernel make a difference in the operations?
- Ubuntu20.04/OpenSUSE: SIGABRT timeout on `task undo`

djmitche:
>     * Fedora38/Fedora39: `test_undo_diff_operations` is failing because 5 operations are found rather than the expected 4. The operation set is including a changed `modified` property. I could change the test to accommodate this but any idea why it's happening? Curiously, when I run the tests in a fedora39 container locally I do not get this error. But why would the kernel make a difference in the operations?

This might be intermittent. Related to GothenburgBitFactory/taskchampion#372, Taskwarrior skips over all of the `set_priority`, `add_tag` etc. methods and just calls `set_value`, except for status (`set_status`) and modified (`set_modified`) in `TDB2::add`.  `TDB2::modify` only uses `set_value`, and does so by comparing the new values (in the Task's `std::map`) to the existing values (in the TCTask). If those differ, it calls `set_value`. I suspect what's happening is that sometimes the `task add` and `task modify` are in the same second, so the `modified` property doesn't get updated. But sometimes they're in different seconds, so it does get updated.

I think this is a case for relaxing the test a little, so that it is happy whether `modified` appears or not -- with a comment explaining why.

>     * Ubuntu20.04/OpenSUSE: SIGABRT timeout on `task undo`

The logs say SIGSEGV, actually. The SIGABRT stuff is just a wrapper around the error message basically saying "<-- if that said SIGABRT then ..".

Those are some older systems, so might have an older malloc or something with a different memory behavior that may be tickling some memory unsafety. I can get a SIGSEGV locally by typing "n" after the undo confirmation prompt, and that occurs while trying to drop a ReplicaOp -- so perhaps those are being double-freed?

ryneeverett:
Does https://github.com/GothenburgBitFactory/taskwarrior/pull/3213/commits/6755ff68e6c209bd1e8be70540751906d5bafd56 give any hint at the underlying issue?

djmitche:
Yikes, UB is wild! I don't know why that would "fix" things other than somehow perturbing the memory layout or allocations or preventing an optimization or something.

```c++
void TDB2::revert ()
{
  auto undo_ops = replica.get_undo_ops();
  if (undo_ops.len == 0) {
    std::cout << "No operations to undo.";
    return;
  }
  if (confirm_revert(undo_ops)) {
    replica.commit_undo_ops(undo_ops, NULL);
  }
  replica.free_replica_ops(undo_ops);
  replica.rebuild_working_set (false);
}
```
I think the issue is here. `undo_ops` is passed around by value, but must maintain a single owner. The `confirm_revert` call doesn't take ownership, but `replica.commit_undo_ops` does, by calling `tc_replica_commit_undo_ops`, which uses `Vec::from(tc_undo_ops)`.  The `Replica::commit_undo_ops` method further takes ownership, and eventually drops the Rust value when it finishes. So the call to `replica.free_replica_ops(undo_ops)` is then double-freeing the list of ops.

ryneeverett:
Is it possible that you've identified a separate problem? The SIGSEGVs we were seeing in test were with negative input to `confirm_revert` so `replica.commit_undo_ops` was never executed.

djmitche:
Oh, that's definitely possible.

djmitche:
```rust
pub unsafe extern "C" fn tc_replica_op_get_property(op: &TCReplicaOp) -> TCString {
    if let ReplicaOp::Update { property, .. } = *op.inner.clone() {
        // SAFETY:
        //  - caller promises to free this string
        unsafe { TCString::return_val(property.into()) }
    } else {
        panic!("Operation has no property: {:#?}", op);
    }
}
```
First, C doesn't really understand references so that argument should be `op: *const TCReplicaOp` rather than a reference. I'm honestly not sure if there's ever a use for using references in an `extern "C"` declaration, but I suspect it just leads Rust to make possibly-unfounded assumptions about the argument (such as that it's non-NULL and points to valid memory that is not aliased and will not be modified during the lifetime of the reference). That `*op.inner.clone()` had me scratching my head and led me to learn something about [how `Box` is "special" to the compiler](https://stackoverflow.com/questions/33653946/dereferencing-boxt-gives-back-value-instead-of-reference). That's parsed as `*(op.inner.clone())`, the inner part of which creates a new `Box` with a cloned `ReplicaOp` in it. The `*` then unboxes this, resulting in an owned `ReplicaOp`, which is then matched against the `if let` pattern. The resulting `property` is an owned `String`, which is converted to a `RustString` and returned. This looks like it has an extra allocation but isn't actually UB. Anyway, I learned some things, so I wrote them down. I think a better way to do this would be `if let ... = op.inner.as_ref() { .. }`, where `as_ref()` creates a `&ReplicaOp` from the `Box<ReplicaOp>`, which can then be matched against, resulting in `property` having type `&String`. Then just clone `property`.

I commented out the call to `confirm_revert` to just return false, and the error persisted. GDB says the error occurs while dropping a ReplicaOp, which suggests a double-free. The `tc_replica_op_list_free` looks OK, so I had a look at `tc_replica_get_undo_ops`, which calls `TCReplicaOpList::from(Vec<ReplicaOp>)`, which builds a vector and converts it to its raw parts. Looks good so far.

Ah!
```rust
/// EXTERN_C void tc_replica_op_list_free(struct TCReplicaOpList oplist);
/// ```
#[no_mangle]
pub unsafe extern "C" fn tc_replica_op_list_free(oplist: *mut TCReplicaOpList) {
```
The C declaration says the oplist is passed by value, while the Rust declaration is taking it by pointer. I don't know the C ABI well enough to know how those look in terms of registers or stack, but `*mut TCReplicaOpList` is 8 bytes and `struct TCReplicaOpList` is 24 bytes so there are definitely bytes going where they shouldn't. So:
```diff
diff --git src/tc/Replica.cpp src/tc/Replica.cpp
index a23a6086b..2cc0765bd 100644
--- src/tc/Replica.cpp
+++ src/tc/Replica.cpp
@@ -167,18 +167,17 @@ void tc::Replica::commit_undo_ops (TCReplicaOpList tc_undo_ops, int32_t *undone_
   if (res != TC_RESULT_OK) {
     throw replica_error ();
   }
 }
 
 ////////////////////////////////////////////////////////////////////////////////
 void tc::Replica::free_replica_ops (TCReplicaOpList tc_undo_ops)
 {
-  std::cout << "FIXME: This avoids a SIGSEGV. Why?";
-  tc_replica_op_list_free(tc_undo_ops);
+  tc_replica_op_list_free(&tc_undo_ops);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
 std::string tc::Replica::get_op_uuid(TCReplicaOp &tc_replica_op) const
 {
   TCString uuid = tc_replica_op_get_uuid(&tc_replica_op);
   return tc2string(uuid);
 }
diff --git taskchampion/lib/src/replica.rs taskchampion/lib/src/replica.rs
index 5deadd268..1a860b566 100644
--- taskchampion/lib/src/replica.rs
+++ taskchampion/lib/src/replica.rs
@@ -787,17 +787,17 @@ pub unsafe extern "C" fn tc_replica_free(rep: *mut TCReplica) {
 }
 
 #[ffizz_header::item]
 #[ffizz(order = 903)]
 /// Free a vector of ReplicaOp.  The vector may not be used after this function returns and must not be freed
 /// more than once.
 ///
 /// ```c
-/// EXTERN_C void tc_replica_op_list_free(struct TCReplicaOpList oplist);
+/// EXTERN_C void tc_replica_op_list_free(struct TCReplicaOpList *oplist);
 /// ```
 #[no_mangle]
 pub unsafe extern "C" fn tc_replica_op_list_free(oplist: *mut TCReplicaOpList) {
     // SAFETY:
     //  - items is not NULL
     //  - items is properly aligned (promised by caller)
     let items = unsafe { (*oplist).items };
     // SAFETY:
```

Sorry for the stream-of-consciousness comment! This was hard to find. In general, this kind of interfacing is really hard! I've tried to make it a bit easier with `ffizz` (the ffizz_passby types would at least help avoid re-inventing value-passing semantics for each new type), but not yet used those in Taskchampion. I haven't been able to make it _much_ easier. There has got to be a better way -- I'm just not sure what it is.

ryneeverett:
> Sorry for the stream-of-consciousness comment! 

I don't have the energy to give this the proper attention this evening but no need to apologize -- I'm sorry for taking on a project that I'm not really competent to complete without hand-holding! I was intentionally prodding you for a thought-dump because I was out of ideas and your thought dumps have thusfar proven helpful and worth reading and re-reading 100% of the time.

ryneeverett:
The changes you suggested make sense but did not fix the SIGSEGV around `tc_replica_op_list_free` because I was getting the same error on https://github.com/GothenburgBitFactory/taskwarrior/pull/3213/commits/d1770fc5c6015aca5331d2267c4f1b0ffd647fe1. However, I was then also running into the double-free issue you'd identified above and fixed it with https://github.com/GothenburgBitFactory/taskwarrior/pull/3213/commits/050ec18b492744dd43a04a2e08a3f416404e3ac8, which incidentally provided another UB "fix" for the `tc_replica_op_list_free` issue.

djmitche:
I see a lot of green! I will do this review soon.

ryneeverett:
Thinking about this a bit more, I think the thing I find most disturbing about these UB issues is that they reveal that ffizz is somewhat violating rust's promise that everything outside an `unsafe` block is safe. It seems like passby might fix this.

djmitche:
Rust doesn't quite say that everything outside an `unsafe` block is safe. Rather, inside an unsafe block it is the programmer's responsibility to make sure that safe Rust can't cause any UB. It's a subtle difference, but does make reviews for things like cargo-vet much more difficult, because you can't _just_ look at the `unsafe` blocks, but also must ensure that any invariants they rely on are also ensured.

Anyway, yes, ffizz-passby was an attempt to at least make this easier for developers. But I'm not convinced that's [meaning using ffizz-passby in TaskChampion] a good idea!

taskchampion/taskchampion/src/lib.rs
```diff
@@ -63,7 +63,7 @@ mod replica;
 pub mod server;
 pub mod storage;
 mod task;
-mod taskdb;
```
djmitche: I don't see `taskdb` referenced elsewhere, so I don't think this is necessary.

taskchampion/taskchampion/src/replica.rs
```diff
@@ -236,10 +236,15 @@ impl Replica {
         Ok(())
     }
 
-    /// Undo local operations until the most recent UndoPoint, returning false if there are no
+    /// Return undo local operations until the most recent UndoPoint, returning false if there are no
     /// local operations to undo.
```
djmitche:
```suggestion
    /// Return undo local operations until the most recent UndoPoint, returning an empty Vec if there are no
    /// local operations to undo.
```

taskchampion/taskchampion/src/taskdb/mod.rs
```diff
@@ -114,11 +114,17 @@ impl TaskDb {
         sync::sync(server, txn.as_mut(), avoid_snapshots)
     }
 
-    /// Undo local operations until the most recent UndoPoint, returning false if there are no
+    /// Return undo local operations until the most recent UndoPoint, returning false if there are no
```
djmitche:
```suggestion
    /// Return undo local operations until the most recent UndoPoint, returning an empty Vec if there are no
```

taskchampion/taskchampion/src/replica.rs
```diff
@@ -236,10 +236,15 @@ impl Replica {
...
-    pub fn undo(&mut self) -> Result<bool> {
-        self.taskdb.undo()
+    pub fn get_undo_ops(&mut self) -> Result<Vec<ReplicaOp>> {
+        self.taskdb.get_undo_ops()
+    }
+
+    /// Undo local operations in storage.
```
djmitche: Perhaps comment on what the return value means?

taskchampion/taskchampion/src/taskdb/undo.rs
```diff
@@ -1,19 +1,58 @@
...
-/// Undo local operations until an UndoPoint.
-pub(super) fn undo(txn: &mut dyn StorageTxn) -> Result<bool> {
-    let mut applied = false;
-    let mut popped = false;
-    let mut local_ops = txn.operations()?;
+// Adapted from https://stackoverflow.com/a/67105052/1938621
+#[cfg(not(test))]
+use log::{debug, info, trace}; // Use log crate when building application
+#[cfg(test)]
+use std::{println as debug, println as info, println as trace}; // Workaround to use prinltn! for logs.
```
djmitche: I think just setting `RUST_LOG=trace` when running the tests is sufficient to accomplish this.

src/tc/Replica.h
```diff
@@ -93,7 +93,15 @@ namespace tc {
     tc::Task import_task_with_uuid (const std::string &uuid);
 // TODO: struct TCTask *tc_replica_import_task_with_uuid(struct TCReplica *rep, struct TCUuid tcuuid);
     void sync(Server server, bool avoid_snapshots);
-    void undo (int32_t *undone_out);
+    tc::ffi::TCReplicaOpList get_undo_ops ();
+    void commit_undo_ops (tc::ffi::TCReplicaOpList tc_undo_ops, int32_t *undone_out);
+    void free_replica_ops (tc::ffi::TCReplicaOpList tc_undo_ops);
```
djmitche, line 98: It'd be nice to handle this freeing with a C++ wrapper, so that e.g., freeing on exception is handled correctly.

I think you've suffered enough for now, so no need to fix this now. And, let's wait until the 3.0 release and then maybe consider a much more substantial overhaul of the FFI, since the current implementation is clearly too complex!

src/tc/Replica.h
```diff
@@ -93,7 +93,15 @@ namespace tc {
...
+    std::string get_op_uuid(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_property(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_value(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_old_value(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_timestamp(tc::ffi::TCReplicaOp &tc_replica_op) const;
+    std::string get_op_old_task_description(tc::ffi::TCReplicaOp &tc_replica_op) const;
```
djmitche, line 104: It'd also be nice for these to be methods on a C++ ReplicaOp type. But again, don't fix in this PR.

taskchampion/lib/src/kv.rs
```diff
@@ -100,6 +120,17 @@ impl CList for TCKVList {
...
 
+impl Default for TCKVList {
+    fn default() -> Self {
+        let (ptr, len, capacity) = vec_into_raw_parts(Vec::default());
+        Self {
+            items: ptr,
+            len,
+            _capacity: capacity,
+        }
+    }
```
djmitche: Preferring to use the PassByValue types:
```suggestion
        // SAFETY:
        //  - caller will free this list
        unsafe { TCKVList::return_val(Vec::new()) }
```

taskchampion/lib/src/replica.rs
```diff
@@ -130,6 +131,54 @@ where
...
+    fn as_ctype(arg: u32) -> Self {
+        match arg {
+            0 => TCReplicaOpType::Create,
+            1 => TCReplicaOpType::Delete,
+            2 => TCReplicaOpType::Update,
+            3 => TCReplicaOpType::UndoPoint,
+            _ => panic!("Bad TCReplicaOpType."),
+        }
+    }
+}
```
djmitche: This appears to be unused - the type is just handled as a C type

taskchampion/lib/src/replica.rs
```diff
@@ -186,6 +235,156 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
...
+            let value_ruststring = unsafe { TCString::val_from_arg(kv.value) };
+
+            taskmap.insert(
+                key_ruststring.into_string().unwrap(),
+                value_ruststring.into_string().unwrap(),
+            );
+        });
+        taskmap
+    }
+}
```
djmitche: This appears to be unused!

taskchampion/lib/src/task.rs
```diff
@@ -1,6 +1,7 @@
 use crate::traits::*;
 use crate::types::*;
 use crate::util::err_to_ruststring;
+use crate::TCKV;
```
djmitche, line 4: Unnecessary?

taskchampion/lib/src/kv.rs
```diff
@@ -68,11 +71,28 @@ impl PassByValue for TCKV {
...
+                let key = RustString::from(k.as_ref());
+                let value = RustString::from(v.as_ref());
+                TCKV::as_ctype((key, value))
+            })
+            .collect();
+        // SAFETY:
+        //  - caller will free this list
+        unsafe { TCKVList::return_val(vec) }
+    }
 }
```
djmitche: This appears to be unused.

In general, we should prefer to use the PassByValue or PassByPointer types, rather than From/Into, as those have clearer safety requirements. For example, the `From<TCKVList> for TaskMap` implementation implicitly assumes that the TCKVList will not be used anymore by the caller.

taskchampion/lib/src/replica.rs
```diff
@@ -186,6 +235,156 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
...
+                tc_replica_op_list.capacity,
+            )
+        };
+
+        tc_replica_op_vec
+            .into_iter()
+            .map(|tc_op| *tc_op.inner)
+            .collect()
+    }
+}
```
djmitche: This is another `From` implementation that hides the unsafety. It can be removed with the suggested change to `tc_replica_commit_undo_ops`, below.

taskchampion/lib/src/replica.rs
```diff
@@ -186,6 +235,156 @@ pub unsafe extern "C" fn tc_replica_new_on_disk(
...
+            .map(|op| TCReplicaOp::from(op))
+            .collect();
+        let (ptr, len, capacity) = vec_into_raw_parts(tc_replica_op_vec);
+        TCReplicaOpList {
+            items: ptr as *mut TCReplicaOp,
+            len,
+            capacity,
+        }
+    }
+}
```
djmitche: The PassByValue implementation for TCReplicaOpList should be able to handle this, although the mapping needs to be handled there. See another comment below.

taskchampion/lib/src/replica.rs
```diff
@@ -408,20 +607,41 @@ pub unsafe extern "C" fn tc_replica_sync(
...
+/// ```c
+/// EXTERN_C TCReplicaOpList tc_replica_get_undo_ops(struct TCReplica *rep);
+/// ```
+#[no_mangle]
+pub unsafe extern "C" fn tc_replica_get_undo_ops(rep: *mut TCReplica) -> TCReplicaOpList {
+    wrap(
+        rep,
+        |rep| Ok(TCReplicaOpList::from(rep.get_undo_ops()?)),
+        TCReplicaOpList::from(Vec::new()),
+    )
```
djmitche: Use the PassByValue mechanics here, after mapping the item types:
```suggestion
    wrap(
        rep,
        |rep| {
            // SAFETY:
            //  - caller will free this value, either with tc_replica_commit_undo_ops or
            //  tc_replica_op_list_free.
            Ok(unsafe {
                TCReplicaOpList::return_val(
                    rep.get_undo_ops()?
                        .into_iter()
                        .map(TCReplicaOp::from)
                        .collect(),
                )
            })
        },
        // SAFETY:
        //  - caller will free this value, either with tc_replica_commit_undo_ops or
        //  tc_replica_op_list_free.
        unsafe { TCReplicaOpList::return_val(Vec::new()) },
    )
```
(That last unsafe bit could be replaced with a `Default` implementation for `TCReplicaOpList`)

taskchampion/lib/src/replica.rs
```diff
@@ -408,20 +607,41 @@ pub unsafe extern "C" fn tc_replica_sync(
...
+pub unsafe extern "C" fn tc_replica_commit_undo_ops(
+    rep: *mut TCReplica,
+    tc_undo_ops: TCReplicaOpList,
+    undone_out: *mut i32,
+) -> TCResult {
     wrap(
         rep,
         |rep| {
-            let undone = i32::from(rep.undo()?);
+            let undo_ops: Vec<ReplicaOp> = Vec::from(tc_undo_ops);
```
djmitche:
```suggestion
            // SAFETY:
            // - `tc_undo_ops` is a valid value, as it was acquired from `tc_replica_get_undo_ops`.
            let undo_ops: Vec<ReplicaOp> = unsafe { TCReplicaOpList::val_from_arg(tc_undo_ops) }
                .into_iter()
                .map(|op| *op.inner)
                .collect();
```

taskchampion/lib/src/replica.rs
```diff
@@ -565,3 +785,179 @@ pub unsafe extern "C" fn tc_replica_free(rep: *mut TCReplica) {
...
+    //  - len is properly aligned (promised by caller)
+    let len = unsafe { (*oplist).len };
+    // SAFETY:
+    //  - capacity is not NULL
+    //  - capacity is properly aligned (promised by caller)
+    let capacity = unsafe { (*oplist).capacity };
+    // SAFETY:
+    //  - items, len, and capacity are the unmodified values returned by vec_into_raw_parts
+    //  (promised by caller)
+    unsafe { Vec::from_raw_parts(items, len, capacity) };
```
djmitche: Again preferring to use the PassByValue functionality:
```suggestion
    debug_assert!(!oplist.is_null());
    // SAFETY:
    // - arg is not NULL (just checked)
    // - `*oplist` is valid (guaranteed by caller not double-freeing this value)
    unsafe {
        TCReplicaOpList::take_val_from_arg(
            oplist,
            // SAFETY:
            //  - value is empty, so the caller need not free it.
            TCReplicaOpList::return_val(Vec::new()),
        )
    };
```

djmitche (changes requested):
Looks good! A lot of the `From` impls can be removed here, and I've added suggestions to prefer using the PassByValue methods, which come with handy lists of safety things to check.

If my copy-pasting is correct, I've tried all of these changes locally and they have succeeded.

Again, thanks for your hard work here  :)

taskchampion/taskchampion/src/lib.rs
ryneeverett: This is used in `taskchampion/taskchampion/src/replica.rs`:

```rs
use crate::taskdb::TaskDb;
```

djmitche: That's in the same crate, though, so does not require public visibility.

taskchampion/lib/src/task.rs
ryneeverett: Without this I get these errors:

```
error[E0412]: cannot find type `TCKV` in this scope
   --> taskchampion/lib/src/task.rs:281:22
    |
281 |         let vec: Vec<TCKV> = task
    |                      ^^^^ not found in this scope
    |
help: consider importing this struct through its public re-export
    |
1   + use crate::TCKV;
    |

error[E0433]: failed to resolve: use of undeclared type `TCKV`
   --> taskchampion/lib/src/task.rs:287:17
    |
287 |                 TCKV::as_ctype((key, value))
    |                 ^^^^ use of undeclared type `TCKV`
    |
help: consider importing this struct through its public re-export
    |
1   + use crate::TCKV;
```

I'm not sure why this isn't popping up in `develop`.

djmitche (approved):
Looks good! That was a legendary PR :)

Issue #3185, opened 2023-10-06
Cloud-Storage Backend


djmitche:
Taskchampion supports pluggability for the server implementation, and currently has two:
 - Local (just sync to a file on disk)
 - Server (sync to `taskchampion-sync-server`)

Let's consider adding a third option: sync to a cloud-storage service like Amazon S3, Google Cloud Storage, or Azure Blob Store. If we could figure out how to implement this in each of the cloud-storage services, then casual users could simply configure their Taskwarrior's with some cloud credentials and an encryption key, and be on their way.

Some quick and probably incomplete thinking on the topic:

The server API is basically reading existing versions and  writing a versions, where each version is contrent-addressible: its "name" is based on its hash. Storage APIs support GET and PUT! The tricky bit is that there must be exactly one "latest" version, so the PUT operation must be a kind of compare-and-swap of the "latest version". 

Do any/all of the cloud storage providers have a mechanism to support this? Either directly in the storage service, or in another service on the same cloud provider?

djmitche:
@ryneeverett ^^ I'd be curious for your thoughts here

dathanb:
Which part would need to be compared-and-swapped? If each version is content-addressable, then collisions only happen if the two writes are actually the same, which is fine, right? So is there an index that points to all the latest versions that needs mutual exclusion during updates?

If that's the case, this should be doable on Amazon S3. All versions can be written without checks (based on the above reasoning about uniqueness), and the new index would be written by calling `CopyObject` on the old index and specifying the `x-amz-copy-source-if-match` predicate identifying the version of the index you intend to replace. There would just need to be a one-time bootstrapping operation where the initial index file is written using `PutObject`, which doesn't support any sort of checks or mutual exclusion.

djmitche:
The part requiring CAS is setting the (single) latest version. We want to build a chain of versions, so the upload of a new version is fine and doesn't conflict, but that must only become the "latest" version if its parent version is the current "latest" version. Thus compare (latest == new version's parent) and swap (swap in new version). If that operation fails, then the client needs to "rebase" onto the new latest version and try again.

CopyObject's conditional functionality sounds useful! I guess that the add-new-version process would go like this:
* Get `latest`, with the content (a hash) going to LATEST, and the object's ETag going to LATEST_ETAG
* Create a new version with hash HASH and based on version LATEST
* Upload it as `versions/HASH`
* Create an object containing HASH at `hashes/HASH`
[...]

PR #3216, opened 2023-12-23
Make the sync server client an optional feature

djmitche:
#### Description

This makes the implementation of syncing that communicates with `taskchampion-sync-server` optional.

This also makes `crypto` an optional feature, as that will also be required for cloud backends (#3185)


djmitche:
@ryneeverett do you mind having a look? I can't actually assign you the review..

taskchampion/taskchampion/src/server/crypto.rs
```diff
@@ -1,8 +1,8 @@
+#![allow(dead_code)] // temporary
```
ryneeverett: How temporary? If you're going to commit this maybe say why it is needed?

taskchampion/taskchampion/src/server/sync/mod.rs
```diff
@@ -21,19 +21,14 @@ const HISTORY_SEGMENT_CONTENT_TYPE: &str = "application/vnd.taskchampion.history
 /// The content-type for snapshots (opaque blobs of bytes)
 const SNAPSHOT_CONTENT_TYPE: &str = "application/vnd.taskchampion.snapshot";
 
-/// A RemoeServer communicates with a remote server over HTTP (such as with
-/// taskchampion-sync-server).
-impl RemoteServer {
-    /// Construct a new RemoteServer.  The `origin` is the sync server's protocol and hostname
+/// A RemoeServer communicates with a taskchampion-sync-server over HTTP .
```
ryneeverett: - RemoeServer -> RemoteServer -> SyncServer
- Remove extra space before period .

BTW, I like the rename. As you figured out, RemoteServer was destined to be too ambiguous once we have cloud backup options.

However, it's interesting that you changed "a remote server over HTTP" to "taskchampion-sync-server" when in fact ["The server-replica protocol is defined abstractly..."](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/taskchampion/docs/src/sync-protocol.md). Maybe I'm reading too much into it since this was clearly a hastily changed line.

ryneeverett (commented):
Looks good to me other than a couple nitpicks.

I'm curious about your motivation for breaking out these features. Is it just about enabling the reduction of unneeded dependencies? Or does it somehow encourage sound design, like decoupling of feature code?

taskchampion/taskchampion/src/server/sync/mod.rs
djmitche: That's a good point -- I didn't think too hard about this edit :)

djmitche:
It's mostly about dependency management. The cloud providers all seem to be async and based on tokio, so they're bringing in a bunch of tokio crates plus a bunch of http crates.

taskchampion/taskchampion/src/server/crypto.rs
djmitche: Oops, too temporary to include in this PR!

ryneeverett (approved):

PR #3227, opened 2023-12-25
Disable Clang's return-type-c-linkage warning

### LINKED ISSUE #3225: taskchampion.h: define TCTaskList before functions returning it
Lots of this on the mac builders:
```
/Users/runner/work/taskwarrior/taskwarrior/taskchampion/lib/taskchampion.h:508:28: warning: 'tc_replica_all_tasks' has C-linkage specified, but returns incomplete type 'struct TCTaskList' which could be incompatible with C [-Wreturn-type-c-linkage]
```

djmitche:
#### Description

Hopefully this makes the mac builders (which I think use Clang?) happier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,9 @@ if (SOLARIS)
 endif (SOLARIS)
 
 # Disable the Clang return-type-c-linkage warning globally. See #3225.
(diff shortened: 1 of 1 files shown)


djmitche:
Much better! Now, who knows enough CMake to review this?

Issue #3214, opened 2023-12-22
Question about adding a hook


MZAWeb:
I have an use case where it'd be very helpful to have a hook at the time of urgency calculation. Basically here: https://github.com/GothenburgBitFactory/taskwarrior/blob/1c11f8b650cdd15526c6694d9032b645485660fa/src/Task.cpp#L2202

Having such hook, while potentially detrimental to speed (like any other hook), would give advanced users a lot of flexibility to bring interesting "business" rules to the urgency calculation. Just as an example, what I'd like to do is change the urgency of a task based on a formula taking as input the value of a numeric UDA I defined. 

I'm happy to implement it and make a PR, but before spending the time I'd like to know if there's interest and a feature like that has any chance of being considered for merging.

Thanks!

ryneeverett:
See also #133 for some ideas about how dynamic urgency calculation might work within taskwarrior configuration.

MZAWeb:
I saw that @ryneeverett but a few thoughts
1) Adding the commitment to parse and execute math formulas from the config file would be a huge step. That feature needs a lot of thinking and testing to make it viable. 
2) Even if we did that, a hook for the dynamic urgency calculation is still a good idea, as it'd allow users to do _whatever_ they want and not just formulas. You may want to check an external source, etc.

ryneeverett:
Fair points! 

My line of thinking is that urgency calculation within configuration requires a pattern that wouldn't have the performance problems of checking external sources on-demand:

1. An external daemon updates the taskwarrior database with whatever data you need from external sources
2. Taskwarrior dynamically calculates urgency based on that data.

Of course this could just as easily be implemented with a hook for (2) and maybe that's a better idea.

MZAWeb:
I had some time during a long flight and put a PR together. Would love to get some comments.

PR #3229, opened 2023-12-26
New hook on urgency calculation

MZAWeb:
#### Description

This PR adds a new on-urgency-calc hook to allow external scripts to modify the urgency of tasks.

- See https://github.com/GothenburgBitFactory/taskwarrior/issues/3214 for context.
- Added tests for the new code
- See sample hook file https://gist.github.com/MZAWeb/e59886bae9227c8a7aeb0b9c6428aa82

### DIFF
--- a/src/Hooks.cpp
+++ b/src/Hooks.cpp
@@ -27,6 +27,7 @@
 #include <cmake.h>
 #include <Hooks.h>
 #include <algorithm>
+#include <string>
 // If <iostream> is included, put it after <stdio.h>, because it includes
 // <stdio.h>, and therefore would ignore the _WITH_GETLINE.
 #ifdef FREEBSD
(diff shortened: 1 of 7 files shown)


MZAWeb:
The failing check seems to be an issue with the environment.

MZAWeb:
@djmitche while the longer term vision[ is being discussed](https://github.com/GothenburgBitFactory/taskwarrior/discussions/3235), do you see any path to get this one hook in? Would help a lot with some stuff I want to implement.

djmitche:
I don't know enough about hooks to feel comfortable reviewing this idea. If someone who uses and understands hooks can look and verify that the idea is nicely in-line with existing hooks, that would be helpful.

MZAWeb:
Makes sense, thanks!

djmitche:
closing as stale. it's still here if we want to re-vitalize it!

PR #3241, opened 2024-01-16
Fix invalid imports in `[{..}]` form

### LINKED ISSUE #3240: Segfault when importing data containing spurious newlines
#### To report a bug...

```
$ cat > import.json << 'EOF'
[{"description":"AAA","project":"Self","tags":[],"annotations":[],"id":24,"entry":"20240113T062723Z","modified":"20240113T173234Z","reviewed":"20240113T173234Z","status":"pending","uuid":"e5139954-71ee-417a-ac17-de801
f2d2454","urgency":0}]
EOF

$ task import < import.json
Importing 'STDIN'
Segmentation fault
```
Interestingly, this works:

```
task import <<< '[{"description":"AAA","project":"Self","tags":[],"annotations":[],"id":24,"entry":"20240113T062723Z","modified":"20240113T173234Z","r
eviewed":"20240113T173234Z","status":"pending","uuid":"e5139954-71ee-417a-ac17-de801
```

* Paste the output of the ```task diag``` command.
```
Unrecognized Taskwarrior file format or blank line in data. in /home/ira/.task/pending.data at line 70
```

Lines 69 and 70 from pending.data:
```
[description:"AAA" entry:"1705127243" modified:"1705204925" project:"Self" r
eviewed:"20240113T173234Z" status:"pending" uuid:"e5139954-71ee-417a-ac17-de801\nf2d2454"]
```
So, there is a stray  `\n` at line 69:  `r\neviewed`.

This happens on win10 wsl, on both 2.5.1 and 2.6.2

djmitche:
Before this change, if an import of data that takes the form of a JSON array of JSON objects results in an error, the import would be re-tried assuming that each line of the file is a JSON object (the old format). However, no check was made that the value actually was an object before casting it to `json::object`, resulting in a segfault.

This adds the check, and handles the failure with a useful error message (from the first attempt to parse the file).

Fixes #3240.

### DIFF
--- a/src/commands/CmdImport.cpp
+++ b/src/commands/CmdImport.cpp
@@ -142,18 +142,25 @@ int CmdImport::import (const std::string& input)
   //   { ... }
   catch (std::string& e)
   {
+    // Read the entire file, so that errors do not result in a partial import.
+    std::vector<std::unique_ptr<json::object>> objects;
     for (auto& line : split (input, '\n'))
     {
       if (line.length ())
       {
         json::value* root = json::parse (line);
-        if (root)
-        {
-          importSingleTask ((json::object*) root);
-          ++count;
-          delete root;
-        }
+        if (root && root->type () == json::j_object)
+          objects.push_back (std::unique_ptr<json::object> ((json::object *)root));
+        else
+          // Throw the existing error, as it's likely more accurate.
+          throw e;
       }
+    }
+
+    // Import the tasks.
+    for (auto& root : objects) {
+      importSingleTask (root.get());
+      ++count;
     }
   }
 
--- a/src/commands/CmdImport.cpp
+++ b/src/commands/CmdImport.cpp
@@ -133,7 +133,7 @@ int CmdImport::import (const std::string& input)
   }
 
   // If an exception is caught, then it is because the free-form JSON
-  // objects/array above failed to parse. This is an indication that the input
+  // objects/array above failed to parse. This may be an indication that the input
   // is an old-style line-by-line set of JSON objects, because both an array of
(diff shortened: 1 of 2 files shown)


ryneeverett (approved):
This is an improvement both in respect to the error message and to parsing all the lines before importing, but I wonder if it would be better to use a smarter heuristic than a blanket exception to determine whether to `throw e` or start parsing the input in the old-style. I'm thinking at the beginning of the `catch` block something like:

```cpp
catch (std::string& e) {
  if (input[0] != "{") {
    throw e;
  }
  // Falling back to old-style...
}
```

This way we could *theoretically* throw helpful error messages for users with old-style input rather than assuming on the basis of no evidence that we're dealing with an invalid input of the new-style rather than the old-style.

djmitche:
Seems like a good idea!

src/commands/CmdImport.cpp
```diff
@@ -142,19 +142,31 @@ int CmdImport::import (const std::string& input)
...
-        {
-          importSingleTask ((json::object*) root);
-          ++count;
-          delete root;
-        }
+        if (root && root->type () == json::j_object)
+          objects.push_back (std::unique_ptr<json::object> ((json::object *)root));
+        else
+          // Throw the existing error, as it's likely more accurate.
+          throw e;
```
ryneeverett: My thought was that it would no longer be necessary to throw an error from the new-style parsing in the middle of old-style parsing. Seems like we should commit to a given format and act accordingly. Maybe just `throw format("Invalid Json: {1}", line)`?

ryneeverett (approved):

PR #3247, opened 2024-01-21
Verify add_snapshot is called with the right version

### LINKED ISSUE #3222: Cloud server: validate new snapshots
The docs say

> The server should validate that the snapshot is for an existing version and is newer than any existing snapshot.

..and we don't currently do that in the PR being prepared for #3185.

Doing so will require iterating over the chain of versions until either finding a version matching the new snapshot or finding a version matching an existing snapshot.

djmitche:
This verification is just checking an assertion from elsewhere in the same binary, but it doesn't hurt!

Fixes #3222.

### DIFF
--- a/taskchampion/taskchampion/src/server/cloud/server.rs
+++ b/taskchampion/taskchampion/src/server/cloud/server.rs
@@ -65,6 +65,9 @@ pub(in crate::server) struct CloudServer<SVC: Service> {
     /// The probability (0..255) that this run will perform cleanup.
     cleanup_probability: u8,
 
+    /// The last version added locally with the `add_version` method.
+    last_version_added: Option<VersionId>,
+
     /// For testing, a function that is called in the middle of `add_version` to simulate
     /// a concurrent change in the service.
     #[cfg(test)]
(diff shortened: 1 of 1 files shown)


ryneeverett (commented):
I don't understand the purpose of this guarantee. Couldn't `add_snapshot` error due to the server restarting and not having the `last_version_added` or another replica pushing a new version? Why would we want an error in such cases?

djmitche:
> I don't understand the purpose of this guarantee.

It is a bit redundant in this situation, where the check is being performed locally and not on the "server" (since there is no server).

> Couldn't `add_snapshot` error due to the server restarting and not having the `last_version_added`

We're assuming that once an object is added to the object store, it won't disappear. All three major cloud providers satisfy this guarantee (with some number of 9's, at least).

> or another replica pushing a new version? Why would we want an error in such cases?

If another replica pushes a new version, that won't affect the local value of `last_version_added`.

We don't need to merge this -- I added it more as a double-check than anything else.

djmitche:
Hm, the following sentence in the docs was the origin of this:

> The server should validate that the snapshot is for an existing version and is newer than any existing snapshot.

But in the context of an (eventually-consistent) object store, "any existing snapshot" is not really a well-defined concept.

I'll scrap this PR and just open one to modify or remove that sentence.

PR #3248, opened 2024-01-21
Update to supported linux distros

djmitche:
This should save a bit of CI time and also ensure that TW is compatible with the latest distro versions.

This doesn't imply a lack of support for older distros -- just not testing on unsupported versions anymore.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -66,27 +66,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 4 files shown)


djmitche:
@tbabej I think this will require updating the branch config in the GitHub settings to not expect the removed versions, and to expect the new versions.

tbabej (approved):
CI changes look good to me, I tweaked the relevant settings for the expected checks and we're good to go now.

PR #3253, opened 2024-01-21
Better links from TW devel docs to TaskChampion

djmitche:
This should help newcomers to the project to navigate the TW/TC distinction and find the relevant code, especially when 3.0 is released.

Fixe #2868.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,6 +1,6 @@
 # Documentation
 
-This directory contains Taskwarrior that is built and installed along with the executable:
+This directory contains Taskwarrior documentation that is built and installed along with the executable:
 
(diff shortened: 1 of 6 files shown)


ryneeverett (approved):
This is a great step in bridging the gap. It would be nice if there were also a section somewhere on developing both tw and tc through the ffi as `rm -rf src/tc/rust/x86_64-unknown-linux-gnu/debug/libtc_rust.a && cargo xtask codegen && cargo build && cmake -DCMAKE_BUILD_TYPE=debug . && make VERBOSE=1 -j4` is far from obvious.

djmitche:
Yeah, I really don't know how to fix that. But, it's worth adding a note to it in the docs. I'll do that in another PR.

PR #3256, opened 2024-01-21
Change 'should' to 'may' in server snapshot requirements

### LINKED ISSUE #3222: Cloud server: validate new snapshots

djmitche:
For an object store backend, none of the suggestions in this paragraph make much sense. "May" is much softer, in terms of https://datatracker.ietf.org/doc/html/rfc2119. And in particular, the cloud server implementation does not do any of these things.

Fixes #3222, and replaces #3247.

### DIFF
--- a/taskchampion/docs/src/sync-protocol.md
+++ b/taskchampion/docs/src/sync-protocol.md
@@ -102,7 +102,7 @@ The request contains the following:
  * version ID at which the snapshot was made, and
  * encrypted snapshot data.
 
-The server should validate that the snapshot is for an existing version and is newer than any existing snapshot.
+The server may validate that the snapshot is for an existing version and is newer than any existing snapshot.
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
From [rfc2119](https://datatracker.ietf.org/doc/html/rfc2119): "These words are often capitalized." On this particular (protocol) page you might consider capitalizing such words so that they're read in this context. I can't explain why but "may" can convey indecisiveness whereas "MAY" seems more intentional. It's definitely a matter of preference though.

djmitche:
I suspect these docs will need a much more careful analysis before they become anything standards-like, so let's call this good enough.

PR #3261, opened 2024-01-24
Use the correct formatting for `tag_` properties

### LINKED ISSUE #3254: taskchampion: get_tags() not fetching user tags
#### To report a bug...

get_tags() seems to not fetch user tags

After creating a task with a tag, I fetch all the tags using the Rust API
Sample code:
```
fn main() -> Result<(), taskchampion::Error> {
    let storage = Box::new(SqliteStorage::new(SQLITE_DB, false)?);
    let mut replica = Replica::new(storage);

    let tasks = replica.all_tasks()?;
    for (u, t) in tasks.clone() {
            for tag in t.get_tags(){
                println!("tag:{}", tag);
            }
        dbg!(t);
    }

    Ok(())
}
```
Result:
```
tag:PENDING
tag:UNBLOCKED
[src/main.rs:14:9] t = Task {
    uuid: 16d379dc-ae97-4e53-8b0e-5637c49a59ed,
    taskmap: {
        "description": "task with a tag",
        "tags": "test",
        "status": "pending",
        "entry": "1705876311",
        "tags_test": "x",
        "modified": "1705876311",
    },
    depmap: DependencyMap {
        edges: [],
    },
}
```

- Built and tested on  develop branch rev 6f1c16fecda2c47683b5a07dca4975e1bd40742d

djmitche:
This resolves a discrepancy between Taskwarrior and Taskchampion as to how tags should be represented.

Note that this appears to be a breaking change, but it will be covered by the existing need to `task export` with a 2.6.x task DB and then `task import` into 3.0. The export format puts tags in a JSON array, so the `tags_..` property does not appear there.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -1537,7 +1537,7 @@ void Task::removeTag (const std::string& tag)
 ////////////////////////////////////////////////////////////////////////////////
 void Task::fixTagsAttribute ()
 {
-  // Fix up the old `tags` attribute to match the `tags_..` attributes (or
+  // Fix up the old `tags` attribute to match the `tag_..` attributes (or
   // remove it if there are no tags)
   auto tags = getTags ();
   if (tags.size () > 0) {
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
Makes sense to me, though I'd give @aktaboot a chance to weigh in.

aktaboot:
sure, all the same to me :) let's settle with this one then

djmitche:
Thanks!

PR #3262, opened 2024-01-25
added BYOS (Bring Your Own SERVICE_ACCOUNT) for GCS authentication

akashShanmugraj:
#### Description

This pull request allows the application to authenticate its Google Cloud Storage buckets through given SERVICE_ACCOUNT credential. 

### DIFF
--- a/taskchampion/lib/src/server.rs
+++ b/taskchampion/lib/src/server.rs
@@ -164,7 +164,7 @@ pub unsafe extern "C" fn tc_server_new_sync(
 ///
 /// ```c
 /// EXTERN_C struct TCServer *tc_server_new_gcp(struct TCString bucket,
-///                                       struct TCString credentia_path,
+///                                       struct TCString credential_path,
 ///                                       struct TCString encryption_secret,
 ///                                       struct TCString *error_out);
 /// ```
--- a/taskchampion/lib/src/server.rs
+++ b/taskchampion/lib/src/server.rs
@@ -187,7 +187,7 @@ pub unsafe extern "C" fn tc_server_new_gcp(
             //  - credential_path ownership is transferred to this function
 
             let credential_path =
-                    unsafe { TCString::val_from_arg(credential_path_argument) }.into_string()?;
+                unsafe { TCString::val_from_arg(credential_path_argument) }.into_string()?;
             let credential_path = if credential_path == "" {
                 None
             } else {
@@ -202,7 +202,7 @@ pub unsafe extern "C" fn tc_server_new_gcp(
                 .to_vec();
             let server_config = ServerConfig::Gcp {
                 bucket,
-                credential_path: Some(credential_path),
+                credential_path: credential_path.map(|cp| cp.to_string()),
                 encryption_secret,
             };
             let server = server_config.into_server()?;
--- a/taskchampion/lib/src/server.rs
+++ b/taskchampion/lib/src/server.rs
@@ -169,7 +169,7 @@ pub unsafe extern "C" fn tc_server_new_sync(
 ///                                       struct TCString *error_out);
 /// ```
 #[no_mangle]
-pub unsafe extern "C" fn tc_server_new_gcp( 
+pub unsafe extern "C" fn tc_server_new_gcp(
     bucket: TCString,
     credential_path_argument: TCString,
     encryption_secret: TCString,
@@ -185,16 +185,17 @@ pub unsafe extern "C" fn tc_server_new_gcp(
             // SAFETY:
             //  - credential_path is valid (promised by caller)
             //  - credential_path ownership is transferred to this function
-            
-            let credential_path:String;
+
+            let credential_path: String;
 
             if credential_path_argument.is_null() {
                 credential_path = String::from("");
             } else {
-                credential_path = unsafe { TCString::val_from_arg(credential_path_argument) }.into_string()?;
+                credential_path =
+                    unsafe { TCString::val_from_arg(credential_path_argument) }.into_string()?;
             }
 
-            // SAFETY:  
+            // SAFETY:
             //  - encryption_secret is valid (promised by caller)
             //  - encryption_secret ownership is transferred to this function
             let encryption_secret = unsafe { TCString::val_from_arg(encryption_secret) }
--- a/taskchampion/lib/src/server.rs
+++ b/taskchampion/lib/src/server.rs
@@ -170,6 +170,7 @@ pub unsafe extern "C" fn tc_server_new_sync(
 #[no_mangle]
 pub unsafe extern "C" fn tc_server_new_gcp(
     bucket: TCString,
+    credentialpath: TCString,
     encryption_secret: TCString,
     error_out: *mut TCString,
 ) -> *mut TCServer {
@@ -179,7 +180,12 @@ pub unsafe extern "C" fn tc_server_new_gcp(
             //  - bucket is valid (promised by caller)
             //  - bucket ownership is transferred to this function
             let bucket = unsafe { TCString::val_from_arg(bucket) }.into_string()?;
-
+            
+            // SAFETY:
+            //  - credentialpath is valid (promised by caller)
+            //  - credentialpath ownership is transferred to this function
+            let credentialpath = unsafe { TCString::val_from_arg(credentialpath) }.into_string()?;
+            
             // SAFETY:
             //  - encryption_secret is valid (promised by caller)
             //  - encryption_secret ownership is transferred to this function
@@ -189,6 +195,7 @@ pub unsafe extern "C" fn tc_server_new_gcp(
 
             let server_config = ServerConfig::Gcp {
                 bucket,
+                credentialpath,
                 encryption_secret,
             };
             let server = server_config.into_server()?;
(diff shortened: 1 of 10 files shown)


djmitche:
This looks like it still has the GCP commits in it. Can you rebase to only include your changes?

akashShanmugraj:
Hey, sorry I can't understand which "GCP" commits I should exclude?

I should remove these commits?

Screenshot 2024-01-25 at 7 36 25 PM

ryneeverett:
I would try:

```bash
git remote set-url upstream https://github.com/GothenburgBitFactory/taskwarrior.git
git fetch upstream
git rebase upstream/develop
git push -f
```

akashShanmugraj:
> I would try:
> 
> ```shell
> git remote set-url upstream https://github.com/GothenburgBitFactory/taskwarrior.git
> git fetch upstream
> git rebase upstream/develop
> git push -f
> ```

Hey, thank @ryneeverett!
What should the `upstream` point to?

I am getting this :
```
$ git remote set-url upstream https://github.com/GothenburgBitFactory/taskwarrior.git
error: No such remote 'upstream'
```

ryneeverett:
My bad, I forgot that if you don't already have `upstream` set you have to `add` it rather than `set-url`:

```bash
git remote add upstream https://github.com/GothenburgBitFactory/taskwarrior.git
```

djmitche:
Thanks @ryneeverett! If that produces a lot of conflicts, you may want to use the interactive rebase:
```
git rebase -i upstream/develop
```
and then remove all of the `pick` lines for commits authored by me. You may also want to remove "removed unused import" and "added atomic import back" since those were merged in another PR.

akashShanmugraj:
Hey! I just realised something.
If you look at [/src/commands/CmdSync.cpp](https://github.com/akashShanmugraj/taskwarrior/blob/f668dd052a73277a148c8ad1cef1be84c22ad3a0/src/commands/CmdSync.cpp#L67) I would have added the line 67:

```rust
std::string gcp_credential_path = Context::getContext ().config.get ("sync.gcp.credential_path");
```

(I assume that) It gets the path to the credential file from the configuration file at `.taskrc` corresponding the config variable name `sync.gcp.credential_path`

And in [/taskchampion/taskchampion/src/server/cloud/gcp.rs](https://github.com/akashShanmugraj/taskwarrior/blob/f668dd052a73277a148c8ad1cef1be84c22ad3a0/taskchampion/taskchampion/src/server/cloud/gcp.rs#L257) I would have added lines 257 & 258:

```rust
let Ok(credential_path) = std::env::var("GCP_CREDENTIAL_PATH") else {
    return None;
};
```

This implementation gets the path of the file containing service account credentials from the operating system's environment variable `GCP_CREDENTIAL_PATH`

I am not sure if this is how it should be done

ryneeverett:
If I understand the question correctly you're asking whether the path should be configured via taskrc, environmental variable, or both? I'd say taskrc only.

djmitche:
Agreed. We do not typically take configuration from `GCP_CREDENTIAL_PATH`.

taskchampion/lib/src/server.rs
```diff
@@ -180,15 +182,27 @@ pub unsafe extern "C" fn tc_server_new_gcp(
             //  - bucket ownership is transferred to this function
             let bucket = unsafe { TCString::val_from_arg(bucket) }.into_string()?;
 
+            // SAFETY:
+            //  - credential_path is valid (promised by caller)
+            //  - credential_path ownership is transferred to this function
+
+            let credential_path =
+                unsafe { TCString::val_from_arg(credential_path_argument) }.into_string()?;
+            let credential_path = if credential_path == "" {
```
djmitche: I think clippy is suggesting using `.is_empty()` instead of comparing to `""` here.

taskchampion/lib/src/server.rs
```diff
@@ -180,15 +182,27 @@ pub unsafe extern "C" fn tc_server_new_gcp(
...
             // SAFETY:
             //  - encryption_secret is valid (promised by caller)
             //  - encryption_secret ownership is transferred to this function
             let encryption_secret = unsafe { TCString::val_from_arg(encryption_secret) }
                 .as_bytes()
                 .to_vec();
-
             let server_config = ServerConfig::Gcp {
                 bucket,
+                credential_path: credential_path.map(|cp| cp.to_string()),
```
djmitche: So at this point, `credential_path` has type `Option<String>`. There's no need to do anything further to it to pass it into the `ServerConfig::Gcp` struct.

taskchampion/taskchampion/src/server/config.rs
```diff
@@ -37,7 +37,7 @@ pub enum ServerConfig {
         /// Bucket in which to store the task data. This bucket must not be used for any other
         /// purpose.
         bucket: String,
-
+        credential_path: Option<String>,
```
djmitche, line 45: Please add a docstring here. It's fine to link to a GCP docs page about credentials files.

taskchampion/taskchampion/src/server/cloud/gcp.rs
```diff
@@ -244,10 +253,16 @@ mod tests {
         let Ok(bucket) = std::env::var("GCP_TEST_BUCKET") else {
             return None;
         };
+
+        let Ok(credential_path) = std::env::var("GCP_CREDENTIAL_PATH") else {
+            return None;
+        };
+
```
djmitche: Oh, I didn't realize this was in tests. Let's use `GCP_TEST_CREDENTIAL_PATH` here.

djmitche (changes requested):
This looks good -- just a few tiny things left!

taskchampion/lib/src/server.rs
akashShanmugraj: Okay!

taskchampion/taskchampion/src/server/cloud/gcp.rs
akashShanmugraj: Okay!

taskchampion/taskchampion/src/server/config.rs
akashShanmugraj: Ah I am extremely sorry, I should have added these docstrings in the previous review itself. Can you check [9eb49fd](https://github.com/akashShanmugraj/taskwarrior/commit/9eb49fd3153cf94c0d63224986e0bce86ecf369c) if the docstring is sufficient or too much?

akashShanmugraj:
> If I understand the question correctly you're asking whether the path should be configured via taskrc, environmental variable, or both? I'd say taskrc only.

Yes, exactly! But how does the application currently take in that (CREDENTIAL_PATH) value?

taskchampion/lib/src/server.rs
akashShanmugraj: Okay!

ryneeverett:
> > If I understand the question correctly you're asking whether the path should be configured via taskrc, environmental variable, or both? I'd say taskrc only.
> 
> Yes, exactly! But how does the application currently take in that (CREDENTIAL_PATH) value?

I don't understand the question. It looks like you've already implemented all the code to parse `sync.gcp.credential_path` from taskrc.

akashShanmugraj:
> I don't understand the question. It looks like you've already implemented all the code to parse `sync.gcp.credential_path` from taskrc.

Yes, then why do I need these lines in [/taskchampion/taskchampion/src/server/cloud/gcp.rs](https://github.com/akashShanmugraj/taskwarrior/blob/f668dd052a73277a148c8ad1cef1be84c22ad3a0/taskchampion/taskchampion/src/server/cloud/gcp.rs#L257) (lines 257 & 258)?

```rust
let Ok(credential_path) = std::env::var("GCP_TEST_CREDENTIAL_PATH") else {
    return None;
};
```

ryneeverett:
I'm not sure but my guess is that the rust test infrastructure doesn't have a configuration system linked to taskrc.

I suspect you mimicked the pattern of `GCP_TEST_BUCKET`. I'm not sure that's necessary though because it seems like you could use a hard-coded relative path in the test. The test bucket is a special case because that actually needs to be configurable since we aren't going to hard-code credentials. (See #3263.)

akashShanmugraj:
> I suspect you mimicked the pattern of `GCP_TEST_BUCKET`

Ha actually yes :)

> The test bucket is a special case because that actually needs to be configurable since we aren't going to hard-code credentials.

Oh Okay!

ryneeverett:
Oh, I got it wrong, GCP_TEST_CREDENTIAL_PATH is also an exception because it also won't be available in the default case unless configured. But I think we're on the same page now and you understand why the credential path needs to be handled differently in test than in the application.

akashShanmugraj:
> But I think we're on the same page now and you understand why the credential path needs to be handled differently in test than in the application.

Yes, thanks @ryneeverett ! 😄

djmitche (approved):
Looks great! Thanks!

akashShanmugraj:
Hey @djmitche thanks for helping me out!! (even after me being sometimes irregular with the followups) 
This is actually my first PR to an open source repository and you have made my first experience so awesome.
I sincerely appreciate your help!

djmitche:
That's fantastic! Let me know if I can point you toward others. The Rust world has a lot of good projects with communities that are helpful and welcoming.

akashShanmugraj:
Thanks @djmitche ❤️!
I will let you know for sure. I guess I'll stick around here for sometime and watch out for some issues.

PR #3273, opened 2024-02-19
Use corrosion instead of CMakeRust

### LINKED ISSUE #2887: src/tc/rust doesn't correctly depend on taskchampion/*

### LINKED ISSUE #3272: Upgrade to a newer CMake
We currently only require CMake 3.0. It seems most CI systems have CMake 3.16 or higher. But a bunch of interesting functionality is only available in even later versions, such as FetchContent.

Let's consider where we can find a balance of compatibility and new functionality.

djmitche:
CMakeRust appears to be unmaintained. Corrosion seems simpler -- it gets a lot of metadata from `cargo`. In particular, it seems to map crate dependencies into CMake, fixing #2887.

The downside is, this requires a newer CMake (3.22), and thus fixes #3272 as well. I don't expect this is going to do well with the GitHub actions..


djmitche:
@tbabej it appears the repo configuration requires the OpenSUSE 15 test, but this PR changes that to OpenSUSE tumbleweed. Can you update the repo settings?

djmitche:
HT to @tylerjw for https://tylerjw.dev/posts/rust-cpp-interop/ and https://tylerjw.dev/posts/rust-cmake-interop-cmake/ which led me to Corrosion (and has me thinking about some other changes to how we do interop).

aktaboot:
https://github.com/YosysHQ/nextpnr another similar example of interop with corrosion

.gitignore
```diff
@@ -13,3 +13,4 @@ tutorials
 .prove
 /target/
 /.idea/
+cargo
```
ryneeverett: I'm not seeing a "cargo" file showing up anywhere.

ryneeverett (approved):
This is a fantastic improvement to developer quality of life. This indeed fixes #2887 and from my perspective goes beyond that -- cmake wouldn't even use the latest taskchampion if you built it manually before and now you can make a change in taskchampion and cmake will rebuild taskchampion without the developer even having to run cargo.

.gitignore
djmitche: Oh, probably a leftover from earlier in the experiment..

Issue #3294, opened 2024-03-26
cannot build task from release tarball task-3.0.0.tar.gz


fjaell:
$:/tmp> tar -xf task-3.0.0.tar.gz 
$:/tmp> cmake -S task-3.0.0 -B build
-- Configuring C++17
-- System: Linux
-- Looking for libshared
-- Found libshared
-- Looking for SHA1 references
-- Looking for libuuid
-- Found libuuid
-- Configuring cmake.h
-- Rust Target: x86_64-unknown-linux-gnu
-- Using Corrosion as a subdirectory
error: failed to load manifest for workspace member `/tmp/task-3.0.0/taskchampion/integration-tests`

Caused by:
  failed to read `/tmp/task-3.0.0/taskchampion/integration-tests/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
CMake Error at /tmp/build/_deps/corrosion-src/cmake/CorrosionGenerator.cmake:24 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 101"

Call Stack (most recent call first):
  /tmp/build/_deps/corrosion-src/cmake/CorrosionGenerator.cmake:217 (_cargo_metadata)
  /tmp/build/_deps/corrosion-src/cmake/Corrosion.cmake:1098 (_generator_add_cargo_targets)
  src/tc/CMakeLists.txt:11 (corrosion_import_crate)

-- Configuring incomplete, errors occurred!
$:/tmp>

djmitche:
Thanks! That definitely seems like a [bug](https://github.com/corrosion-rs/corrosion/issues/494) in Corrosion.

@jschleus mentioned that just removing the offending line, pointing to `taskchampion/integration-tests`, from `Cargo.toml` fixed the issue, so perhaps that's a good workaround. I'm not sure how to actually solve the issue.

djmitche:
It looks like the fix to this is to just edit the `Cargo.toml` during the release process. I'll make a note of this in the release instructions, so the next patch release will hopefully work better!

djmitche:
I suspect #3209 is the long-term fix here, by the way.

Theblackcat98:
After following @djmitche instructions (**removing** line 7 of file task-3.0.0/Cargo.toml)
And **creating** a "/build" folder outside of task-3.0.0.
I was successfully able to run:
`cmake -S task-3.0.0 -B build`
and then finally:
`cd build/` & `make install`

---
[...]

PR #3302, opened 2024-03-27
Edit `Cargo.toml` during release

djmitche:
This should ensure that the next release builds without a manual workaround, as described in #3294.


huantrinh1802:
Hi @djmitche, would this PR #3302 also address the issue in https://github.com/Homebrew/homebrew-core/pull/167082?

djmitche:
Hi @huantrinh1802 - this PR just changes the release instructions, which I would not expect homebrew maintainers to read.

I expect a Taskwarrior 3.0.1 will have a fix to this (in the sense that I will follow these instructions!)

PR #3315, opened 2024-03-30
use CMake project settings

felixschurk:
This adds a description as well as the homepage to the CMake settings. Further it would also now use the numbering scheme as supposed to in CMake, this way other people could now pin a specific version if taskwarrior is included in another project.
Documentation from CMake is https://cmake.org/cmake/help/latest/command/project.html

```sh
❯ ./problems
/home/felix/data/taskwarrior/build-test/test/./problems:55: SyntaxWarning: invalid escape sequence '\S'
  file = re.compile("^# (?:./)?(\S+\.t)(?:\.exe)?$")
/home/felix/data/taskwarrior/build-test/test/./problems:56: SyntaxWarning: invalid escape sequence '\d'
  timestamp = re.compile("^# (\d+(?:\.\d+)?) ==>.*$")
Failed:                        

Unexpected successes:          

Skipped:                       

Expected failures:             
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,10 @@
 cmake_minimum_required (VERSION 3.22)
+
+project (task
+  VERSION 3.0.0
+  DESCRIPTION "Taskwarrior - a command-line TODO list manager"
+  HOMEPAGE_URL https://taskwarrior.org/)
+
 set (CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 
 include (FetchContent)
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Looks great - thanks!

I'm definitely a newcomer to CMake and haven't learned anything more than the bare minimum to get things to work. Until very recently, TW was using an ancient version of CMake, too. So I imagine there are lots of examples of "doing it the old way". Thanks for removing this one!

djmitche:
```
/home/felix/data/taskwarrior/build-test/test/./problems:55: SyntaxWarning: invalid escape sequence '\S'
  file = re.compile("^# (?:./)?(\S+\.t)(?:\.exe)?$")
/home/felix/data/taskwarrior/build-test/test/./problems:56: SyntaxWarning: invalid escape sequence '\d'
  timestamp = re.compile("^# (\d+(?:\.\d+)?) ==>.*$")
  ```
I wonder if this is from a new version of Python? It looks like the tests passed anyway (and are passing in GitHub Actions),

PR #3319, opened 2024-03-31
Remove 'sync' from default verbose flags

### LINKED ISSUE #3297: 3.x always warns that sync is required
From @jschleus in #3295:
> the new task command outputs always "There are XX local changes.  Sync required.".

djmitche:
Fixes #3297 (except for folks who have an explicit "verbose" in their config, but they can adjust)

### DIFF
--- a/src/feedback.cpp
+++ b/src/feedback.cpp
@@ -207,6 +207,11 @@ void feedback_unblocked (const Task& task)
 ///////////////////////////////////////////////////////////////////////////////
 void feedback_backlog ()
 {
+  // If non-local sync is not set up, do not provide this feedback.
+  if (Context::getContext ().config.get ("sync.encryption_secret") == "") {
+    return;
+  }
+
   if (Context::getContext ().verbose ("sync"))
   {
     int count = Context::getContext ().tdb2.num_local_changes ();


ryneeverett (approved):
I hadn't realized that `verbose=on` is the default which makes this a tougher decision -- ideally sync messages would be included in verbose output and the default would be off.

Maybe we should add to the synchronization documentation a recommendation to enable sync output in configuration? Any user using the sync support should probably have this enabled, right?

ryneeverett:
In fact, maybe this output could be automatically included if sync options are configured?

djmitche:
I don't actually think this particular message is that useful -- I've certainly turned it off locally, and I do use sync (but do so on a crontask).

But, including it when sync options are configured seems a good middle-ground.

djmitche:
Updated to only remind if non-local sync is set up.

ryneeverett (commented):
1. If you're using `sync.local.server_dir` don't you still want these messages?
2. Maybe the benefit of automatically making the smart choice here is worth it, but it seems like the logic here is liable to get out of sync (no pun) with the sync configuration options. As the sync options grow over time I don't see anyone thinking to update this logic. That's a non-issue if we manage this through the verbosity flags.

djmitche:
As I said earlier, I don't really see the value to this at all, but certainly not for local sync.

I suspect every "real" sync option will use an encryption secret, so this seems reasonably safe. I'm not sure what alternative you're suggesting -- the earlier draft of this PR did manage it through verbosity settings.

ryneeverett:
Isn't the point of this to remind people to run `task sync`? You still need to do that if you're using local sync.

ryneeverett:
I was suggesting that the earlier draft might have been advantageous over this one because it managed it through verbosity settings. If the value of these messages is questionable, turning them off completely unless users opt-in seems reasonable.

On the other hand, the current version seems to restore the tw-2 behavior of only giving these messages if sync settings are configured. If you prefer to take this more conservative approach I have no objection.

> In fact, maybe this output could be automatically included if sync options are configured?

To clarify, I now see that you've implemented this suggestion of mine. I'm now questioning whether it's really a good idea.

djmitche:
I think this was distracting for people using local sync (whether intentionally or not), but still useful for those setting up remote/cloud sync. Or to put it another way, your suggestion was a good one, and I agree.

PR #3321, opened 2024-03-31
Issue a warning if .data files remain

### LINKED ISSUE #3303: Warn if `.data` files exist
..it might make sense for `task` itself to indicate, on startup, that `.data` files still exist and suggest an export/import. Maybe that's too late to rescue the old version of `task`, but the same would be true if the warning was issued in `make install`.

_Originally posted by @djmitche in https://github.com/GothenburgBitFactory/taskwarrior/issues/3295#issuecomment-2024124391_

djmitche:
This will help users realize that they have updated to an incompatible version and must export and import their tasks.

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -62,9 +62,8 @@ void TDB2::open_replica (const std::string& location, bool create_if_missing)
     Color warning = Color (Context::getContext ().config.get ("color.warning"));
     std::cout << warning.colorize (
       format ("Found existing '.data' files in {1}", location)) << "\n";
-    std::cout << "  Taskwarrior's storage format changed in 3.0.\n";
+    std::cout << "  Taskwarrior's storage format changed in 3.0, requiring a manual migration.\n";
     std::cout << "  See https://github.com/GothenburgBitFactory/taskwarrior/releases.\n";
-    std::cout << "  Remove the files to silence this warning.\n\n";
   }
   replica = tc::Replica(location, create_if_missing);
 }
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -57,6 +57,15 @@ TDB2::TDB2 ()
 ////////////////////////////////////////////////////////////////////////////////
 void TDB2::open_replica (const std::string& location, bool create_if_missing)
 {
+  File pending_data = File (location + "/pending.data");
+  if (pending_data.exists()) {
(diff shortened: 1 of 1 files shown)


src/TDB2.cpp
```diff
@@ -57,6 +57,15 @@ TDB2::TDB2 ()
...
 void TDB2::open_replica (const std::string& location, bool create_if_missing)
 {
+  File pending_data = File (location + "/pending.data");
+  if (pending_data.exists()) {
+    Color warning = Color (Context::getContext ().config.get ("color.warning"));
+    std::cout << warning.colorize (
+      format ("Found existing '.data' files in {1}", location)) << "\n";
+    std::cout << "  Taskwarrior's storage format changed in 3.0.\n";
+    std::cout << "  See https://github.com/GothenburgBitFactory/taskwarrior/releases.\n";
+    std::cout << "  Remove the files to silence this warning.\n\n";
```
ryneeverett:
```suggestion
```

I wouldn't make this suggestion in a warning message because a user might delete these files without exporting first. Really they should probably move them elsewhere for a temporary backup but IMO just pointing them towards longer form documentation is all a warning message should aspire to do.

src/TDB2.cpp
```diff
@@ -57,6 +57,15 @@ TDB2::TDB2 ()
 ////////////////////////////////////////////////////////////////////////////////
 void TDB2::open_replica (const std::string& location, bool create_if_missing)
 {
+  File pending_data = File (location + "/pending.data");
+  if (pending_data.exists()) {
+    Color warning = Color (Context::getContext ().config.get ("color.warning"));
+    std::cout << warning.colorize (
+      format ("Found existing '.data' files in {1}", location)) << "\n";
+    std::cout << "  Taskwarrior's storage format changed in 3.0.\n";
```
ryneeverett: I'm not in love with this wording but I feel like some kind of "call to action" other than (implicitly) "delete your .data files" would be good.

```suggestion
    std::cout << "  Taskwarrior's storage format changed in 3.0, requiring a manual migration.\n";
```

ryneeverett (changes requested):

src/TDB2.cpp
djmitche: Thanks, good advice!

ryneeverett (approved):

PR #3342, opened 2024-04-06
Update `task news` to support 3.0.0

### LINKED ISSUE #3320: Update `task news` output

djmitche:
This updates the `task news` command to support newer versions than just 2.6.0.  It also updates the release instructions to suggest including news in new releases.

Fixes #3320.

### DIFF
--- a/src/commands/CmdNews.cpp
+++ b/src/commands/CmdNews.cpp
@@ -540,9 +540,9 @@ void NewsItem::version3_0_0 (std::vector<NewsItem>& items) {
     /*background=*/"",
     /*punchline=*/
     "The sync functionality for Taskwarrior has been rewritten entirely, and no longer\n"
-    "supports taskserver/taskd. Instead, the recommended solution is a cloud-storage backend,\n"
-    "although taskchampion-sync-server is also available. As part of this change, the\n"
-    "on-disk storage format has also changed.\n",
+    "supports taskserver/taskd. The most robust solution is a cloud-storage backend,\n"
+    "although a less-mature taskchampion-sync-server is also available. See `task-sync(5)`\n"
+    "For details. As part of this change, the on-disk storage format has also changed.\n",
     /*update=*/
     "This is a breaking upgrade: you must export your task database from 2.x and re-import\n"
     "it into 3.x. Hooks run during task import, so if you have any hooks defined,\n"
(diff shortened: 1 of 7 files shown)


djmitche:
Apologies for all the review requests @ryneeverett -- no good deed (such as reviewing previous PRs) goes unpunished (being asked to review more!)

src/commands/CmdNews.cpp
```diff
@@ -546,23 +606,27 @@ int CmdNews::execute (std::string& output)
...
-    std::cout << format ("\n({1}/{2}) ", i+1, items.size ());
-    items[i].render ();
+  if (items.empty ()) {
+    std::cout << bold.colorize ("You are up to date!\n");
+  } else {
+    // Print release notes
+    std::cout << bold.colorize (format (
+      "\n"
+      "============================================\n"
+      "Taskwarrior {1} - {2} Release highlights\n"
```
ryneeverett: I don't really have a better suggestion but I find this format confusing. Dashes are often used within version numbers which is why I don't typically see them used to denote ranges. I tried a few formulations using version specifier syntax (e.g., `>{1}, <={2}`) but couldn't find one that was a clear improvement.

You didn't introduce this but I also find it odd that "highlights" is not capitalized.

src/commands/CmdNews.cpp
```diff
@@ -546,23 +606,27 @@ int CmdNews::execute (std::string& output)
...
+      "============================================\n",
+      news_version,
+      current_version,
+      (full_summary ? "All" : (major_items ? "Major" : "Minor"))
```
ryneeverett: This `format` argument is no longer used and could be deleted.

src/commands/CmdNews.cpp
```diff
@@ -500,6 +530,28 @@ void CmdNews::version2_6_0 (std::vector<NewsItem>& items) {
...
+  NewsItem sync {
+    version,
+    /*major=*/true,
+    /*title=*/"New data model and sync backend",
+    /*bg_title=*/"",
+    /*background=*/"",
+    /*punchline=*/
+    "The sync functionality for Taskwarrior has been rewritten entirely, and no longer\n"
+    "supports taskserver/taskd. Instead, the recommended solution is a cloud-storage backend,\n"
+    "although taskchampion-sync-server is also available. As part of this change, the\n"
```
ryneeverett: As discussed elsewhere, I think you could reword this in a way that doesn't cause as much backlash by only saying *why* the cloud storage backend is better. Words like "robust" and "immature" are more to the point than "recommended". Recommendations are based on values and I would recommend people choose the sync solution that is best aligned with their own values.

ryneeverett (approved):

ryneeverett:
> Apologies for all the review requests @ryneeverett -- no good deed (such as reviewing previous PRs) goes unpunished (being asked to review more!)

No problem, I need to get better at reading C++ but usually lack the motivation without an explicit request. :)

src/commands/CmdNews.cpp
djmitche: I could replace it with "through".

ryneeverett: Only if you think that's better. Neither that nor what you currently have seem strictly accurate because the `news_version` release notes are actually not included, it's everything after `news_version`, right?

Another idea: `{1} -> {2}`.

djmitche: That's true! But I suspect users will forgive some loose use of range expressions :)

ryneeverett (approved):
Thanks!

Issue #3333, opened 2024-04-03
Replace invalid unicode with U+FFFD in `task import`


djmitche:
The `task export` from 2.6.2 just sort of blasts out the bytes it finds in `*.data`, regardless of whether those are valid utf-8 (and thus valid JSON). We can't change behavior of released code, but we can make `task import` a bit more forgiving, and translate those to U+FFFD (�) in the imported tasks, rather than crashing.

djmitche:
Blocked on https://github.com/GothenburgBitFactory/libshared/issues/77

PR #3343, opened 2024-04-06
Add test or current import behavior on invalid UTF-8

djmitche:
This test validates the behavior seen in #3333. The fix will be in https://github.com/GothenburgBitFactory/libshared/issues/77.

### DIFF
--- a/test/basetest/utils.py
+++ b/test/basetest/utils.py
@@ -27,9 +27,9 @@ ON_POSIX = 'posix' in sys.builtin_module_names
 # Directory relative to basetest module location
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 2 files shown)


ryneeverett (commented):
Am I correct in understanding that the test will fail when the issue is fixed and that is intended to be a catalyst to update the test?

If that's the case I wonder if it wouldn't be better to write the failing test now and `@unittest.skip` it. At the very least add a comment or better yet message (`self.assertTrue(false, msg='blah blah blah')`) explaining that failure is a good thing.

djmitche:
Yeah, that's a good point. I'm not sure how (or if!) this will get solved. I think it's best to just abandon this PR -- it demonstrated that the bug exists, and that's really all it can do.

PR #3348, opened 2024-04-07
Install corrosion as submodule.

ryneeverett:
#### Description

This will enable nixpkgs -- and any other distribution that builds in a network sandbox and/or wants to use their own corrosion package rather than building another one -- to do so without patching taskwarrior.

Since we're already using submodules for libshared I don't think this should make the build process any more complicated for anyone else.

See
https://github.com/NixOS/nixpkgs/issues/300679#issuecomment-2041252688 for context.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "src/libshared"]
 	path = src/libshared
 	url = https://github.com/GothenburgBitFactory/libshared.git
+[submodule "src/tc/corrosion"]
+	path = src/tc/corrosion
+	url = https://github.com/corrosion-rs/corrosion.git
--- a/src/tc/CMakeLists.txt
+++ b/src/tc/CMakeLists.txt
@@ -1,11 +1,6 @@
 cmake_minimum_required (VERSION 3.22)
 
-FetchContent_Declare (
-    Corrosion
-    GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-    GIT_TAG v0.4.7
-)
-FetchContent_MakeAvailable(Corrosion)
+add_subdirectory(${CMAKE_SOURCE_DIR}/src/tc/corrosion)
 
 # Import taskchampion-lib as a CMake library.
 corrosion_import_crate(


ryneeverett:
BTW, I'm not strongly opinionated on this but it seemed like an option that probably wasn't considered. If the current way is more manageable for maintainers that's a higher priority.

djmitche (approved):
I think this makes sense -- at least to be consistent with libshared.

jcgruenhage:
Maybe I'm wrong with that assumption, but I would have thought most distributions will end up including corrosion directly from their repository and not from a git submodule? See https://corrosion-rs.github.io/corrosion/setup_corrosion.html#installation for the docs on that. Then, all you'd have to do in taskwarrior is `find_package(Corrosion REQUIRED)`, as with other dependencies.

I still think this change makes sense, in terms of consistency with libshared, but downstream distributions will probably be patching this bit anyway, and have already been, as seen under https://github.com/chimera-linux/cports/blob/master/user/taskwarrior/patches/0001-use-system-corrosion-instead-of-fetching-it-with-git.patch for example.

djmitche:
The advantage of FetchContent and submodules is, we can specify exactly the revision we want. Since we're new to corrosion, it's unclear how important that is, but at least for development that's one fewer variables that might lead a build to fail on one system and succeed on another.

I think we also want to optimize for the source package building for developers "out of the box" without having to customize the system environment too much. Requiring too much customization leads to solutions of "[never mind, just use this docker image](https://www.envoyproxy.io/docs/envoy/latest/start/building)". But that sort of customization is also precisely what distro maintainers want -- so there's some tension there.

This particular dependency is probably too peripheral to worry much about, and I think distros can patch out the build system as they have done here without too much trouble.

jcgruenhage:
Yep, fully agreed, I think this is a reasonable way to go about this. The only reason I wanted to comment on this thread at all is that I don't think this statement is accurate:

> This will enable nixpkgs -- and any other distribution that builds in a network sandbox and/or wants to use their own corrosion package rather than building another one -- to do so without patching taskwarrior.

This patch here does enable building taskwarrior in a network sandbox, but does not directly allow using their own corrosion package.

But as I said before: I still think that at least for the time being, the change as it is does make sense, as corrosion is fairly new and isn't even being shipped in Debian yet, so you can't rely on your devs even having it available in their distro of choice.

PR #3362, opened 2024-04-13
Throw std::strings on sync server errors

### LINKED ISSUE #3326: Invalid UUID given for `sync.server.client_id` results in unhelpful error
I am trying to setup the synchronization on the 3.0 version of taskwarrior. I don't want to rely on Google services for this and prefer the self-hosted solution. `man 5 task-sync` mentioned that I can use taskchampion sync server for this, so I set it up on my machine (version 0.4.1).

I have the following lines in my .taskrc:
```
sync.server.origin="http://$MYHOSTNAME"
sync.server.client_id="$MYUUID"
sync.encryption_secret="$MYSECRET"
```
However, when I run `task sync`, all I get is this: `Unknown error. Please report.`

`task diag` output:
```
task 3.0.0
   Platform: Linux

Compiler
    Version: 13.2.1 20230801
 Compliance: C++17

Build Features
     Commit: 3e41fb604
      CMake: 3.29.0
 Build type: 

Configuration
       File: /home/sergey/.taskrc (found), 2449 bytes, mode 100644
       Data: /home/sergey/.task (found), dir, mode 40755
    Locking: Enabled
         GC: Enabled
  rc.editor: nvim
Hooks
     System: Enabled
   Location: /home/sergey/.task/hooks
     Active: on-modify.timewarrior (executable)
   Inactive: 

Tests
   Terminal: 120x30
       Dups: Scanned 1723 tasks for duplicate UUIDs:
             No duplicates found
 Broken ref: Scanned 1723 tasks for broken references:
...
```

djmitche:
Throwing `const char[]` is not the same! Confirmed experimentally that this fixes #3326.


src/commands/CmdSync.cpp
```diff
@@ -24,6 +24,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
+#include <iostream>
```
ryneeverett: Is this necessary?

ryneeverett (approved):
It's unfortunate that there doesn't seem to be any infrastructure to test or validate configuration but that's a problem for another day.

src/commands/CmdSync.cpp
djmitche: Nope, it was for debugging!

djmitche:
> It's unfortunate that there doesn't seem to be any infrastructure to test or validate configuration but that's a problem for another day.

Agreed. I'm also a little sad about the lack of enforcement of exception types -- how many other places are we throwing things that aren't `std::string`? `std::string_view`?

Issue #3338, opened 2024-04-05
Start using clang-format to format C++ code


djmitche:
The C++ style and formatting in Taskwarrior is unusual, and difficult to keep up with by hand or in reviews. Let's let computers do the work for us! Ideally clangd-format can be configured to enforce roughly the style used in Taskwarrior, then run once over the codebase and enforced in CI thereafter.

This should make it easy for contributors to bang together some C++, run `clangd-format` on it, and then make a PR, so we don't need to make review comments about the formatting and also maintain a consistent formatting for the project.

ryneeverett:
Nitpick: isn't it [clang-format](https://clang.llvm.org/docs/ClangFormat.html)?

Perhaps a separate issue but what would we think about checking in a compile_commands.json file so [clangd](https://clangd.llvm.org/) works out of the box?

djmitche:
Oh, good point regarding the spelling!

CMake can actually create `compile_commands.json` with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` - it's working for me! Maybe documenting that somewhere would be good?

felixschurk:
I looked into it already a little bit. 
As far as my search suggested we can setup `clang-format` quite simple with providing the necessary file in the repository. 
However, in order to have it useful one would need to automate the process of running clang-format, as not every developer is using an IDE, or we do not even know which IDE's are used.

Searching a bit deeper into this direction [Stack Overflow Question](https://stackoverflow.com/questions/55965712/how-do-i-add-clang-formatting-to-pre-commit-hook), I stumpled over the [pre-commit](https://pre-commit.com/) framework, which also provides a [pre-commit.ci](https://pre-commit.ci/) service which would allow us to use e.g. `clang-format` for the C, C++ code, [black](https://github.com/psf/black) for Python files as well as maybe some other linter for the Rust code. 

Is this a way we would like to go here?

djmitche:
Yes, I think we will want to enforce this on PRs. 

Two important considerations: can the tool be configured to the particular style of Taskwarrior, and do we need to make a huge PR to bring the whole codebase into compliance?

felixschurk:
Yes and yes.

As far as I know one can configure `clang-format` to be as one likes, however I assume as described in #1987 already pointed out we are quite close to Google Coding Style. 

[...]

PR #3370, opened 2024-04-16
Recommend LSP's in development docs

ryneeverett:
Per conversation in #3338.

There are already a lot of documented compile options so I think we're better off suggesting that everybody create a compile_commands.json whether or not they're using an LSP because it doesn't cost much.

While I was at it it seemed reasonable to mention rust LSP too. Now that rls is deprecated I'm not sure there is any competitor to rust-analyzer worth mentioning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required (VERSION 3.22)
 
+set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 project (task
   VERSION 3.0.0
   DESCRIPTION "Taskwarrior - a command-line TODO list manager"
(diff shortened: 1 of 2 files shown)


felixschurk:
Would it maybe then not just make sense to set it in the CMake list, such that everybody gets them by default?
This would keep the compile commands a bit cleaner.

As you said they do not hinder anyone at anything, they only would help.

felixschurk:
It would be adding following line after the `cmake_minimum_required ...`

`set(CMAKE_EXPORT_COMPILE_COMMANDS ON)`
like that

```cmake
cmake_minimum_required (VERSION 3.22)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
project (task
...........
```

felixschurk:
Ah yes sorry that I did not mention.

Yes, it places it in the `build` directory as one could have multiple build directories with different `compile_command.json` settings e.g. different build systems, compiler or even with additional flags such as coverage etc.

djmitche (approved):
Nice, thanks! Rust analyzer works well with the TC code out of the box.

djmitche:
The rust failures are my fault

PR #3387, opened 2024-04-22
Fix test script paths

### LINKED ISSUE #3351: Test scripts assume `task` is at `src/task`.
Instead, `task` appears in the build directory. Ideally we could figure out from cmake where that is?

felixschurk:
#### Description
In this version, all the python scripts would get copied to the corresponding `build` directory, where then based on the CMAKE variables the paths are updated in `utils.py`.

Further, I appended `.py` to the python tests, to make them more distinguishable between the unit tests which reside in the same folder.

This would close #3351.

With the update to now also execute the integration tests and more in the `run_all` script it produces a lot of warning/errors etc. 

```sh
❯ ./problems
Failed:                        
variant_xor.t                      24

Unexpected successes:          

Skipped:                       
variant_xor.t                      31

Expected failures:             
variant_xor.t                       8
```


ryneeverett:
> Further, I appended .py to the python tests, to make them more distinguishable between the unit tests which reside in the same folder.

So the python tests would have a different extension in the build folder than in the source? I'd question whether that really makes things easier to follow.

What is the advantage of moving the tests into the build folder versus `s/src\/task/task/g` and manipulating the `PATH`?

test/basetest/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+configure_file(__init__.py __init__.py COPYONLY)
+configure_file(compat.py compat.py COPYONLY)
+configure_file(exceptions.py exceptions.py COPYONLY)
+configure_file(hooks.py hooks.py COPYONLY)
+configure_file(meta.py meta.py COPYONLY)
+configure_file(task.py task.py COPYONLY)
+configure_file(testing.py testing.py COPYONLY)
+configure_file(utils.py utils.py)
```
djmitche, line 8: I guess it's critical that COPYONLY doesn't appear here, as that's how the `${..}` get a value in this file?

djmitche (commented):
I think it makes sense to move the Python tests to the build directory, as that's where the built binaries from the C++ tests appear.

I agree that renaming the tests in that process is counter-intuitive. But, I'd also mildly prefer that these files were named `.py` in the source code, too -- several times I have opened `foo.t` in my editor only to realize it's a built binary. Although now that we use a build directory I suppose this isn't an issue.

One concern I have with this plan is that I think `configure_file` only copies files when `cmake` is run, and not on `make`. There's a similar issue with the manpages. It means that it's easy to edit a test file and run `make test`, only to find out that it used the old file contents. Is there a better solution that could make e.g., `build/test/foo.py` a dependency of `test/foo.py` so that an update to the latter will be copied to the former when running `make`?

felixschurk:
> I agree that renaming the tests in that process is counter-intuitive. But, I'd also mildly prefer that these files were named `.py` in the source code, too -- several times I have opened `foo.t` in my editor only to realize it's a built binary. Although now that we use a build directory I suppose this isn't an issue.

Yes I agree, it is quite counter-intuitive renaming them only when moving them to the build folder. However I would also prefer to have the source files with the proper ending.. `*.t.py` for python tests, `*.t.sh` for bash tests and `*.t.cpp` for unit tests.
One can handle that in a separate issue. 


> One concern I have with this plan is that I think `configure_file` only copies files when `cmake` is run, and not on `make`. There's a similar issue with the manpages. It means that it's easy to edit a test file and run `make test`, only to find out that it used the old file contents. Is there a better solution that could make e.g., `build/test/foo.py` a dependency of `test/foo.py` so that an update to the latter will be copied to the former when running `make`?

When I modify a test file in `./test`, e.g. adding something and then run `make test` either in the `build` folder or `cmake --build build -t test` in the root directory of taskwarrior. 
It does completely rerun the configure as well as build process of cmake and updates the corresponding files also in the `build` directory. 

@djmitche Could you check if we mean the same thing? Not that we would miss it.

test/basetest/CMakeLists.txt
felixschurk: Exactly, that is the file where the variables are updated, which is not possible when `COPYONLY` is set.

felixschurk:
> What is the advantage of moving the tests into the build folder versus `s/src\/task/task/g` and manipulating the `PATH`?

I think also here the problem is that per-se the build folder can have completely random names, eg. Clion does name the build folder by default something like `cmake-build-relase`, where VS Code names it `build`. 
Personally, I sometimes name them based on the compiler `build-clang` or `build-gcc`.

Now only on the specific invocation for that build folder (as one can multiple build folder) we do know the proper location of the source file.

I also first tried with only moving the `utils.py` file (as this is the only one where something is changed), however then I would still need to update the python scripts in the source directory to find the appropriate `utils.py` file in the corresponding build directory (where one then no longer knows how it is named).
Also the source tree is under version control, so when one would update there the path it would always show up as a change in git.

felixschurk:
What I currently not understand is the `macos` tests, here in GitHub Actions they are skipped and I assume that I can not run them locally as I am not on macos. 

Do we use/run the test cases somewhere?

ryneeverett:
See #3242.

djmitche:
Indeed, I do see
```
-- Build files have been written to: /home/dustin/p/taskwarrior/build
```
when I modify a Python script and then run `make -Cbuild test`.

So, I think this is a strict improvement, moreso if the rename to `.py` is removed.

felixschurk:
Alright, I removed the extension and the corresponding glob pattern. 
I still think it might make sense to generally rename the tests according to their type, but thats a new issue. As currently the editor is not able to recognize that it is a python file, and thus thinks its a perl script or so.

There are failing test cases, but I do not know if they are ones not intended to be run or so. Currently, all python files are getting copied over and thus also executed. Here I would need to have some feedback which ones are expected to fail/not be executed or if some are broken due to the setup etc.

ryneeverett:
> I still think it might make sense to generally rename the tests according to their type, but thats a new issue. As currently the editor is not able to recognize that it is a python file, and thus thinks its a perl script or so.

:+1: to changing the extensions in the source (rather than just the build)

test/run_all
```diff
@@ -70,7 +70,7 @@ class TestRunner(object):
         self._outputq = Queue()
 
     def _find_tests(self):
-        for test in glob.glob("*.t") + glob.glob("*.t.exe"):
+        for test in glob.glob("*.t")
```
ryneeverett:
```suggestion
        for test in glob.glob("*.t"):
```

ryneeverett:
> I think it makes sense to move the Python tests to the build directory, as that's where the built binaries from the C++ tests appear.

I kind of get the symmetry argument but unless there is a less verbose and error-prone way to copy the files than to name each one I don't think it should prevail. It seems like a reasonable choice to run python from the source directory because it doesn't need to be built.

test/run_all
felixschurk: Fixes it in 21cfb87342435d15630c6c2dc82ab875cf78f19e

felixschurk:
> > I think it makes sense to move the Python tests to the build directory, as that's where the built binaries from the C++ tests appear.
> 
> I kind of get the symmetry argument but unless there is a less verbose and error-prone way to copy the files than to name each one I don't think it should prevail. It seems like a reasonable choice to run python from the source directory because it doesn't need to be built.

I do not really see a verbose copy (listing in the CMakeLists.txt) as an error-prone way of having an overview of which test we want to run. I mean in the end it goes more towards the direction that every executable is treated as a kind of target in CMake, similar as it is done with the [CPP unit tests](https://github.com/GothenburgBitFactory/taskwarrior/blob/eaef05ee95934e74db0f19dd1470c8424b17962c/test/CMakeLists.txt#L21).

Having them explicitly listed provides in my opinion a straight forward way in seeing which test are really executed etc., especially compared to the current approach relying on permissions of the files, that one can not even see now on GitHub. I mean looking at the `CMakeLists.txt` everyone can directly see which tests which be executed or are part of the current test suite. Looking at the [test folder](https://github.com/GothenburgBitFactory/taskwarrior/tree/eaef05ee95934e74db0f19dd1470c8424b17962c/test) I do not know how one is figuring out that e.g. [tw-1837.t](https://github.com/GothenburgBitFactory/taskwarrior/blob/eaef05ee95934e74db0f19dd1470c8424b17962c/test/tw-1837.t) is not executable and thus will not be run, without cloning the repo, checking out the PR/branch and checking for permissions.


I do get the argument that one could run the tests from the source files, but consider the case that you want to build e.g. with different compilers, different versions of taskwarrior or maybe some coverage build. Using the out of source build this can easily be done with different build folders e.g. `build-gcc`, `build-clang`, `build-debug`, `build-release`, `build-v3.0.1` and `build-v3.0.2`. As they are all in separate build directories they will all have a corresponding `task` executable. 
How would I now tell the python script in the source directory (simply checking out the corresponding branch for the corresponding builds) where it should look for the executable?
One solution would be to explicitly pass then the path to the `task` executable, e.g. `run_all ~/build-gcc/src/task` or `run_all ~/build-v3.0.1/src/task`. 
But then also comes the question, where should the output of the corresponding run go?

I mean that one can do this kind of different builds only got possible with the out-of-source build, so before it was not possible. But I do see quite some advantage in having the option to easily compare different executables or maybe even just build types.

ryneeverett:
> the current approach relying on permissions of the files

Ah, well it's easier to see explicitly listing the files as an improvement on *that*. Once the files are renamed to *.t.py* though, it seems like a better approach would be to go by the file extension.

> How would I now tell the python script in the source directory (simply checking out the corresponding branch for the corresponding builds) where it should look for the executable?

I'm not sure you understood my suggestion:

> What is the advantage of moving the tests into the build folder versus s/src\/task/task/g and manipulating the PATH?

I was suggesting prepending the correct `task` binary to the `PATH` environment variable at runtime and having the tests invoke `task` that way rather than through a full path to the binary.

Seems like this might make sense for the C++ tests too. Is there a reason they need to be rebuilt for each build folder?

Is part of the motivation here cross-compiling? If the idea is to have each build folder ready to ship to a target platform where it is tested then your approach makes a lot more sense to me.

felixschurk:
> Ah, well it's easier to see explicitly listing the files as an improvement on _that_. Once the files are renamed to _.t.py_ though, it seems like a better approach would be to go by the file extension.

I would not go with a globbing pattern based on the extension, as this will not retrigger the build process correctly. See [documentation from CMake](https://cmake.org/cmake/help/latest/command/file.html#filesystem) and a [StackOverflow discussion about pros and cons](https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak). 

> I was suggesting prepending the correct `task` binary to the `PATH` environment variable at runtime and having the tests invoke `task` that way rather than through a full path to the binary.

Hm yes I dont really understand how that would work. I mean how would I add the correct `task` binary to the PATH? Without dealing then with maybe duplications or aliasing existing other `task` executables?
If using the `task` which is on path, I think this is somehow supported with the `run_all` script (I do not find the corresponding option right now but I saw it in some file.)

Could you maybe elaborate your thoughts? Especially the part with how to add it also to the path I do not really get. (I mean I know how to add programs/folders etc. to the path, but how would it behave when on the system already a `task` is on path etc.)

> Seems like this might make sense for the C++ tests too. Is there a reason they need to be rebuilt for each build folder?

They would need to rebuild as the build folder could correspond to different setups, e.g. `Release` vs. `Debug` or some other different flags, or even just be based on different branches of the repository.

> Is part of the motivation here cross-compiling? If the idea is to have each build folder ready to ship to a target platform where it is tested then your approach makes a lot more sense to me.

I never worked with cross-compiling so I do not know.

ryneeverett:
> I mean how would I add the correct `task` binary to the PATH? Without dealing then with maybe duplications or aliasing existing other `task` executables?

Shadowing binaries on PATH is simple -- the first match is used so you'd insert the path you want at the beginning.

However, I see now why this isn't a great option since the build folder is variable and we're now running the tests from the build folder. I don't see an obvious advantage to a variable build folder for taskwarrior but it seems like a deliberate design decision on cmake's part. (Presumably motivated by cross-compilation.)

Thanks for all the explanation here and elsewhere. I'm still adjusting to cmake myself but I trust that the change is for the best.

djmitche:
I'm also adjusting to CMake! Lists of files are a common thing, it seems -- maybe that's why the file is called `CMakeLists.txt` :)

I'm surprised to learn that tests must be executable, and I wonder if I've written test scripts that aren't ever run in CI! 

Looking at https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/8827600818/job/24235098558?pr=3387 it seems like the test failed for two tests in `template.t` -- the rest are "TODO". I see the same running locally with this PR applied. It looks like that file is a template for writing Python tests, so probably shouldn't run (and maybe should be removed entirely, as it's not findable for someone who would want such a template, and definitely not obviously a template for other test files even when one does find it.

djmitche:
I'd suggest either removing that file or marking it as not a test in CMakeLists.txt.

If there are other non-executable files that don't pass when made into tests, I think disabling them and filing an issue listing them is a good strategy to make progress.

felixschurk:
I was working on it at the same time right now and also came to the conclusion that the file `template.t` should not be added. :D

Actually for having there the nicer overview I played already around with CTest locally as mentioned in https://github.com/GothenburgBitFactory/taskwarrior/issues/3351#issuecomment-2063144077, but I will create a separate issue for that. 

Seams like now the test are no longer failing :)

djmitche (approved):

djmitche:
Great! Thanks to both of you!

PR #3389, opened 2024-04-22
Remove debug print

### LINKED ISSUE #3386: Version prepended to task lists
All commands that return a list of tasks prefix the output with the version number, so example the output of `task reports` is unchanged, but `task ls` now begins like this:

```
❯ task ls
3.0.1
ID D Project              Tags Due   Description
```

I noticed it starting in version 3.0.1 because it breaks the [shell integration](https://github.com/romkatv/powerlevel10k/issues/2648) I use, but I couldn't find mention of it in the release notes. Is this an intentional change?

I did a quick skim of the [release deltas](https://github.com/GothenburgBitFactory/taskwarrior/compare/v3.0.0...v3.0.1) but the location of the output didn't jump out at me immediately.

djmitche:
Fixes #3386.

I'm surprised this didn't cause test failures!

### DIFF
--- a/src/Version.cpp
+++ b/src/Version.cpp
@@ -95,7 +95,6 @@ bool Version::operator==(const Version &other) const {
 
 ////////////////////////////////////////////////////////////////////////////////
 bool Version::operator!=(const Version &other) const {
-  std::cout << other;
   return std::tie(major, minor, patch) !=
          std::tie(other.major, other.minor, other.patch);
 }


djmitche:
I think we should release 3.0.2 with this fix -- thoughts?

Bckempa:
Thank you for the quick fix!

> I think we should release 3.0.2 with this fix -- thoughts?

I've got a minor nit about how news works (that is, unless you know about `news all` you'll always be randomly bugged to (re-)read the news. Do you want to take a look at that for a 3.0.2 or is that not worth holding this back?

djmitche:
Please file that as an issue - I agree but it was clearly designed that way so I'm not sure we should change it.

I'd like to ship 3.0.2 as a bugfix for this serious bug.

ryneeverett (approved):

ryneeverett:
> I'm surprised this didn't cause test failures!

I suspect this is because the cli tests make heavy use of `assertIn` and so allow unlimited "extra" output. The most basic tests for each command ought to more strictly validate the output than they do.

PR #3407, opened 2024-04-27
Renaming test files according to their language

### LINKED ISSUE #3404: Renaming test files according to their language
> I agree that renaming the tests in that process is counter-intuitive. But, I'd also mildly prefer that these files were named `.py` in the source code, too -- several times I have opened `foo.t` in my editor only to realize it's a built binary. Although now that we use a build directory I suppose this isn't an issue.

Yes I agree, it is quite counter-intuitive renaming them only when moving them to the build folder. However I would also prefer to have the source files with the proper ending.. `*.t.py` for python tests, `*.t.sh` for bash tests and `*.t.cpp` for unit tests.
One can handle that in a separate issue. 

_Originally posted by @felixschurk in https://github.com/GothenburgBitFactory/taskwarrior/issues/3387#issuecomment-2071344693_

mattsmida:
#### Description

This PR updates the naming convention of the Python and shell  test files to be consistent with the one used by the C++ test files. E.g., `test_name.t` becomes `test_name.t.py` or `test_name.t.sh`.

#### Additional information

```
Failed:                        
Unexpected successes:          
Skipped:                       
Expected failures:             
```


test/run_all
```diff
@@ -70,7 +70,8 @@ class TestRunner(object):
         self._outputq = Queue()
 
     def _find_tests(self):
-        for test in glob.glob("*.t") + glob.glob("*.t.exe"):
+        for test in glob.glob("*.t") + glob.glob("*.t.exe") \
```
ryneeverett: is the first pattern still needed? Are there any remaining `*.t` files?

test/stress_test
```diff
@@ -13,7 +13,8 @@ import sys
 def find_tests():
     tests = []
 
-    for test in glob.glob("*.t") + glob.glob("*.t.exe"):
+    for test in glob.glob("*.t") + glob.glob("*.t.exe") \
```
ryneeverett: is the first pattern still needed? Are there any remaining `*.t` files?

ryneeverett (approved):

test/run_all
felixschurk: Yes, the `*.t` is currently still needed as the cpp unit tests (ending `*.t.cpp`) are created as targets with the ending `*.t`. 
See, this lines where `src_FILE` is set based on line 21:
https://github.com/GothenburgBitFactory/taskwarrior/blob/eaef05ee95934e74db0f19dd1470c8424b17962c/test/CMakeLists.txt#L30-L36

Currently the `*t.exe` is not in use. 
Same applies for https://github.com/GothenburgBitFactory/taskwarrior/pull/3407#discussion_r1581716066.

felixschurk:
I do not see the need for having the `test_namer` folder in the repository. 
As if then by now when someone creates a new test one should adhere to the coding convention.

felixschurk:
Other than that it does look good so far. But on another notice, currently the python tests are not run, see https://github.com/GothenburgBitFactory/taskwarrior/issues/3351#issuecomment-2043664761.

Meaning currently the output of `./problems` is not reliable or only counts for the unit tests.

djmitche:
I will leave this to the two of you for review - please @ me if you need the "merge" button clicked!

test/run_all
mattsmida, line 1: @felixschurk If it makes sense to include the `.test.cpp` files here, we can include them too. The reason for the omission is that none of the C++ files are marked as executable.

test/stress_test
mattsmida, line 1: @felixschurk Same comment as in `tests/run_all`.

mattsmida:
Thank you @felixschurk and @ryneeverett for the feedback. I have made the requested changes:
- `test-namer` directory is gone
- Now using `*.test.*` instead of `*.t.*`
- Updated documentation and supporting orchestration scripts

felixschurk:
It does look good. :)

Due to the update in #3387, we also do need to update the CMakeLists etc. This was at the state where you forked not yet implemented. 
I did not want you to go thru that hassle and therefore rebased and updated it, however I am unable to push to this PR directly. 

Specifically I mean https://github.com/felixschurk/taskwarrior/commit/f8f72a71769e8851e76b0d7907dc2e91ceee6e30 and https://github.com/felixschurk/taskwarrior/commit/18802ecac0c529a5646c338911ec039fd02a3dc5, you can see in my fork of your fork https://github.com/felixschurk/taskwarrior/commits/rename-test-files-3404/. 

I think one is only able to push to a PR if it is allowed when creating the PR and one is a maintainer of the project, but that I do not know how it is here. @djmitche do you know why it might be not working?

Nevertheless, in order to keep going you could add my for as a remote, pull down the two commits and incorporate them into your branch which you can then push up again. Or you can also copy the changes into the files (might be the easiest).

test/run_all
felixschurk: Yes it does make sense to include them, see https://github.com/felixschurk/taskwarrior/commit/f8f72a71769e8851e76b0d7907dc2e91ceee6e30.

For explanaition :)

The reason why they are not marked as executable is that these are `cpp` unit tests, meaning they need to get compiled before they can get executed. This is done as setting them as a CMake target, as we use this as a meta build system (generates the build files such that the compiler knows what to do).
These files are added in, 
https://github.com/GothenburgBitFactory/taskwarrior/blob/d093ce3d844939abebcbfb9f18f1342341f18bbf/test/CMakeLists.txt#L63
where the first argument of the function defines the name of the executable. 
Hence in our case now in the updated version `feature.test.cpp`, you do can find these files then in the `build` folder.

test/stress_test
felixschurk: Ah yes, should probably also be renamed. 
Unfortunately, I do not know how exactly the `stress_test` should work etc. 

In order to have it also in the build folder, one would add a line similar to https://github.com/GothenburgBitFactory/taskwarrior/blob/d093ce3d844939abebcbfb9f18f1342341f18bbf/test/CMakeLists.txt#L70
e.g. 
```cmake
configure_file(stress_test stress_test COPYONLY)
```
be added. Then it should also be able to execute it in the build folder.

Currently, this file is nowhere documented, only the commit https://github.com/GothenburgBitFactory/taskwarrior/commit/1371c9616a9beb67c34eb7e07cf12283cce865e8 gives some information about it. 

I do not know if we really need this test? Spoken from the point of view that it is not part of any CI etc. and was probably not really executed in the last couple of years.

felixschurk: @djmitche @ryneeverett 
Do you guys know more about the `stress_test` file? I mean what one can do with it is pretty clear, my question is mainly if it is really used?

ryneeverett:
@felixschurk You could make a pull request directly against @mattsmida's rename-test-files-3404 branch. IMO that's generally a better strategy than pushing directly.

*Edit:* Or maybe you can't and that's what you were saying. I tried but his repo doesn't seem to be a valid base repository in github UI. :man_shrugging:

felixschurk:
Yes, I think that would be a good thing. But I am also unable to create a PR against @mattsmida fork. 

But generally it seams to be available, e.g. I could create a PR against @ryneeverett's fork.

mattsmida:
Hi @felixschurk -- I believe this is ready now. The two commits https://github.com/felixschurk/taskwarrior/commit/f8f72a71769e8851e76b0d7907dc2e91ceee6e30 and https://github.com/felixschurk/taskwarrior/commit/18802ecac0c529a5646c338911ec039fd02a3dc5 have been incorporated into this PR. I did the "copy" method that you suggested in the last sentence of your comment: https://github.com/GothenburgBitFactory/taskwarrior/pull/3407#issuecomment-2083798120. 

(My apologies for configuring my fork such that PRs could not be created against it -- I think that was the default setting that I left unchanged when making the fork.)

felixschurk (approved):

felixschurk:
Hei, thank you a lot for the work. It does look good. 

Currently I am only a bit baffled as here the CI did not run, but I think that maybe a maintainer needs to start it manually. In my branch they did run sucessfully, so that should not be a problem.

@djmitche 
* [ ] Could you please run the CI?
* [x] Link the issue to this PR?

When it is successful, the PR is ready to be merged :)

This PR will close #3404. (Does not work unfortunately for linking it)

test/stress_test
djmitche: I don't know of a use for it. I think it can be dropped -- if we need it again, Git will remember.

test/CMakeLists.txt
```diff
@@ -37,8 +69,138 @@ endforeach (src_FILE)
 
 configure_file(run_all run_all COPYONLY)
 configure_file(problems problems COPYONLY)
+configure_file(bash_tap.sh bash_tap.sh COPYONLY)
+configure_file(bash_tap_tw.sh bash_tap_tw.sh COPYONLY)
+
+add_subdirectory(basetest)
+add_subdirectory(simpletap)
+
```
djmitche: These don't have `CMakeLists.txt` files in them, so `make` fails locally for me. Was that missed in `git add`?

djmitche (changes requested):
I linked the issue. And, I'm not sure why CI didn't run, but I'm guessing it's because there are merge conflicts? Once those are resolved I can click the buttons.

test/CMakeLists.txt
felixschurk: I checked it out, its not that they are missing in `git add`, but the develop branch is behind this develop branch. Therefore the files are missing.

felixschurk:
@mattsmida 
 As djmitche pointed out yes, there are the CMakeLists missing in the folders. 
 In order to have them best would be to update your develop, and rebase/merge the changes into this branch. 
Then the CI should run.

You can also verify it locally if it is building, in the current state it fails to generate the build files.

mattsmida:
@djmitche @felixschurk 
I followed Felix's advice:
> In order to have [the CMakeLists] best would be to update your develop, and rebase/merge the changes into this branch. Then the CI should run.

Now the CI is running and I'm seeing that there are no longer conflicts between this branch and the base branch. So I believe it's ready to merge. Thank you both for bearing with me!

djmitche (approved):
Looks great, works for me locally, can confirm that changing a test script correctly runs the changed script. 🚀

PR #3415, opened 2024-04-29
Include the whole error message in errors from Rust

### LINKED ISSUE #3411: Incomplete error messages on sync failure
In #3395, there's an error with sync, but only the context of the error appears, not the actual error:

https://github.com/GothenburgBitFactory/taskchampion/blob/5168f0b86415826aa84f70d1e072b6364dd431aa/taskchampion/src/replica.rs#L233

Let's figure out where that info is getting lost (my guess is, in the FFI layer) and fix it.

djmitche:
This is a bit of a frustrating design on anyhow's part!

I've verified by making `sync()` return an error that this does, indeed, include all of the details.

### DIFF
--- a/taskchampion/lib/src/util.rs
+++ b/taskchampion/lib/src/util.rs
@@ -1,7 +1,15 @@
 use crate::string::RustString;
 
-pub(crate) fn err_to_ruststring(e: impl std::string::ToString) -> RustString<'static> {
-    RustString::from(e.to_string())
+pub(crate) fn err_to_ruststring(e: anyhow::Error) -> RustString<'static> {
+    // The default `to_string` representation of `anyhow::Error` only shows the "outermost"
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
I'm assume there are already newlines in the context? Seems like it wouldn't be very readable otherwise...but it's an improvement no matter how badly it's formatted.

djmitche:
This is sort of following the Go system of error composition:
```go
foo, err := frobnicate()
if err != nil {
  return nil, fmt.Errorf("Could not frobnicate: %w", err)
}
```
which prepends a few words and `:` for each layer of error handling. So, hopefully newlines won't be required, although the error may get beyond 80 characters. In my testing, it was something like "Failed to synchronize with server: Sync Failed: FakeError".

PR #3421, opened 2024-04-30
remove .gitignore files and symbolic links/aliases

### LINKED ISSUE #3420: Remove obsolete .gitignore files and patterns
It mainly came to my mind that also some of our `.gitignore` files are now basically obsolete, as we do not need to ignore build files etc. 
This would also then give a developer which uses an in-source build, a feedback that something is not right. Feedback in the sense a lot of untracked files in git, which should be suspicious I assume.

When I was working over #3271. I just knew from the documentation that out-of-source is preferred by CMake and was always following it, now I started to realize a why that makes sense. One does not need to ignore certain files in the source and thus can more easily also build different versions of the application or test things out.

felixschurk:
* remove symbollic links in the src directory as they are no longer working with the out-of-source build
* remove .gitignore in the documentation (is build in build folder not needed)
* remove CMake folders as they are also no longer present in the source directory

This would close #3420.

```sh
❯ ./problems
Failed:                        
calendar.t                          2

Unexpected successes:          

Skipped:                       
datetime-negative.t                 9

Expected failures:             
undo.t                              2
```


djmitche (approved):
Nice cleanup!

Issue #3414, opened 2024-04-28
Allow URLs with paths in sync server config


djmitche:
From #3361. @davidhelbig did the Taskchampion part of this in https://github.com/GothenburgBitFactory/taskchampion/issues/367. Once #3209 is done, the remainder needs to be done in this repo:

>     * changing the name of the `sync.server.origin` config to `sync.server.url`, but keeping backward compatibility

djmitche:
Fixed in #3423.

Issue #3209, opened 2023-11-23
Move taskchampion to its own project and depend on it in Taskwarrior


djmitche:
The idea here is to make Taskwarrior rely on Taskchampion as a dependency, rather than integrating the two into a single codebase.

This comes from GothenburgBitFactory/taskwarrior#2887 and a few other issues with trying to intermingle Rust and C++ projects. The meaning of "project" is still vague -- is that another repo? Submodules, like with libshared?

We want to optimize for:
 * Ease of contribution to Taskchampion, using "normal" Rust stuff like `cargo`
 * Ease of use of Taskchampion by applications other than Taskwarrior
 * Conceptual clarity about the API boundary between TC and TW.

What I mean by that last point is that any change to Taskchampion's public API should be reviewed on its own merits, and not just added "because this Taskwarrior functionality needs it." For example, a `Task::get_ff4_string()` that returned the encoding of a task in FF4 format (the format used in the text-based `.task` files) wouldn't make sense, since nothing except Taskwarrior uses that format.

djmitche:
I think we can release Taskwarrior 3.0 before doing this, but I think it's important that this occur.

aktaboot:
mentioning #3187 for relevance's sake

aktaboot:
I strongly support the split of taskwarrior and taskchampion, but maybe a clearer list of what has to be done and what has to change would make things clearer.

- Where will taskchampion-lib, integration tests, xtask and sync-server end up ?
- Does some more splitting into repositories make sense ?

I am not yet very familiar with the code base, and probably not the right person to be suggesting this but I feel that further clarification on the split would be helpful.
Are there any issues that are blocking, or some necessary clarifications needed to advance on this issue ?

djmitche:
As you've seen, there are five crates within the Taskchampion workspace.
 - `taskchampion-lib` is closely tied to `taskchampion` so I think it must be in the same repo.
 - `xtask` is just development utilities (code generation) so should be in the same repo.
 - `sync-server` will, I think, become a reference implementation rather than anything especially "production ready". I suspect most users will be using the new cloud service sync, instead. Since it's an implementation of a specification, it would be OK to move it out of the repo. It also has a bunch of dependencies that are not shared with the other crates, meaning more to download and compile to run `cargo test` on the whole workspace.
[...]

PR #3423, opened 2024-05-01
Relax sync.server.origin to allow paths

chromy:
Hi! How would you feel about relaxing the restriction that `sync.server.origin` must be protocol + domain (https://tcsync.example.com) and allow paths also (https://example.com/tcsync/)?

This is useful for me since I run a bunch of services on a single server behind `nginx`:

```
proxy_set_header X-Forwarded-For $remote_addr;
location /calibre/ {
    proxy_buffering off;
    proxy_pass http://127.0.0.1:8080$request_uri;
}
location /calibre {
    # we need a trailing slash for the Application Cache to work
    rewrite /calibre /calibre/ permanent;
}

# Web UI for taskwarrior
location /tasks-chromy/ {
    proxy_pass http://127.0.0.1:8001$request_uri;
}

# root page
location / {
    root /usr/share/nginx/html;
    try_files /index.html =404;
}

# ...etc
```

Previously I ran `taskd` on its dedicated port (53589) which meant I had to configure this ports to be open etc which was always a little worrying. Now with the advent of `taskchampion-sync-server` I can very almost run it behind port 80/443:
```
location /taskchampion/ {
    proxy_pass http://127.0.0.1:7999/;
}
```
with exception of [this check](https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/taskchampion/taskchampion/src/server/sync/mod.rs#L34-L39). Just removing that was sufficient but I have beefed up the error handling a little and added some unittests while I was here.

I appreciate this is a little niche ...but running ones own `taskchampion-sync-server` instance is already fairly niche so maybe it won't so uncommon given the selection effect ;)

#### Description

Previously sync.server.origin had to be a scheme + FQDN, for example: https://tcsync.example.com
After this we support paths in sync.server.origin, for example: https://example.com/tcsync/
If a path is present it must end with a slash.

This allows for taskchampion-sync-server run behind e.g. nginx on port 80 / 443 at the same time as other services. This is desirable since:
- It avoids the need to choose between a. configuring additional non-standard ports or b. having dedicated HW/VMs/DNS routes just for taskchampion.
- Increases the flexibility of taskchampion.

Also:
- Error on some degenerate URLs (mailto:tcsync@example.com)
- Handle sync.server.origin = https://tcsync.example.com?foo=bar more correctly. Previously we geneated invalid URLs of the form: https://tcsync.example.com?foo=barv1/foo/bar.
- Add some unittests for the various cases
- Fix a typo impelementations -> implementations.


ryneeverett:
This appears to be implementing #3414.

djmitche:
> This appears to be implementing #3414.

It is! Awesome! @davidhelbig already landed the Taskchampion part of this in https://github.com/GothenburgBitFactory/taskchampion/issues/367, so what remains is #3209 (which will remove much of the taskchampion code modified here) and then the Taskwarrior part of this work, which is present in this PR.

So, if you don't mind waiting a bit for me to finish up #3209 and then rebasing this, that will finish the work! I've been trying to get to #3209 but honestly have been spending all my available time replying to issues. I have a few hours to myself tonight so hopefully can get it done during that time.

djmitche:
OK, #3209 has landed! We'll need to make another TaskChampion release to depend on from here, but it's late :)

However, I see that this PR doesn't change the name of the config option. An origin doesn't, by definition, have a path, and I think we should be consistent about that naming, as mentioned in #3414. One the new version of TaskChampion is out, would you be willing to make an update to TaskWarrior to use a new config name (with compatibility with the old)?

djmitche:
Some of the failing checks are because of #3485.

chromy:
Thanks! Rebased, now I think the failures are due to `taskwarrior` still being on TaskChampion v0.5 and the new changes not being released yet. I assume I have started to run the TaskChampion releasing checklist (slowly due to train internet) to see if anything will need fixing but I will only be able to get so far until some with ACLs for crates.io needs to step in.

src/commands/CmdSync.cpp
```diff
@@ -78,23 +83,28 @@ int CmdSync::execute (std::string& output)
     std::ostringstream os;
     os << "GCP bucket " << gcp_bucket;
     server_ident = os.str();
-  } else if (origin != "") {
+  } else if (server_url != "") {
```
djmitche, line 86: To make things easier for users, let's automatically convert `sync.server.origin` to a URL, if `sync.server.url` is not set.

djmitche:
https://docs.rs/taskchampion/0.6.0/taskchampion/index.html is released, with the origin -> url change!

src/commands/CmdSync.cpp
chromy: Sounds great! - but I'm not sure what you mean exactly by "convert sync.server.origin to a URL"?

I think both `sync.server.url` and `sync.server.origin` are strings in this context and I think I already made it so if `sync.server.url` is not set and `sync.server.origin` is set we use `sync.server.origin` as the 'URL'.

djmitche: I meant what you've done above on line 73:
```c++
std::string server_url = url == "" ? origin : url;
```
If that was already present when I made the comment, I apologize!

src/commands/CmdSync.cpp
```diff
@@ -78,23 +83,28 @@ int CmdSync::execute (std::string& output)
...
+    os << "Sync server at " << server_url;
     server_ident = os.str();
   } else {
     throw std::string ("No sync.* settings are configured. See task-sync(5).");
   }
 
   std::stringstream out;
-  if (Context::getContext ().verbose ("sync"))
+  if (origin != "") {
+    out << "sync.server.origin is deprecated use sync.server.url instead.\n";
```
djmitche:
```suggestion
    out << "sync.server.origin is deprecated. Use sync.server.url instead.\n";
```

djmitche (approved):
Looks good with a simple grammar fix!

PR #3424, opened 2024-05-01
Fix Python SyntaxError in tests/version.test.py

cclauss:
% `ruff check`
```
error: Failed to parse test/version.test.py:92:54: Expected a statement
Error: test/version.test.py:92:54: E999 SyntaxError: Expected a statement
```

#### Description

Replace this text with a description of the PR.

### DIFF
--- a/test/version.test.py
+++ b/test/version.test.py
@@ -77,7 +77,7 @@ class TestVersion(TestCase):
         self.assertIn("MIT license", out)
         self.assertIn("https://taskwarrior.org", out)
 
-    @unittest.expectedFailure
+    @unittest.skip(reason="See #3424")
     def test_under_version(self):
         """_version and diagnostics output expected version and syntax"""
         code, out, err = self.t("_version")
--- a/test/version.test.py
+++ b/test/version.test.py
@@ -89,7 +89,7 @@ class TestVersion(TestCase):
         if os.path.exists("../.git"):
           if 2 >= len(version) > 0:
               git = version[1]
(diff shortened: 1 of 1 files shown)


djmitche:
Pretty concerning that this didn't cause the test to fail!

Editing `run_all` to just run that file shows:
```
# version.test.py
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.25.8) or chardet (5.2.0) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
  File "/home/dustin/p/taskwarrior/build/test/version.test.py", line 92
    self.assertRegex(git, r'\([a-f0-9]*\)'))
                                           ^
SyntaxError: unmatched ')'
Passed:                             0
Expected failures:                  0
Runtime:                         0.37 seconds
```
so it would seem that a parse failure is ignored!

And it seems that this has been the case for a while -- this file now actually has test failures. @cclauss can you try to fix those up?

cclauss:
https://docs.python.org/3/library/unittest.html#unittest.skip

This will need to be fixed and the `skip()` removed.

cclauss:
Sorry...  I have pointed out the problem but I cannot propose a fix.

djmitche:
> Sorry... I have pointed out the problem but I cannot propose a fix.

That's totally fair!  I added a few more `skip(..)` to get the tests to pass.  We should be able to fix up the rest in #3430 .

djmitche (approved):

PR #3427, opened 2024-05-02
Remove taskchampion source from this repo

### LINKED ISSUE #3209: Move taskchampion to its own project and depend on it in Taskwarrior

djmitche:
This leaves `taskchampion-lib` more-or-less unchanged, but moves it to `src/tc/lib`. It also leaves `cargo xtask codegen` in place. I suspect that https://github.com/GothenburgBitFactory/taskchampion/issues/374 will basically remove the need for `taskchampion-lib` (and ffizz-header), so until that time I don't think we need to spend too much time on it. I'd like to get something like this landed quickly since we've currently got two copies of the `taskchampion` crate and that's already led to duplication (#3423).

I made a test release tarball with this and confirmed that it still builds.


doc/devel/contrib/rust-and-c++.md
```diff
@@ -1,6 +1,6 @@
 # Rust and C++
 
-Taskwarrior has historically been a C++ project, but as part of an [ongoing effort to replace its storage backend](https://github.com/GothenburgBitFactory/taskwarrior/issues/2770) it now includes a Rust library called TaskChampion.
+Taskwarrior has historically been a C++ project, but as part of an [effort to replace its storage backend](https://github.com/GothenburgBitFactory/taskwarrior/issues/2770) it now includes a Rust library called TaskChampion.
```
ryneeverett:
```suggestion
Taskwarrior has historically been a C++ project, but as of taskwarrior-3.0.0, the storage backend is now provided by a Rust library called TaskChampion.
```

I doubt the link is very helpful any more and the language doesn't say that the storage backend was actually replaced but suggests that this is just being attempted with an "effort".

doc/devel/contrib/rust-and-c++.md
```diff
@@ -1,6 +1,6 @@
...
 To develop Taskwarrior, you will need both a [C++ compiler and tools](./development.md), and a [Rust toolchain](https://www.rust-lang.org/tools/install).
 However, most tasks will only require you to be familiar with one of the two languages.
```
ryneeverett:
```suggestion
```

I don't think these sentences say anything that isn't either false or obvious at this point.

doc/devel/contrib/rust-and-c++.md
```diff
@@ -9,10 +9,9 @@ However, most tasks will only require you to be familiar with one of the two lan
...
 It defines an abstract model for this data, and also provides a simple Rust API for manipulating replicas.
 It also defines a method of synchronizing replicas and provides an implementation of that method in the form of a sync server.
-TaskChampion provides a C interface via the `taskchampion-lib` crate.
+TaskChampion provides a C interface via the `taskchampion-lib` crate, at `src/tc/lib`.
 
 Other applications, besides Taskwarrior, can use TaskChampion to manage tasks.
-Applications written in Rust can use the `taskchampion` crate, while those in other languages may use the `taskchampion-lib` crate.
 Taskwarrior is just one application using the TaskChampion interface.
 
 You can build TaskChampion locally by simply running `cargo build` in the root of this repository.
```
ryneeverett: Is this still true? Same for the line below referencing the ../../rust subdirectory.

ryneeverett (approved):
Would it make sense to edit or remove `.github/pull_request_template.md`? Seems like it was of questionable value anyway since we were mostly deleting it.

doc/devel/contrib/rust-and-c++.md
djmitche: Nope, it's not :)

djmitche:
Oh, I forgot about the pull request template. I'll make a new PR.

PR #3430, opened 2024-05-02
Treat a nonzero exit status as a failure

### LINKED ISSUE #3429: Test files that fail without output are not considered "failed"
See https://github.com/GothenburgBitFactory/taskwarrior/pull/3424

djmitche:
Fixes #3429. cc @cclauss!

@felixschurk what do you think? If this looks like the right fix, I can update the tests that now fail.

### DIFF
--- a/test/lexer.test.cpp
+++ b/test/lexer.test.cpp
@@ -37,7 +37,7 @@
 int main (int, char**)
 {
 #ifdef PRODUCT_TASKWARRIOR
-  UnitTest t (1259);
+  UnitTest t (1255);
 #else
   UnitTest t (1235);
 #endif
--- a/test/lexer.test.cpp
+++ b/test/lexer.test.cpp
@@ -37,7 +37,7 @@
 int main (int, char**)
 {
 #ifdef PRODUCT_TASKWARRIOR
-  UnitTest t (1253);
+  UnitTest t (1259);
 #else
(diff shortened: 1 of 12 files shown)


djmitche:
For my reference later:
```
Failed:                        
eval.test                           1
lexer.test                          1
t.test                              1
tdb2.test                           1
util.test                           1
version.test                        1
view.test                           1
```

test/lexer.test.cpp
```diff
@@ -37,7 +37,7 @@
 int main (int, char**)
 {
 #ifdef PRODUCT_TASKWARRIOR
-  UnitTest t (1253);
+  UnitTest t (1259);
```
ryneeverett: What is this about?

ryneeverett (approved):
Seems like the DIY test infrastructure has really been biting us lately and there are probably better solutions these days.

test/lexer.test.cpp
djmitche: Oh, I think I need to update that again. It sets the number of tests expected -- part of TAP.

djmitche:
> Seems like the DIY test infrastructure has really been biting us lately and there are probably better solutions these days.

Agreed, although thanks to @felixschurk and your work recently, it's a lot better than it has been. I'll add that to my list of things to talk about!

Issue #724, opened 2018-02-13
[TW-689] Out-of-source build broken


taskwarrior:
_Ben Boeckel on 2011-07-01T03:55:54Z says:_ 

 Patches attached for some minor CMake fixes to support out-of-source builds.

If you'd like CTest integration done as well, I can work on that. Also available, I also have a set of macros that will add targets for running valgrind and callgrind over the tests.

taskwarrior:
Migrated metadata:
```
Created: 2011-07-01T03:55:54Z
Modified: 2014-02-09T02:06:33Z
```

taskwarrior:
_Paul Beckingham on 2011-07-01T13:26:42Z says:_ 

 Patches applied.  Thank you Ben.  Your name is now in the AUTHORS file.

taskwarrior:
_Paul Beckingham on 2011-07-01T21:38:09Z says:_ 

 Thanks Ben.  CTest we don't want, but callgrind and valgrind woudl be realyl nice!  Thank you!

taskwarrior:
_Ben Boeckel on 2011-07-02T13:24:20Z says:_ 

 Why not CTest? If it doesn't do something that you want, it might be possible.

Currently my script works by adding a function which:

 * Adds a test, running the program
 * Adds a target for valgrind and callgrind running over the test
 * Adds targets for groups of tests

A new variant which doesn't add valgrind targets would have to be used for the perl-based tests.

A list of example targets with tests (at)include_a@ and (at)include_b@ grouped together would have:

 * (at)test@ --- runs all tests (the real CTest target)
 * (at)test-include@ --- runs all include tests
 * (at)test-include_a@ --- runs the (at)include_a@ test
 * (at)test-include_b@ --- runs the (at)include_b@ test
 * (at)valgrind@ --- runs all tests through valgrind
 * (at)valgrind-include@ --- runs all include tests through valgrind
 * (at)valgrind-include_a@ --- runs the (at)include_a@ test through valgrind
 * (at)valgrind-include_b@ --- runs the (at)include_b@ test through valgrind
 * (at)callgrind@ --- runs all tests through callgrind
 * (at)callgrind-include@ --- runs all include tests through callgrind
 * (at)callgrind-include_a@ --- runs the (at)include_a@ test through callgrind
 * (at)callgrind-include_b@ --- runs the (at)include_b@ test through callgrind

taskwarrior:
[...]

PR #3431, opened 2024-05-02
add targets for individual tests

djmitche:
This means you can do `make -Cbuild verbose.test` and it will run just that test.

cc @felixschurk - WDYT?

### DIFF
--- a/doc/devel/contrib/development.md
+++ b/doc/devel/contrib/development.md
@@ -56,7 +56,7 @@ First switch to the test directory:
 ```
 Then you can run all tests, showing details, with
 ```
-    $ make VERBOSE=1
+    $ make VERBOSE=1 test
 ```
 Alternately, run the tests with the details hidden in `all.log`:
 ```
@@ -65,6 +65,10 @@ Alternately, run the tests with the details hidden in `all.log`:
 Either way, you can get a summary of any test failures with:
 ```
     $ ./problems
+```
+You can run a single test suite, with source file `foo.test.cpp` or `foo.test.py`, with
+```
+    $ make foo.test
 ```
 
 Note that any development should be performed using a git clone, and the current development branch.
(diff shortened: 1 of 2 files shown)


ryneeverett (commented):
This feature doesn't seem very useful without documentation.

djmitche:
Good point - docs added!

ryneeverett (approved):

felixschurk:
Hei, sorry can not to deeply look into it right now as I am on the road. I will have a proper look on Tuesday. :)

But I do think if we want individual targets it's then better to go with ctest.

Which adds an individual target for every test in an more unique way (no need for custom commands).

I played around with it a few days ago already, one can see it in this branch https://github.com/felixschurk/taskwarrior/tree/ctest-as-test-runner.

Maybe check it out and run
```sh
cmake -S . -B build
cmake --build build
cmake --build build -t build_test
ctest --test-dir build/test
```
if I have it right this would then build and the second run the tests/all test.
Of course you can add the `-j <# cores> flag to both build and ctest commands to run it in parallel.
It will also show how/why the test failed (e.g. some fail with segmentation fault)

For a specific test one could do
```sh
ctest --test-dir build/test -R <name of test>
````
I prefixed the cpp test with `unittest_` and the python test with `integrationtest_`.


#724 was made a "decision" against ctest, however I think there is also a small misunderstanding about ctest, it does not replace a test suite - only the way how the suite is run, e.g. it would replace the `run_all` script.
So it also does not interfere with the TAP protocol or anything.
However it also gives the chance to have a proper output, choosable level of parallelisms and is a more standard way. Which will also be supported by basically every IDE and allows some handy features as returning just the failed test e.g. then with verbose output.
This would allow us also to have a cleaner test output in the CI as well as a better overview which test do what. Currently in my opinion that's a bit messy with the 4000 lines output of the `run_all -v`.

djmitche:
CTest sounds very appealing!  I don't think this PR will make implementing that any more difficult, and I've found it pretty useful in a few other PRs (and in fact have accidentally pushed a cherry-pick of this PR a few times now!) so I'll merge this.

Let's get an issue or PR filed to consider a switch to CTest.

felixschurk:
Yes sounds like a good plan to me, I will pick it up on Tuesday :)

PR #3433, opened 2024-05-02
Fix SyntaxWarning invalid escape sequence in Python code

### LINKED ISSUE #3388: Invalid escape sequences in tests
When running the python tests, warnings about invalid escape sequences, as in #3316 do appear all over the place. 

```sh
SyntaxWarning: invalid escape sequence
```

In order to fix them, one can prefix the corresponding string with an `r` to have it as a raw string.
Probably this corresponds with the used Python version, which is currently under Fedora 40 `Python 3.12.2`.

cclauss:
#### Description

These SyntaxWarnings are currently sprinkled throughout the GitHub Actions test results.

https://docs.astral.sh/ruff/rules/invalid-escape-sequence

% `ruff check --select=W605 --statistics`
```
471	W605	[*] Invalid escape sequence: `\ `
```
% `ruff check --select=W605 --fix`
```
Found 471 errors (471 fixed, 0 remaining).
```
% `ruff rule W605`
# invalid-escape-sequence (W605)

Derived from the **pycodestyle** linter.

Fix is always available.

## What it does
Checks for invalid escape sequences.

## Why is this bad?
Invalid escape sequences are deprecated in Python 3.6.

## Example
```python
regex = "\.png$"
```

Use instead:
```python
regex = r"\.png$"
```

Or, if the string already contains a valid escape sequence:
```python
value = "new line\nand invalid escape \_ here"
```

Use instead:
```python
value = "new line\nand invalid escape \\_ here"
```

## References
- [Python documentation: String and Bytes literals](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)

### DIFF
--- a/test/context.test.py
+++ b/test/context.test.py
@@ -204,7 +204,7 @@ class ContextManagementTest(TestCase):
         """Test that context is unset if its definition has been removed."""
         self.t('context define work project:Work', input='y\ny\n')
         self.t('context work')
-        code, out, err = self.t('context delete work', input='y\n\\y\n')
+        code, out, err = self.t('context delete work', input='y\ny\n')
         self.assertIn("Context 'work' deleted.", out)
 
         # Assert that taskrc does not countain context work definition
--- a/test/context.test.py
+++ b/test/context.test.py
@@ -204,7 +204,7 @@ class ContextManagementTest(TestCase):
         """Test that context is unset if its definition has been removed."""
         self.t('context define work project:Work', input='y\ny\n')
         self.t('context work')
-        code, out, err = self.t('context delete work', input='y\n\y\n')
+        code, out, err = self.t('context delete work', input='y\n\\y\n')
         self.assertIn("Context 'work' deleted.", out)
 
         # Assert that taskrc does not countain context work definition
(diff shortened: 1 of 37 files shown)


test/context.test.py
```diff
@@ -204,7 +204,7 @@ def test_context_delete_unset_after_removal(self):
         """Test that context is unset if its definition has been removed."""
         self.t('context define work project:Work', input='y\ny\n')
         self.t('context work')
-        code, out, err = self.t('context delete work', input='y\n\y\n')
+        code, out, err = self.t('context delete work', input='y\n\\y\n')
```
djmitche:
```suggestion
        code, out, err = self.t('context delete work', input='y\ny\n')
```
This is similar to line 205. Interesting that the `y` `\y` input also works!

test/quotes.test.py
```diff
@@ -125,7 +125,7 @@ def test_parser_ending_with_two_escape_slashes(self):
 
     def test_backslashes(self):
         """1436: Prove to the reader that backslashes are eaten twice (which means
-           \\ --> \) once by Python, and once more by some mystery process
+           \\ --> \\) once by Python, and once more by some mystery process
```
djmitche:
```suggestion
           two backslashes to one) once by Python, and once more by some mystery process
```

djmitche (approved):
Fantastic- thank you so much! I have two changes but they are very minor. I'll just commit them and merge.

PR #3439, opened 2024-05-03
Try to run the Python tests with pytest

cclauss:
## https://docs.pytest.org


ryneeverett:
Could you comment on the motivation here? I'm a fan pytest as a test runner for python projects, but what's the point if it can only run the python subset of taskwarrior's tests?

Issue #3395, opened 2024-04-23
3.0.0 sync with GCloud suddenly stopped working - "Failed to synchronize with server"


jcoffa:
Set up syncing with google cloud by following the steps outlined in `man task-sync` after upgrading from 2.6.2 -> 3.0.0. Verified that it was working and have been using this solution for syncing for a couple weeks now across 3 devices. Have performed successful syncs on all 3 of these devices over the past couple weeks. Everything was working as expected until today for some reason.

Attempted to run `task sync` as normal. Got an error message saying "Failed to synchronize with server". Fair enough, I usually forget to run `gcloud auth login` before attempting a sync anyways. Successfully authenticate in the browser window that opens using the same google account I've always used. Run `task sync` again, same generic error message "Failed to synchronize with server".

* I have not changed anything about my Taskwarrior config on this device or any other device since the initial sync setup
* Haven't modified anything in Google Cloud
* I only use Google Cloud to sync with Taskwarrior and have no other storage buckets, roles, or service accounts in this project other than the ones I made by following the instructions in `man task-sync`
* Highly doubt I've synced anywhere near enough to hit any quota limits in my free tier of Google Cloud; my task database is quite small (115 *total* tasks) and have synced less than a dozen times
* Verified that the JSON keys file is unmodified compared to when my syncing was last working, it's still in my `task` directory as expected, and the full absolute path + file name matches the value of my `sync.gcp.credential_path` taskrc property
* Verified that the correct project is selected via `gcloud config get-value project`
* Verified that the storage bucket still exists; I can still see all the task data in there:

![image](https://github.com/GothenburgBitFactory/taskwarrior/assets/35551680/5435f4f4-c9b4-4835-9a52-aaf0a4e1a4bf)

Output of `task diag`:

```
task 3.0.0
   Platform: Linux

Compiler
    Version: 13.2.1 20230801
       Caps: +stdc +stdc_hosted +LP64 +c8 +i32 +l64 +vp64 +time_t64
 Compliance: C++17

Build Features
     Commit: 3e41fb604
      CMake: 3.29.1
    libuuid: libuuid + uuid_unparse_lower
 Build type:

Configuration
       File: /home/joseph/.config/task/taskrc (found), 1890 bytes, mode 100644
       Data: /home/joseph/.config/task (found), dir, mode 40755
    Locking: Enabled
         GC: Enabled
    $VISUAL: nvim
Hooks
     System: Enabled
[...]

PR #3442, opened 2024-05-06
Update task-sync(5) to include delete perms for GCP sync

### LINKED ISSUE #3441: Include `storage.objects.delete` permission for GCP
Without this permission, things don't go well - #3395.

jcoffa:
Fixes #3441, as discovered via #3395 

`grep -R "storage.objects" .` revealed that only the man page for `task-sync(5)` mentions GCP role permissions so that is the only place it was added to.


djmitche (approved):
Thanks!

Issue #3416, opened 2024-04-29
Modification of task description by example on-add hook does not take effect


maertsen:
For an on-add hook I've developed myself, I don't see modifications to the description reflected in the db. I can reproduce this behaviour with the example on-add from https://taskwarrior.org/docs/hooks_guide/, which is the focus of this bug report. I modified the print statement to make it work with python 3.

* What command(s) did you run?

The debug command [from the example](https://taskwarrior.org/docs/hooks_guide/#debugging) (with an additional `+test` to make deletion easier):

```
$ task rc.debug.hooks=2 add +test -- Fix tw-98765
Timer Config::load (/Users/mya/.config/task/taskrc) 0.000895 sec
No context set
Found hook script /Users/mya/.config/task/hooks/on-add-bug-link.py
Found hook script /Users/mya/.config/task/hooks/on-add.message-id.py
  MODIFICATION tags <-- add 'test'
  MODIFICATION description <-- 'Fix tw-98765'
Hook: Calling /Users/mya/.config/task/hooks/on-add-bug-link.py
Hook: input
  {"description":"Fix tw-98765","entry":"20240429T132014Z","modified":"20240429T132014Z","status":"pending","uuid":"ffae184e-a938-4112-8e75-582e2cf9b86f","tags":["test"]}
Hooks: args
  api:2
  args:task rc.debug.hooks=2 add +test -- Fix tw-98765
  command:add
  rc:/Users/mya/.config/task/taskrc
  data:/Users/mya/Library/Application Support/task
  version:3.0.2
Timer Hooks::execute (/Users/mya/.config/task/hooks/on-add-bug-link.py) 0.033543 sec
Hook: output
  {"description": "Fix https://github.com/GothenburgBitFactory/taskwarrior/issues/tw-98765", "entry": "20240429T132014Z", "modified": "20240429T132014Z", "status": "pending", "uuid": "ffae184e-a938-4112-8e75-582e2cf9b86f", "tags": ["test"]}
  Link added
Hook: Completed with status 0
 
Hook: Calling /Users/mya/.config/task/hooks/on-add.message-id.py
Hook: input
  {"description": "Fix https://github.com/GothenburgBitFactory/taskwarrior/issues/tw-98765", "entry": "20240429T132014Z", "modified": "20240429T132014Z", "status": "pending", "uuid": "ffae184e-a938-4112-8e75-582e2cf9b86f", "tags": ["test"]}
Hooks: args
  api:2
[...]

PR #3443, opened 2024-05-06
Test modification of a task by an on-add hook (test case for #3416)

maertsen:
This PR adds a test case that shows the bug described in #3416. I confirmed it fails on current develop.

I think the bug's root cause is https://github.com/GothenburgBitFactory/taskwarrior/commit/5bb98579841d41e680280ada7bf3a3e9a9940553#r141727451. 

For a fix:
`TBD2::update()` as a method was removed in the referenced commit. And it seems calling `TBD2::modify()` instead will fire on-modify hooks when saving changes by on-add to the database. That's about as far as I got, I hope this contribution will help someone else make progress.

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -81,6 +81,9 @@ void TDB2::add (Task& task)
   std::string uuid = task.get ("uuid");
   changes[uuid] = task;
 
+  // run hooks for this new task
+  Context::getContext ().hooks.onAdd (task);
+
   auto innertask = replica.import_task_with_uuid (uuid);
 
   {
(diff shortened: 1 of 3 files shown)


maertsen:
Hmm, the simplest fix is reordering the on-add hook invocation before the save. That fixes the issue while passing the test suite. However, I don't know if it is the behaviour that you want. Anything more complicated is beyond my affinity with C++.

maertsen:
On a related note, https://github.com/GothenburgBitFactory/taskwarrior/pull/3443/commits/3e14d92a4272a48d981b5b3a3dbaa2286d01e2e5 is a copy of https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/scripts/hooks/on-add.the with some added comments.

djmitche (approved):
Sorry this took me so long to look at -- good test, good fix!

djmitche:
This also fixes #3455! Can you take a look at #3457 for me before I merge it?

Issue #3445, opened 2024-05-07
Change test runner to ctest


felixschurk:
## Some thoughts about testing/performance etc..
As a general regard for the tests and how we call it.
I thought a bit about it and for me personally it is quite confusing how one does run all test (unit tests, integration tests etc.) as well as how one could build only a specific test.
E.g. when I would like to work on a feature which affects only a subset of tests it would be beneficial to only run the corresponding tests when developing, and then have a simple interface to run all test when one would be ready.
Also on this approach I think most of the IDEs should be able to recognize the proper tests and then also coverage etc should be quite simple to add in the future.

## What could one do?
I guess the most forward way would be to use [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html) as it is bundled with CMake and thus already part of our toolchain. 
With CTest one can also execute shell scripts, python files etc. meaning we do not need to rework existing tests, only how they are executed/build with CMake.

## How would it look like? Some demo repository
I just searched and found [ctest-demo](https://github.com/tgamblin/ctest-demo/) as a simple sample repository (not affiliated with me or something). Maybe to have a feeling do:
```sh
git clone https://github.com/tgamblin/ctest-demo.git
cd ctest-demo
cmake -S . -B build
cmake --build build
ctest --test-dir build
```
This would give a first overview how it is looking/how it might be looking.
If one would like to see a bit bigger example [this repository](https://github.com/STCE-at-RWTH/AD-Driver-Generator/tree/develop) I worked on and we used there the CTest for the test invokation (Mainly liked because its a bit larger than the sample, but yet I am able to understand everything :D).

I played around already a bit with the taskwarrior repository, and I should be able to implement it here.

## What would need to be done?
* [x] adding all existing tests to the CMakeLists
* [x] update build pipelines to pickup new schema
* [x] update documentation for the developer
* [x] remove obsolute `run_all` and `problems` scripts
* [x] update documentation in `test/README`

@djmitche and @ryneeverett if you guys think it's a reasonable idea to go this direction, I would start working on it, probably with a separate issue.

_Originally posted by @felixschurk in https://github.com/GothenburgBitFactory/taskwarrior/issues/3351#issuecomment-2063144077_

PR #3446, opened 2024-05-07
add ctest as test driver

felixschurk:
* add `enable_testing()` in top level CMakeLists to discover tests
* add `add_test` for both unit tests as well as integration tests with prefixing the tests with the corresponding test type
* remove custom targets for tests
* remove unneeded warning from https://gitlab.kitware.com/cmake/cmake/-/issues/16062
* adds basic setup for resolving #3445 

Basically one can now invoke the whole test suite with running 
```sh
cmake -S . -B build
cmake --build build --target build_tests
ctest --test-dir build/test 
```
and one does has a quite decent output of the actual test runs.

Parallel running `-j 8`, running specific test cases `-R cpp` or `-R variant`, having minimal progress bar `--progress` as well as repeating tests until failure `--repeat-until-failure 100` is also possible with [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html).


ryneeverett (commented):
By switching from CMD to RUN for the test invocation in the docker images, the tests are now run when the images are *built* and nothing happens when they are *run*.

felixschurk:
Yes that is true, I realized that yesterday but could not fix it on the spot. My knowledge with Docker is limited. 

Would it maybe make sense to move the test invocation out into the `test.yaml` file?
I mean updating this line 
https://github.com/GothenburgBitFactory/taskwarrior/blob/60575a19675713fd5cc209c77733b8ffc3f442e5/.github/workflows/tests.yaml#L105
to 
```yaml
run: docker-compose run test-$CONTAINER ctest --test-dir build/test -j 8 --rerun-failed --output-on-failure
```

This way we would not have duplicated code in all the docker files, however it maybe would be a bit less/or maybe even more clear how we do run the tests.

ryneeverett:
In my opinion reducing the duplication is beneficial and passing the command into the `docker run` invocation is equally clear to having a `CMD` in the image.

felixschurk:
I am a bit sorry for all the failing CI jobs and commit updates, however locally testing it with one image/container it does run. So I am guessing there is a problem with escaping etc. in the `tests.yaml` file.

I had a similar issue already when setting up another CI. (locally working but in GH Actions not).

ryneeverett:
I don't think --entrypoint is what you want. It looks like you were close with 18888160c637e4f9df419475e03d229c4c2351e1. CI failed with "No such service: test-fedora40 ctest". Maybe you just need some quotes so it splits the arguments correctly?
```yml
        run: docker-compose run 'test-${{ env.CONTAINER }}' 'ctest --test-dir build/test -j 8 --output-on-failure --rerun-failed'
```

(Not 100% sure if it should be single or double quotes but I assume the template variables are substituted prior to evaluation and thus don't care.)

felixschurk:
Hm, yes definetely the splitting fixed one problem. 
I then detected some problem with a whitespace (which as somehow different encoded).

Now it does not find some file/directory. Let's see what the problem ist.

test/CMakeLists.txt
```diff
@@ -1,13 +1,5 @@
 cmake_minimum_required (VERSION 3.22)
 
-# This is a work-around for the following CMake issue:
-# https://gitlab.kitware.com/cmake/cmake/issues/16062
-# The issue has been fixed in CMake 3.11.0; the policy is set
-# to OLD for compatibility with older versions of CMake only.
-if(POLICY CMP0037 AND ${CMAKE_VERSION} VERSION_LESS "3.11.0")
-  cmake_policy(SET CMP0037 OLD)
-endif()
-
```
djmitche, line 10: Thanks for getting rid of this!

djmitche (commented):
This is great! I love the more succinct output. A few notes from my experience that don't need to be fixed before this is landed:

 * The `ctest` invocation, and similarly `make -Cbuild test`, don't seem to rebuild tests if their source files have changed, which can lead to surprises when a test run keeps failing after the test has been fixed. Maybe `cmake --build-and-test` helps with this?
 * The tests take a loong time to run. `ctest` takes a `-jN` option just like `make`, which might be a useful thing to recommend (although it did cause `ids.test.py` to fail for me!)
 * Having a documented way to run a single test will be useful, too.

One thing that does need to be addressed here, is updating the docs on how to run the tests (`doc/devel/contrib/development.md`).

felixschurk:
> In my opinion reducing the duplication is beneficial and passing the command into the `docker run` invocation is equally clear to having a `CMD` in the image.

I am sorry, I was not able to work that out. Maybe it makes sense to have this as a new issue (should be an easy one) when someone knows GitHub Actions and Docker.

felixschurk:
Currently, missing is the update of the [test/README](https://github.com/GothenburgBitFactory/taskwarrior/blob/2361521449bc02c9aa0be00dcc98a2d25aadb31b/test/README). 


I also realized now that basically the `--repeat` or respectively `--repeat-until-failure` options from `ctest` do work as a stress test, which we removed in #3447, therefore the functionality is still available, however now a bit more unified.

doc/devel/contrib/development.md
```diff
@@ -49,28 +49,47 @@ cmake --build build-clang
...
-    $ cd build/test
-```
-Then you can run all tests, showing details, with
-```
-    $ make VERBOSE=1 test
-```
-Alternately, run the tests with the details hidden in `all.log`:
+For running the test suite [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html) is used.
+Before one can run the test suite the `task_executable` must be built.
+After that also the `build_tests` executable must be build, which can be done over:
```
djmitche:
```suggestion
After that also the `build_tests` target must be build, which can be done over:
```

doc/devel/contrib/development.md
```diff
@@ -49,28 +49,47 @@ cmake --build build-clang
...
-Either way, you can get a summary of any test failures with:
+This would run all the test in serial and might take some time.
+
+### Running tests in parallel
+```sh
+ctest --test-dir build -j <number-of-jobs>
 ```
-    $ ./problems
+
+Further it is adviced to add the `--output-on-failure` option to `ctest`, to recieve a verbose output if a test is failing as well as the `--rerun-failed` flag, to invoke in subsequent runs only the failed ones.
```
djmitche, line 71: Nice!

djmitche (approved):
Looks good - let's merge it!

PR #3457, opened 2024-05-15
Add test that on-add returning 1 does not add task

djmitche:
This test existed, but didn't notice that the task was actually added. The bug itself was fixed in #3443.


maertsen:
The weird thing is that I actually wondered how rejection was supposed to work prior to #3443, but did not look if the test was complete. I haven’t run this, but the logic looks sounds to me.

djmitche:
Thanks for looking!

Issue #3465, opened 2024-05-27
Value limit=page not recognized.


Andonome:
When I run `task show`, I get the following message:

```
Your .taskrc file contains these unrecognized variables:
  limit

```
Checking the taskrc file, I have `limit=page`.

If I run `task list` without this, I get more tasks than can fit onto the terminal (`urxvtc`).
But when running with `limit=page`, I only get the top tasks, which can fit into the current terminal.

So `limit=page` has been recognized, and it's working as intended.  But then again, the vim highlights won't show this as a proper option, so maybe it's been forgotten about?

djmitche:
The `show` command's implementation has a long list of recognize config options. Maybe this was just omitted from the list? If so, a PR to restore it would be great (and a check that it appears in the manpage as well).

Andonome:
I've put in a [pull request](https://github.com/GothenburgBitFactory/taskwarrior/pull/3466).

djmitche:
Still need to add the config to `.taskrc` :)

Andonome:
The [PR](https://github.com/GothenburgBitFactory/taskwarrior/pull/3467) is in now ;)

djmitche:
Thanks for the fixes!

We'd love to have further help. Is there a particular thing you have in mind to improve? From your other repositories it seems you do a lot of technical (in the sense of very-detailed!) writing. I think the docs on https://taskwarrior.org/docs/ could use some updating. They certainly feel like they are the collected notes of a decade of developers, rather than a conscientious attempt to document an application. If you'd like to get more involved with coding, we can find something in that direction, too.

Andonome:
It's in the task list.

The task files are a little difficult to read. Would people appreciate having them in markdown format, then automatically translated with `lowdown`?  The [lowdown](https://kristaps.bsd.lv/lowdown) program allows easy translation to html, pdf, and various man page formats.

djmitche:
That sounds good -- I only _barely_ understand troff!

PR #3466, opened 2024-05-27
Let task show recognize limit in taskrc

Andonome:
This is just 2 line changes to recognize that `limit=page` (or `limit=n`) is allowable in the `taskrc` file.

Related to #3465.

I'm not a coder, but I read the instructions in the README.md files, and tested with `./src/task`, and it's working as expected.

The man page for `task` already shows how to use `limit=page`. Should this be added to `man taskrc`? If so, why are there two nearly-identical `taskrc` files at `doc/man`?


djmitche (approved):
Nice work - thanks!

Are you seeing both `doc/man/taskrc.5.in` and `doc/man/taskrc.5`? Only the first of those is a "source" file - the other is generated from the first when you run `cmake` and just has a few minor substitutions. Yes, this should be added to `taskrc` as well. I'll merge this PR now, so perhaps you can make another to add to `taskrc`?

PR #3467, opened 2024-05-27
Placing limit in man 5 taskrc

Andonome:
The new line is in the taskrc file.  I didn't think building would be that easy, so I might try some other fixups when time allows.

Unfortunately, github seems to have destroyed my commit - I suppose it's that 'Allow edits by maintainers' button?
I've unticked it, but then maybe that will cause problems? Should I

- rebase and leave the last commit which Github changed?
- Or simplify to a [new branch with one commit](https://github.com/GothenburgBitFactory/taskwarrior/commit/c0370a14af404bd11586bec8a96aaf30acff834f)?

### DIFF
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -242,6 +242,13 @@ Determines whether to use ioctl to establish the size of the window you are
 using, for text wrapping.
 
 .TP
+.B limit:25
+Specifies the desired number of tasks a report should show, if a positive
+integer is given.  The value 'page' may also be used, and will limit the
+report output to as many lines of text as will fit on screen. Default value
+is '25'.
+
+.TP
 .B defaultwidth=80
 The width of output used when auto-detection support is not available. Defaults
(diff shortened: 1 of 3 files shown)


djmitche:
> The new line is in the taskrc file. I didn't think building would be that easy, so I might try some other fixups when time allows.

Great!

> Unfortunately, github seems to have destroyed my commit - I suppose it's that 'Allow edits by maintainers' button? I've unticked it, but then maybe that will cause problems? Should I

I don't think that's related. That option would let me push changes to your branch (such as if I wanted to fix a typo and merge without bothering you for another round of review). I didn't do that in this case!

>     * Or simplify to a [new branch with one commit](https://github.com/GothenburgBitFactory/taskwarrior/commit/c0370a14af404bd11586bec8a96aaf30acff834f)?

This is typically the correct strategy -- one branch per PR. This is typically called a "topic branch".

That said, the change to the manpage here looks good, and GitHub seems willing to merge it, so I'll click the button and I think everything will be fine.

PR #3468, opened 2024-05-28
remove custom commands and custom scripts

### LINKED ISSUE #3462: Custom commands blocking usage of different build file generators
The currently highlighted lines do block the usage of e.g. Ninja as a build file generator, due to the different handling of the `$` character.

https://github.com/GothenburgBitFactory/taskwarrior/blob/8a807af2ef1cc99529beda4697fac2b39c727cec/CMakeLists.txt#L161-L176

As far as I understand these commands, they can be used to check out a specific PR and then play around in this repository. 
I think this functionality is now similarly given when we work over the .devcontainer (#3144), which is also a unique interface where everyone should be able to work in the same docker file. 

I would propose to remove the highlighted lines, as well as the corresponding docker files. As I assume one can achieve the same with a bit more standard approach (Start Devcontainer, Checkout PR, Run Tests).

felixschurk:
Fixes #3462.


djmitche (approved):
Thanks!

PR #3471, opened 2024-05-31
Add note about gcp application-default credentials to (5)task-sync

QuantumBJump:
The `task-sync` man page only mentions the `sync.gcp.credential_path` as part of setting up custom IAM roles, but it is also required if using application-default credentials. Not setting the credential path causes `task sync` to panic, as taskchampion tries to create a GcpService but unwraps `credential_path` in the process. This doesn't fix the taskchampion problem (if possible I plan to make a PR for that too), but it does make it easier for others to deal with this error if they encounter it.

### DIFF
--- a/doc/man/task-sync.5.in
+++ b/doc/man/task-sync.5.in
@@ -86,9 +86,12 @@ Authenticate to the project with:
     $ gcloud config set project $PROJECT_NAME
     $ gcloud auth application-default login
 
+This will download a file named "application_default_credentials.json" (the
+location will be shown in the output of the gcloud login command)
 Then configure Taskwarrior with:
 
     $ task config sync.gcp.bucket               <bucket-name>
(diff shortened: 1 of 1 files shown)


djmitche:
I would prefer to fix the issue, rather than document a workaround. AIUI application-default credentials typically do not require a path. What kind of time-frame do you think is reasonable for that fix?

QuantumBJump:
It's my first time contributing to open source so I might need some more time to familiarise myself with the taskchampion codebase and I'm currently in a field, so it might take up to a little over a week, but yes, I'm absolutely happy to prefer a proper fix!

I'll get it done asap :)

djmitche:
Sounds good - let me know how we can help!

QuantumBJump:
GothenburgBitFactory/taskchampion#394 fixes this so that the application-default credentials are used if not set otherwise in the config. Therefore this docs update is unneeded.

PR #3472, opened 2024-06-02
Add a release-check action to verify release tarballs have all the necessary parts

### LINKED ISSUE #3428: Check packaging in CI
As a quick check:
 - make a release tarball
 - untar it and build it

This would catch one of the bugs that led to 3.0.1.

djmitche:
This action creates a release tarball, un-tars it, and builds from it. This should verify that no critical piece is missing. Fixes #3428.

### DIFF
--- a/.github/workflows/release-check.yaml
+++ b/.github/workflows/release-check.yaml
@@ -24,5 +24,5 @@ jobs:
           tar -xf build/task-*.tar.gz &&
           cd task-*.*.* &&
           cmake -S. -Bbuild &&
-          make -Cbuild task_executable
+          cmake -Bbuild --target task_executable
 
--- a/.github/workflows/release-check.yaml
+++ b/.github/workflows/release-check.yaml
@@ -1,4 +1,4 @@
-name: tests
+name: release-tests
 on: [push, pull_request]
 jobs:
   check-tarball:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-check.yaml
```diff
@@ -0,0 +1,28 @@
...
+          override: true
+
+      - name: make a release tarball and build from it
+        run: |
+          cmake -S. -Bbuild &&
+          make -Cbuild package_source &&
+          tar -xf build/task-*.tar.gz &&
+          cd task-*.*.* &&
+          cmake -S. -Bbuild &&
+          make -Cbuild task_executable
```
ryneeverett: Why are we using make rather than cmake here?

ryneeverett (approved):

.github/workflows/release-check.yaml
djmitche: Huh,
```
cmake -S. -Bbuild --target task_executable
```
doesn't work, but omitting `-S.` fixes it:
```
cmake -Bbuild --target task_executable
```
so I guess we can use that!

djmitche: Oh, `-B` and `--build` aren't the same!

ryneeverett (approved):
I like seeing alignment of CI and docs, thanks!

PR #3473, opened 2024-06-02
Only warn about .data files when showing reports

### LINKED ISSUE #3419: Warnings leak into bash auto-completion
`task --version` is 3.0.2

I type `task new` with intention to run `task news` command ~and read the news about my tasks have been annihilated~. After typing `w` I get tired of typing, so, as hackers usually do in these situations, press "tab" to trigger bash auto-completion. I expect last argument to be completed to "news", but instead I see a multi-line warning duplicated several times

task export shall not restore your data, give up hope

I believe **auto-completion** is not supposed to trigger these or any kind of warnings at all, because that leads to poor UI. It may be worth to review and rework the "warnings" mechanism to exclude them from auto-completion output.

**UPD** A more realistic example: I was doing instructions from https://taskwarrior.org/docs/upgrade-3/, i. e. downgraded taskwarrior, exported my tasks into ~/tasks.json, upgraded back to the future, typed `task import rc.hooks=0 tas`... and pressed `<Tab>` to autocomplete the file name... which filled my screen with warnings

djmitche:
This avoids the warning appearing in shell completion, for example.

Fixes #3419.

### DIFF
--- a/src/commands/CmdCustom.cpp
+++ b/src/commands/CmdCustom.cpp
@@ -281,7 +281,6 @@ int CmdCustom::execute (std::string& output)
       format ("Found existing '*.data' files in {1}", location)) << "\n";
     std::cerr << "  Taskwarrior's storage format changed in 3.0, requiring a manual migration.\n";
     std::cerr << "  See https://github.com/GothenburgBitFactory/taskwarrior/releases.\n";
-    std::cerr << "  Remove the '*.data' files to silence this warning.\n";
   }
 
   feedback_backlog ();
--- a/src/commands/CmdCustom.cpp
+++ b/src/commands/CmdCustom.cpp
@@ -30,6 +30,7 @@
 #include <sstream>
 #include <map>
 #include <vector>
+#include <iostream>
 #include <algorithm>
 #include <stdlib.h>
 #include <Context.h>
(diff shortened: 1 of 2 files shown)


src/commands/CmdCustom.cpp
```diff
@@ -272,6 +273,16 @@ int CmdCustom::execute (std::string& output)
...
 
+  std::string location  = (Context::getContext ().data_dir);
+  File pending_data = File (location + "/pending.data");
+  if (pending_data.exists()) {
+    Color warning = Color (Context::getContext ().config.get ("color.warning"));
+    std::cerr << warning.colorize (
+      format ("Found existing '*.data' files in {1}", location)) << "\n";
+    std::cerr << "  Taskwarrior's storage format changed in 3.0, requiring a manual migration.\n";
+    std::cerr << "  See https://github.com/GothenburgBitFactory/taskwarrior/releases.\n";
+    std::cerr << "  Remove the '*.data' files to silence this warning.\n";
```
ryneeverett:
```suggestion
```
I still don't like this line for the same reasons as before: https://github.com/GothenburgBitFactory/taskwarrior/pull/3321#discussion_r1545527051

ryneeverett (commented):
I don't understand the code base well enough to give a technical review, but it strikes me as wrong if bash completion is calling `open_replica`.

src/commands/CmdCustom.cpp
djmitche: Oh, I thought I removed that sentence at your request. Oops!

djmitche:
> I don't understand the code base well enough to give a technical review, but it strikes me as wrong if bash completion is calling `open_replica`.

Yeah, this was also opening the TDB2 files. I'd expect a refactor here could help. But for the moment this fixes the user-visible bug.

felixschurk (approved):

PR #3479, opened 2024-06-12
Be resilient to a missing 'entry' in urgency

### LINKED ISSUE #3478: Task missing `entry` causes assertion failure calculating urgency
#### To report a bug...

My configuration no longer works with the line:
`report.next.filter=+READY or (+PENDING and scheduled.before:eod and not +BLOCKED)`

I believe this is synonymous with:
`report.next.filter=+PENDING and (-BLOCKED or scheduled.before:eod)`

Running `task "+PENDING and (-BLOCKED or scheduled.before:eod)` returns same message.

* I expected the filter to function.

* Observed behaviour was an error message:
`task: /usr/src/debug/task/task/src/Task.cpp:2322: float Task::urgency_age() const: Assertion `has ("entry")' failed.
zsh: IOT instruction (core dumped)  task`

* Running `task <filter>` with `and` carried no error. This bug is only with the `or` keyword.

* Output of ```task diag```:
```
task 3.0.2
   Platform: Linux

Compiler
    Version: 13.2.1 20230801
 Compliance: C++17

Build Features
     Commit: bc86a1e53
      CMake: 3.29.2
 Build type:

Configuration
       File: /home/username/.config/task/taskrc (found), 5928 bytes, mode 100644
       Data: /home/username/.local/share/task (found), dir, mode 40755
     TASKRC: /home/username/.config/task/taskrc
   TASKDATA: /home/username/.local/share/task
    Locking: Enabled
         GC: Enabled
    $EDITOR: nvim
Hooks
     System: Enabled
   Location: /home/username/.local/share/task/hooks
     Active: on-add.taskwarrior-autotagger.py    (executable) (symlink)
             on-modify.taskwarrior-autotagger.py (executable) (symlink)
             on-modify.timewarrior               (executable)
   Inactive: on-add.trackwarrior                 (not executable) (symlink)
             on-modify.timetracking              (not executable)
             on-modify.trackwarrior              (not executable) (symlink)

Tests
   Terminal: 264x73
       Dups: Scanned 1455 tasks for duplicate UUIDs:
             No duplicates found
 Broken ref: Scanned 1455 tasks for broken references:
             No broken references found
```

djmitche:
Any combination of properties is possible - Taskwarrior should make the best of the tasks it finds, rather than crashing.

Fixes #3478.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -2319,7 +2319,8 @@ float Task::urgency_due () const
 ////////////////////////////////////////////////////////////////////////////////
 float Task::urgency_age () const
 {
-  assert (has ("entry"));
+  if (!has ("entry"))
+    return 1.0;
 
   Datetime now;
   Datetime entry (get_date ("entry"));


felixschurk:
Can we reproduce the bug with a test, to make sure it will not slip in at a later point in time?

But I also do not know how we can get a task into the state of missing the entry date.

djmitche:
Indeed, `task import` seems to add the `entry` if it's not included. I don't know how else this could be done in a test. Even a C++ test wouldn't be able to do that, without interfacing directly with TC, which seems very low-level for something like this.

I think we will need to go back and perform a more thorough job of testing TW for its handling of cases with "missing" properties, so maybe we can find a better way to test at that point (perhaps by accessing taskchampion directly).

felixschurk:
Yes probably. But thats another issue. :D

felixschurk (approved):

PR #3482, opened 2024-06-14
Default values for blocking tasks is to inherit urgency from blocked tasks

### LINKED ISSUE #3111: Dependencies can have lower computed urgency than the tasks they block.
Taskwarrior supports marking tasks as dependent on other tasks, as described on this ['Best Practices' page](https://taskwarrior.org/docs/best-practices/).

For example:
```task ID modify depends:OTHER_ID```

The blocking nature of the relationship between the two tasks this is meant to represent means that the dependency should always have a higher urgency than the dependent task.
Taskwarrior does add a weight to the urgency of a task, as shown by the alterations to urgency when the `depends` relationship is added in the following example:

```
$ task pro:test add "dependent"
Created task 75.

$ task pro:test add "dependency"
Created task 76.

$ task pro:test

ID Age   Project Description Urg 
75 10s   test    dependent      1
76  7s   test    dependency     1

$ task 75 mod depends:76
Modifying task 75 'dependent'.
Modified 1 task.

$ task pro:test

ID Age   Deps Project Description Urg 
76 22s        test    dependency     9
75 25s   76   test    dependent     -4
```

So as expected, marking one task as dependent on another will modify urgencies such that the dependency has higher urgency than the dependent.

However, if other factors boost the urgency of the dependent, this change is not reflected in the dependency's urgency. For example, if we raise the priority of the dependent and give it a deadline:

```
$ task mod 75 priority:H
Modifying task 75 'dependent'.
Modified 1 task.

$ task 75 mod due:eod
Modifying task 75 'dependent'.
Modified 1 task.

$ task pro:test

ID Age   Deps P Project Due   Description Urg 
75 15min 76   H test    12h   dependent   10.6
76 15min        test          dependency     9
```

The urgency of the dependent was boosted, but the dependency's urgency was unchanged.
**The relative urgencies no longer reflect the blocking relationship between the two tasks.**
A user who has carefully input the dependency and trusts the system to track these relative urgencies is misled by what is reported and must do additional mental work to identify and account for the system's error.

Does Taskwarrior support alternatives to the default urgency calculation?
It seems that the default urgency calculation is a weighted-sum based on various attributes, as described [here](https://taskwarrior.org/docs/urgency/).

Modifying, e.g., `urgency.blocked.coefficient` and `urgency.blocking.coefficient` could decrease the chances that a blocking relationship would be misrepresented, but a mathematical model of urgency more complex than weighted-sum would be necessary to guarantee that the blocking relationship are reflected by urgency without overwhelming the other factors.

Mathematically, there are ways of combining the `depends` relationship data with the existing urgency coefficients -- excluding the 'blocked' & 'blocking' coefficients -- to compute urgencies that have a similar character to the existing urgency calculation while always respecting blocking relationships.
Perhaps something from linear algebra, for example, can elegantly keep the general character of the current model, while incorporating the guarantee that dependencies will have higher urgency than their dependents. Or, a more discrete procedure can certainly do it, replacing the 'blocked' and 'blocking' coefficients while guaranteeing that urgency will properly order dependencies.

If Taskwarrior does have a means of supporting alternative urgency calculation methods, then maybe someday I'll get to this. I'm reminded of the need for it every time I use the `depends` feature.

00sapo:
Fixes #3111


djmitche (approved):
I agree this is a better default.

Tests are failing for #3485, but I'll hold off merging until that is fixed.

djmitche (approved):

djmitche:
Rebased onto the latest..

PR #3483, opened 2024-06-14
create coverage actions

### LINKED ISSUE #3413: Coverage data in GH Actions
When looking, what accidentally happened in #3271, e.g. kicking out the python tests, I guess it would make sense to have some coverage data available in the repository, automatically build by GH Actions on every PR. To provide a broad overview and an easy to see metric for incoming PRs.

As I can see, this was available in #1984, but probably got lost when transitioning to GitHub.
Maybe it can also be used when new functionality is added to have some feedback if it is tested or not. 

I guess a starting point would be to use [Code Coverage Summary](https://github.com/marketplace/actions/code-coverage-summary) actions and instrument the code accordingly and collecting the data over [gcovr](https://github.com/gcovr/gcovr).
In order not to blow the complete GH Actions maybe it would make sense to run it only after the test were successful and maybe also only in one image (Debian, Ubuntu or Fedora) as the coverage should not change depending on the OS.

felixschurk:
Closes #3413.

With this PR a coverage comment is automatically created and updated over the corresponding actions file and attached to the PR. 

It is also possible to link the generated coverage report to some online viewer such as [Coveralls](https://coveralls.io/), [SonarCloud](https://sonarcloud.io/) to get even more insight e.g. evolution over coverage, deeper insights which lines are covered etc. 
I do think this makes sense in general, but only if we would then also look at the metrics. Otherwise, it is more dead code/dead connection and a simple comment is probably more helpful.


felixschurk:
It is also possible to link/trigger this workflow only after the successful run of e.g. the tests workflow. 

Generally speaken, we could link / depend workflows that they might have a more particular order e.g. `checks` -> `tests` -> `Coverage Report` -> `release-tests`. Mainly an idea to have it a bit more structured.
I have not yet looked to deep into the workflows directly if that would make sense.

felixschurk:
I have no clue, why all other tests are failing. As this PR has no changes in the source files nor build system.

djmitche:
I pushed #3484 to see if the tests fail there, too.

djmitche:
They did, so those failures are unrelated. I'll see if I can sort them out today.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,46 @@
+name: Coverage Report
+
+on:
+  push:
+    branches: [ "develop" ]
```
ryneeverett: I'm not sure we need to run after merging into develop because there's no chance of us seeing the report. (Unless the `develop` report is used to compare to PR reports.)

ryneeverett (commented):
- I like your suggestion of making this workflow run only after all others have passed.
- It seems like this report is only going to show the current coverage but I think what we're really interested in seeing in the PR is the *change* in coverage. This is where I suspect using a third-party tool like Coveralls comes in handy because such a tool would store the historical data for this purpose. IMO, ideal would be to only make a comment if there is a significant decrease in coverage, but even just showing the change would be an improvement.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Install apt packages
+        run: sudo apt-get install -y build-essential cmake git uuid-dev faketime locales python3 curl gcovr ninja-build
+
+      - name: Check out this repository
+        uses: actions/checkout@v4.1.6
+
+      - name: Build and test this project
+        run: |
+          cmake -S . -B build -G Ninja K-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=--coverage
```
djmitche: What is the `K` here?

djmitche (commented):
I tend to agree, that this isn't especially helpful without a baseline. I think coveralls can do that, though -- and even post a comment only if some conditions apply? It would be nice to get alerts if PRs dramatically reduce the coverage rate or introduce a lot of new un-covered code!

felixschurk:
> * I like your suggestion of making this workflow run only after all others have passed.
> 
>     * It seems like this report is only going to show the current coverage but I think what we're really interested in seeing in the PR is the _change_ in coverage. This is where I suspect using a third-party tool like Coveralls comes in handy because such a tool would store the historical data for this purpose. IMO, ideal would be to only make a comment if there is a significant decrease in coverage, but even just showing the change would be an improvement.


Yes, thats actually what the Coveralls Action would do. 

> I tend to agree, that this isn't especially helpful without a baseline. I think coveralls can do that, though -- and even post a comment only if some conditions apply? It would be nice to get alerts if PRs dramatically reduce the coverage rate or introduce a lot of new un-covered code!

So I would say we set it up with coveralls, as their action does both of the mentioned suggestions automatically. 
The [GitHub Coveralls Action](https://github.com/marketplace/actions/coveralls-github-action), I can setup. 

I already requested from the GothenBurgBitFactory Organization Access for adding the Repository to Coveralls, I just don't know who will get the notification for that one.

felixschurk:
Okay, I checked it a bit out with the running one action after another and it did not directly work as expected. 

However, when we want to use the coveralls action (which would make sense) then it does need to run on pull requests to post the comment etc. thus linking it over the `workflow_run` I assume is not working. 

However, what we could still do is to use then this action (the coverage one) to test is only on Ubuntu 22.04, and if this one is working, then test it on all other distros. Meaning in a bad case we would save the 8x4 minutes. But thats up for discussion, as we also currently do not hit any CI limits to my knowledge its not really required.

djmitche:
I think that sounds fine. Most failures fail on every builder, so all but the first are a waste anyway.

tbabej:
> I already requested from the GothenBurgBitFactory Organization Access for adding the Repository to Coveralls, I just don't know who will get the notification for that one.

This should be approved now.

felixschurk:
Integration with Coveralls is now setup and can be seen [here](https://coveralls.io/github/GothenburgBitFactory/taskwarrior). I assume we would need to also set up the comment function by inviting the coveralls bot [see here the documentation](https://docs.coveralls.io/app-notifications).


On another note, I think it's quite interesting to see that over the last commits different tests fail, e.g. over a time-out from `task`. I tried to reproduce that locally somehow, however I am not able to do.
I am now wondering if that has to do as they now run "natively" under the GH Actions runner and not in a separate Container in the GH Actions runner.

djmitche:
An example failure is `task config` failing with SIGABRT, which doesn't seem like it could time out -- it doesn't do anything lengthy.

felixschurk:
Alright, seams if I found a way to properly chain them. 
I am a bit unsure why in previous commits the tests where "randomly" failing, but over the last two commits it was now stable - so I hope this will stay so.

With the current setup the `tests` first runs the coverage check, if this one is successful it will build on all the machines. 

Further, the small badge is added to the README.
[Coverage Status](https://coveralls.io/github/GothenburgBitFactory/taskwarrior?branch=add-coverage-actions)

djmitche:
Nice work - thanks!

felixschurk:
Okay this was not intended to merge it without the review!
I thought I can set it to automatically squash and merge after the reviews, however seams to be not the case. 

Sorry for that!

felixschurk:
Currently, the settings for Coveralls are set to:

80% Overall Coverage (currently 84%)
5% Coverage Decrease with New Code allowed

I have not yet set it up that Coveralls leaves a PR Comment about the change in coverage, however if wanted this could also be setup [see the documentation](https://docs.coveralls.io/app-notifications).

djmitche:
Hm, I'm seeing the SIGABRT in https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/9613757645/job/26517145647?pr=3494

In wonder if that's due to the `-j8` in the `ctest` invocation? I was about to make a PR to _add_ `-j8` for the build, so I can try that.

felixschurk:
You mean for the coverage invocation build?

There it should not be needed to add a `-j 8` flag, as it uses Ninja which should do it automatically (using all available cores). 

Now reading over the [documentation from the runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) I see they only have four cores - so maybe we should only add a `-j 4` flag. :D

Ah and also the [documentation of `ctest`](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-j) would allow us to run it in maximum parallel when we just pass `--parallel` to it (either limited by number of processors or two, whatever is larger).

djmitche:
I think I had already marked this as "Approved" - that's probably why it merged? Anyway, it's fine :)

djmitche:
I didn't realize Ninja was parallelizing already. But, removing the parallelism from ctest did help the tests not fail. I'm not sure the tests are _designed_ to run in parallel -- they might be using some shared resource?

felixschurk:
Hm, I assumed as even the old test script before #3446 allowed at least for a flag to run the test suite in parallel that this should not be a problem. 

What I think is confusing and does not really add up is that in all the distro tests, we do run the tests in parallel (`j8`) and there was never a spurious failure. So I am assuming that maybe here the GH Runner is playing some role and maybe there it's having some influence. But also no clue why a container, in a container should change there a lot.

djmitche:
Could it be different versions of ctest?

I hate autocorrect

djmitche:
One thought, is that it's possible the measurement of the number of processors is different in a container. I have forgotten most of what I ever knew about Docker. But perhaps the tasks weren't really running in parallel in the container?

djmitche:
Huh, well, not running tests in parallel didn't help -- that PR failed anyway.

PR #3486, opened 2024-06-14
Remove unnecessary cbindgen-related Makefile

djmitche:
I noticed this leftover while working on another PR. We don't use cbindgen at all.


djmitche:
Test failures are #3845.

PR #3493, opened 2024-06-19
Consider all news "major"

### LINKED ISSUE #3391: News does not remember last viewed version without undocumented option
Running taskwarrior 3.0 and 3.0.1 I've been getting random reminders to read the version news, despite having read both `task news` and `task news 3.0.1 minor`. I would expect that to satisfy the automatic news version tracking.

Checking with `task show | grep version` I noticed that the `news.version` variable had not been updated. I had to check the [`CmdNews`](https://github.com/GothenburgBitFactory/taskwarrior/blob/a9995808eca56e87841f153c82694c142a83fd6f/src/commands/CmdNews.cpp#L653) source to discover that the saved "read version" is only updated when `task news all` is run, but the "all" specifier doesn't seem to be documented anywhere?

Logically something seems off, it only reverts the signal handler and asks for donations if the "all" command is used?

Not sure exactly what the intended behavior is and if that ending if case needs to be rescoped but I think at minimum reading the news should update the config flag. As a minimal change, maybe the "all" setting that updates the config flag should be the default?

I am willing to contribute a PR for this if given guidance on the preferred behavior.

djmitche:
This has the effect that `task news` will unconditionally update the config with the new version once news has been shown (assuming the user does not kill the process first).

Fixes #3391.

### DIFF
--- a/src/commands/CmdNews.cpp
+++ b/src/commands/CmdNews.cpp
@@ -58,7 +58,7 @@ CmdNews::CmdNews ()
   _uses_context          = false;
   _accepts_filter        = false;
   _accepts_modifications = false;
-  _accepts_miscellaneous = true;
+  _accepts_miscellaneous = false;
   _category              = Command::Category::misc;
 }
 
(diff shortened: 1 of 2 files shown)


djmitche:
apologies for reviewer-shopping :)

felixschurk (approved):

PR #3500, opened 2024-06-20
Include version in default-generated .taskrc

### LINKED ISSUE #3498: Set version to the latest when creating a new `.taskrc`
The `task news` command determines how much news to show using the `news.version` config. However, a new `.taskrc` does not have this value set, and thus will remind the user to run `task news` and tell them about every new feature since 2.5.0.

Instead, let's set that to the current version.

djmitche:
This will avoid new users being prompted with all the news since 2.5.0.

I don't see other tests for the default .taskrc creation, so I didn't add a new one. I did test this by hand though.

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -40,6 +40,7 @@
 #include <Variant.h>
 #include <Datetime.h>
 #include <Duration.h>
+#include <Version.h>
 #include <shared.h>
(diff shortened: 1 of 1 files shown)


tbabej:
This is a really good idea!

tbabej (approved):

PR #3502, opened 2024-06-21
Build in parallel, test sequentially

djmitche:
It seems the tests fail with SIGABRT when run with `-j 8` - perhaps there's just too much going on for a 4-thread processor to keep up? Or perhaps the tests interfere with each other? So, this removes the parallelism from tests.

However, _building_ in parallel should be fine, so let's do that.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,7 +17,7 @@ jobs:
         run: cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=--coverage
 
       - name: Build project
-        run: cmake --build build -j 8 --target build_tests
+        run: cmake --build build --target build_tests
 
       - name: Test project
         # NOTE: running these in parallel (`-j N`) can result in spurious failures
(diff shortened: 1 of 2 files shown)


felixschurk:
https://github.com/GothenburgBitFactory/taskwarrior/pull/3483#issuecomment-2182696577

Maybe to have it easier to talk about.

felixschurk:
Maybe checking it out with having the systems building automatically with maximum parallelism (build no argument, as Ninja is parallel)
> Builds are always run in parallel, based by default on the number of CPUs your system has. Underspecified build dependencies will result in incorrect builds - From Ninja - Manual

And using for CMake also the `--parallel` flag, as of CMake/CTest 2.29 which should also just use the possible maximum number? https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-j

.github/workflows/tests.yaml
```diff
@@ -17,10 +17,11 @@ jobs:
         run: cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=--coverage
 
       - name: Build project
-        run: cmake --build build --target build_tests
+        run: cmake --build build -j 8 --target build_tests
```
felixschurk: Here `-j 8` is not needed as this build is with Ninja.

felixschurk (changes requested):

djmitche:
Well, the coverage failed with SIGABRT here, so I suppose this is not the right fix :(

PR #3506, opened 2024-06-22
remove test certificates and link in python code

felixschurk:
With taskwarrior > 3.0 these are no longer required and thus can be
removed.


ryneeverett (approved):
I don't know what the test certs were for but they're clearly not needed now since the tests pass without them.

djmitche (approved):
It looks like they have been around for a [looooong time](https://github.com/GothenburgBitFactory/taskchampion/commit/29d5de1154711827d76d948aab7314499d6c30cf).

PR #3509, opened 2024-06-23
Properly tag monospaced text in man pages. (Improves HTML rendering)

sebastiancarlos:
This would be immediately useful to improve the rendering of the man pages in third-party websites such as manned.org and the arch man pages:
- https://mankier.com/1/task
- https://man.archlinux.org/man/task.1

The regular console output, or websites that render the manpage in monospaced fonts, will not be affected by this change.

This change could also help with eventually rendering the manpages in the documentation (https://github.com/GothenburgBitFactory/tw.org/issues/794) website and mixing monospaced and variable-width fonts.

To test the HTML output:
```bash
git clone git@github.com:jacksonp/manner.git # used by ManKier
./manner/manner.php task.1.in >| task.1.html && $BROWSER task.1.html
```

### Before
Screenshot 2024-06-23 at 12 19 30 PM
### After 
Screenshot 2024-06-23 at 12 19 48 PM

### DIFF
--- a/doc/man/task-color.5.in
+++ b/doc/man/task-color.5.in
@@ -8,14 +8,18 @@ It should be mentioned that Taskwarrior is aware of whether its output is going
 to a terminal, or to a file or through a pipe.  When Taskwarrior output goes to
 a terminal, color is desirable, but consider the following command:
 
+.nf
     $ task list > file.txt
+.fi
 
 Do we really want all those color control codes in the file?  Taskwarrior
 assumes that you do not, and temporarily sets color to 'off' while generating
 the output.  This explains the output from the following command:
 
+.nf
     $ task show | grep '^color '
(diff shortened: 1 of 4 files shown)


djmitche (approved):
Fantastic, thanks! Does `task-sync.5.in` need similar attention?

djmitche:
Oh, I just saw

> Note: I ignored task-sync.5 as it is deprecated.

That's not the case -- sync is not deprecated, just the old `taskserver` aka `taskd`.

sebastiancarlos:
> That's not the case -- sync is not deprecated, just the old `taskserver` aka `taskd`.

Ah, I missed that. I did the same to `task-sync.5.in` now.

djmitche (approved):
Awesome, thanks!

PR #3511, opened 2024-06-23
Document the 'status' attribute in the man page

sebastiancarlos:


djmitche (approved):
that seems important...

Issue #3507, opened 2024-06-23
SIGABRT in coverage runs


djmitche:
Hm, I'm seeing the SIGABRT in https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/9613757645/job/26517145647?pr=3494

In wonder if that's due to the `-j8` in the `ctest` invocation? I was about to make a PR to _add_ `-j8` for the build, so I can try that.

_Originally posted by @djmitche in https://github.com/GothenburgBitFactory/taskwarrior/issues/3483#issuecomment-2182689611_

djmitche:
@felixschurk we can continue the conversation from #3483 here (since that PR is closed and can't be re-opened).

felixschurk:
I do think it has to be with the GitHub Ubuntu image, but I do not know yet how to debug it. I think maybe pulling the GitHub image locally and running in it the tests leads to some insights?

I assume it does not has to do anything directly with the parallel flags from `ctest` as locally, as well as in all the distro tests, the tests are run in parallel and we never saw such an error. 

For the time being, as a workaround, one can just rerun the failing CI runs... and hope it will work the next time.

felixschurk:
When looking over the failed test runs [first], [second] and [third] all of them fail on some `task config ...` invocation.

[first]: https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/9629105362/job/26558080511
[second]: https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/9629096449/job/26558063202
[third]: https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/9620433739/job/26538608969

djmitche:
So another hypothesis I had was that the timeout logic was just waiting for a signal, and failing to retry on things like EINTR. But `wait_process` and `wait_condition` actually poll for `kill(pid, 0)` to return an error. It _looks_ pretty solid, but maybe there's some other weird dependency on kernel behavior there?

djmitche:
Hm, I added some `print` calls in `wait_process` and `wait_condition` -- and now it won't fail.

djmitche:
Ah, I spoke too soon - it fails!

```
ok 3 - bash_completion.test.py: no gc coming from bash completion
Mon Jun 24 01:28:04 2024 wait_process(6246, None)
Mon Jun 24 01:28:04 2024 wait_condition tries=100 timeout=1 sleeptime=0.01
[...]

PR #3512, opened 2024-06-23
Add logging for coverage failures

djmitche:
Maybe this will help figure out what's up with #3507.


felixschurk:
I think the 10s timeout is the one that did the job. 
Currently the CI is "failing" as Coveralls has already the data from this build, and therefore does not konow what to do. But there are no longer the tests failing.

PR #3523, opened 2024-06-27
increase timout to 10s and run on ubuntu-latest

### LINKED ISSUE #3507: SIGABRT in coverage runs

felixschurk:
As identified in #3512 it seams as the problem was the internal timeout. Closes #3507.

I mainly created the "new" PR that we do not need to commit more times with the logging etc, and that the Coveralls action has a normal chance to run successfully.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 
 jobs:
   coverage:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Install apt packages
         run: sudo apt-get install -y build-essential cmake git uuid-dev faketime locales python3 curl gcovr ninja-build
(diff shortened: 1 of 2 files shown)


djmitche (approved):
Looks good!

1s seems like it would be long enough, but I suppose if another tenant on the same machine starts swapping or otherwise slows the system down, it could take >1s to complete the invocation. 10s seems a lot safer, and a few re-runs suggest that's the case.

djmitche:
I re-ran that a few times and it's all green, so I say let's do it.

PR #3525, opened 2024-06-28
Consistently exclude `WAITING` tasks from reports which filter by status

sebastiancarlos:
If I understand correctly, the virtual status "waiting" is deprecated. This is why some reports include the explicit "-WAITING" filter even though it is not necessary: The idea is that being explicit will be needed in the future when the "waiting" status is removed.

But I noticed that some reports do not include the "-WAITING" filter, so I added it to all reports that filter by status.


ryneeverett:
You're saying that "WAITING" the [virtual tag](https://taskwarrior.org/docs/terminology/#virtual-tag) is deprecated, but not "waiting" the status/state, right? If I understand the terminology correctly, "virtual status" seems like an ambiguous combination of the two terms.

estebanlas-orangeloops:
I actually meant "virtual status", from `Task.cpp`:

```cpp
Task::status Task::textToStatus (const std::string& input)
{
       if (input[0] == 'p') return Task::pending;
  else if (input[0] == 'c') return Task::completed;
  else if (input[0] == 'd') return Task::deleted;
  else if (input[0] == 'r') return Task::recurring;
  // for compatibility, parse `w` as pending; Task::getStatus will
  // apply the virtual waiting status if appropriate
  else if (input[0] == 'w') return Task::pending;

  throw format ("The status '{1}' is not valid.", input);
}

...

Task::status Task::getStatus () const
{
  if (! has ("status"))
    return Task::pending;

  auto status = textToStatus (get ("status"));

  // Implement the "virtual" Task::waiting status, which is not stored on-disk
  // but is defined as a pending task with a `wait` attribute in the future.
  // This is workaround for 2.6.0, remove in 3.0.0.
  if (status == Task::pending && is_waiting ()) {
      return Task::waiting;
  }

  return status;
}
```

sebastiancarlos:
I actually meant "virtual status". As called in `Task.cpp`:

```cpp
Task::status Task::textToStatus (const std::string& input)
{
       if (input[0] == 'p') return Task::pending;
  else if (input[0] == 'c') return Task::completed;
  else if (input[0] == 'd') return Task::deleted;
  else if (input[0] == 'r') return Task::recurring;
  // for compatibility, parse `w` as pending; Task::getStatus will
  // apply the virtual waiting status if appropriate
  else if (input[0] == 'w') return Task::pending;

  throw format ("The status '{1}' is not valid.", input);
}

...

Task::status Task::getStatus () const
{
  if (! has ("status"))
    return Task::pending;

  auto status = textToStatus (get ("status"));

  // Implement the "virtual" Task::waiting status, which is not stored on-disk
  // but is defined as a pending task with a `wait` attribute in the future.
  // This is workaround for 2.6.0, remove in 3.0.0.
  if (status == Task::pending && is_waiting ()) {
      return Task::waiting;
  }

  return status;
}
```

Also [this PR](https://github.com/GothenburgBitFactory/taskwarrior/pull/2563)

djmitche (approved):
Good thinking! This was definitely an oversight when I removed that status.

PR #3526, opened 2024-06-28
Clarify relation between the DUE tag and the rc.due setting in task(1)

sebastiancarlos:
Also, clarify that `DUE` is based on a number of days since the current date.


djmitche (approved):
Thanks!

PR #3531, opened 2024-07-03
Update to taskchampion 0.6

felixschurk:


src/tc/lib/src/server.rs
```diff
@@ -125,7 +125,7 @@ pub unsafe extern "C" fn tc_server_new_sync(
             // SAFETY:
             //  - origin is valid (promised by caller)
             //  - origin ownership is transferred to this function
-            let origin = unsafe { TCString::val_from_arg(origin) }.into_string()?;
+            let url = unsafe { TCString::val_from_arg(origin) }.into_string()?;
```
djmitche, line 128: This is the minimum to make it compile :)

I think @chromy will do the rest in #3423.

djmitche (approved):

PR #3532, opened 2024-07-03
Update performance scripts

felixschurk:
This allows to run the performance scripts now again with the out-of-source build. 
Also update the script to use the new `*.sqlite` database and no longer the `*.data` files.

This way we can then also use the `performance` target to see if we did improve the performance with some modifications.


djmitche (approved):
Nice -- thanks for taking care of that!

PR #3535, opened 2024-07-05
Fix formatting in task-sync manpage

djmitche:
This fixes unnecessary indentation in this page, and also puts the GCP permissions in alphabetical order.


felixschurk:
Would it make sense to add a note that it's a "paid" service? 

Or did I just do something wrong when I was setting it up :D

djmitche:
This is just a formatting PR, but that might make sense in a further PR.

You didn't do anything wrong - Google does charge for it (although not much).

felixschurk:
Yes, it's really like nothing.

Perfect, then let's ship the formatting.

PR #3540, opened 2024-07-06
Restore 'task purge' functionality

### LINKED ISSUE #3399: Restore `task purge` support
`task status:deleted purge`
In v2.x, it works. In v3, it prints out
```As of version 3.0, this command has no effect.
Deleted tasks are removed from the task list automatically after they expire
```
* Why I need this?

  - I am writing a neovim plugin for interacting with Taskwarrior so I often create dummy tasks to test. 
  - Often I make a mistake, I will delete the task then reuse the previous command with correct info
  - It would be great to not see the deleted ones when running `all` report :)

djmitche:
This is partially a revert of 87467f72c71972f918b5892461f401edd1d51c60. The purge command is modified to only actually call `Replica::delete_task` after all possible reasons _not_ to purge anything have been handled.

Some minor changes were also required to the test -- in particular, the id of the child task that causes the error is nondeterministic.

Fixes #3399.


src/tc/lib/src/replica.rs
```diff
@@ -501,6 +501,35 @@ pub unsafe extern "C" fn tc_replica_import_task_with_uuid(
     )
 }
 
+#[ffizz_header::item]
+#[ffizz(order = 902)]
+/// Delete a task.  The task must exist.  Note that this is different from setting status to
+/// Deleted; this is the final purge of the task.
+///
+/// Deletion may interact poorly with modifications to the same task on other replicas. For
+/// example, if a task is deleted on replica 1 and its description modified on replica 1, then
```
ryneeverett:
```suggestion
/// example, if a task is deleted on replica 1 and its description modified on replica 2, then
```

I think this is what you meant.

src/commands/CmdPurge.cpp
```diff
@@ -45,12 +49,135 @@ CmdPurge::CmdPurge ()
...
+////////////////////////////////////////////////////////////////////////////////
+// Purges the task, while taking care of:
+// - dependencies on this task
+// - child tasks
+void CmdPurge::purgeTask (Task& task, std::vector<Task>& tasks)
+{
+  handleDeps (task);
+  handleChildren (task, tasks);
+  tasks.push_back(task);
+}
```
ryneeverett, line 61: `purgeTask` should probably be renamed to reflect this functions purpose now that it does not actually complete the purge. Maybe `purgeReferences`?

However, I'm not sure I understand the point of this change. In the PR description you write:

> The purge command is modified to only actually call `Replica::delete_task` after all possible reasons _not_ to purge anything have been handled.

Comparing this PR to 87467f72c71972f918b5892461f401edd1d51c60 ...

```cpp
void CmdPurge::purgeTask (Task& task, int& count)
{
  Context::getContext ().tdb2.purge (task);
  handleDeps (task);
  handleChildren (task, count);
  count++;
}
```

It seems like the change you're referring to is:

1. Before, `purgeTask` executed `tdb2.purge` prior to clearing references. Since errors could be encountered while clearing references, this opened the door to dangling references to purged tasks. Now, references are cleared before executing `tdb2.purge`.
2. Before, `tdb2.purge` was called on each task being purged in the loop, so an error in clearing references on a later task could cause some tasks to be purged but not others. Now, `tdb2.purge` is not called on any tasks until references for all tasks have been cleared.

To me, change (1) seems like a clear improvement (which could have been achieved just as easily by moving `tdb2.purge` to the end of `purgeTask`) but I'm not sure I understand the purpose behind (2). It seems like the new behavior would be that an error in clearing references could cause references to be cleared from tasks prior in the loop but now those tasks don't get purged from the database, so it isn't such a clear improvement.

ryneeverett (approved):

src/commands/CmdPurge.cpp
djmitche: The `tdb2.purge` does not take effect immediately, but waits until `commit`. That was a surprise to me! But it means that, in the 2.6.2 implementation, if any errors occurred, or if the use answered "no" to one of the prompts, no change would take place. That's the behavior I've tried to reproduce by only calling `delete_task` at the end.

`purgeTask` is sort of doing a recursive descent to establish a list of tasks to delete. I'm open to a better name, but I don't think `purgeReferences` is it.

ryneeverett (approved):

src/commands/CmdPurge.cpp
ryneeverett: > The `tdb2.purge` does not take effect immediately, but waits until `commit`. That was a surprise to me! But it means that, in the 2.6.2 implementation, if any errors occurred, or if the use answered "no" to one of the prompts, no change would take place. That's the behavior I've tried to reproduce by only calling `delete_task` at the end.

Thanks, that clears it up!

> `purgeTask` is sort of doing a recursive descent to establish a list of tasks to delete.

`handleDeps` also removes the task as a blocker from other tasks. Maybe `handleRelations` to signify that its purpose is to remove relationships with multiple types of linked tasks?

ryneeverett: I think a vague name would be kinder to future code readers than a clear name which misrepresents what the function does.

PR #3544, opened 2024-07-07
Update CmdDiagnostics.cpp

koleesch:
Remove duplicate check, because sqlite doesn't allow duplicate uuids


djmitche:
Dup of #3545

PR #3545, opened 2024-07-07
Update CmdDiagnostics.cpp

koleesch:
removed duplicate check, because sqlite doesn't allow uuids


djmitche:
Looking good!  It looks like there are a few tests to fix up..

koleesch:
Now it seems to be working with the tests. I've checked out the project, but on my computer not all tests are covered. So it is an iterative process. Sorry for that.

djmitche (approved):
Nice, thanks! I'm curious what tests don't run locally?

PR #3546, opened 2024-07-07
Add support for task expiration

### LINKED ISSUE #3402: Optional support for automatically expiring old tasks
As of version 3.0, Taskwarrior apparently removes deleted tasks "automatically after they expire". This feature is not documented anywhere, and the user is only made aware of it if they try to run the `purge` command (which has been deprecated and no longer works).

This new behavior appears to be silent, automatic, undocumented, and irreversible.

The meaning of the "deleted" status in Taskwarrior 2 was confusing, because deleted tasks would remain in the user's task list and data files unless the user explicitly used the `purge` command to remove the data permanently. However, while confusing, this behavior was the way Taskwarrior worked for years (decades?). As a result, due to deleted tasks being preserved (unless purged) and despite the weird semantics, I (and I suspect many other users) began "deleting" tasks as a way of marking them as cancelled (a semantic Taskwarrior has no command for).

Now, with Taskwarrior 3, all records of those tasks are apparently at risk of being automatically purged? That is not desirable behavior.

Taskwarrior should not automatically or silently purge user data unless explicitly configured to do so by the user.

After updating to Taskwarrior 3, I checked the information for my old deleted tasks and tasks that I created and deleted after the update. Nowhere in the information for those tasks did I find an `Expiry` field. The tasks have their usual `Entered`, `End`, `Last modified` fields, etc, but nothing telling me when the task will expire and be automatically purged.

1. When and under what conditions a deleted task will be considered "expired" and automatically purged needs to be presented to the user in that task's information.

2. The new behavior of automatic task purging needs to be documented properly and (at least on first run) explicitly enabled by the user.

3. Taskwarrior 3 should implement a "cancelled" status and respective commands, so that users can track not only the tasks they have completed, but the tasks they are not going to complete.

djmitche:
This adds optional support for task expiration, implemented by calling the relevant Replica method.

This is currently fixed at 180 days, but we can add configuration later.

I chose to do this on sync for two reasons:
 * Any modifications to tasks from other replicas are likely to have just been received, updating the `modified` timestamp and saving a task from being updated-and-deleted. Such a task will end up in a zombie state with only the updated properties surviving.
 * This is a fairly expensive operation (scan of the full DB) so not something we should do interactively, like renumbering the working set.

Fixes #3402.


src/Context.cpp
```diff
@@ -853,13 +854,8 @@ int Context::dispatch (std::string &out)
     // The command know whether they need a GC.
     if (c->needs_gc ())
     {
-      run_gc = config.getBoolean ("gc");
```
djmitche, line 856: This value is written but never read, so I removed it.

An early draft of this PR linked expiration to gc, so I was in this file making modifications. This simplification seemed worth keeping.

felixschurk (approved):
I looked over it and as far as I can understand everything works as expected and is documented.

djmitche:
> I looked over it and as far as I can understand everything works as expected and is documented.

I can't ask for anything more than that in a review!

felixschurk:
Interesting that here the tests all run fine, however when merged into develop (https://github.com/GothenburgBitFactory/taskwarrior/commit/213b9d3aee099a86e1efdeb32e73632e30f0844b) the tests fail in some of the runners. 

I don't understand why and how.

djmitche:
Huh, I re-ran them and they passed. If they continue to fail we can try reverting this, but it really seems unrelated.

felixschurk:
Yes, I don't think it has to do with the changes.
But still strange.

PR #3556, opened 2024-07-14
Rename 'expiration.on-sync' to 'auto-purge.on-sync'

### LINKED ISSUE #3402: Optional support for automatically expiring old tasks

djmitche:
Taskwarrior uses "expire" to refer to deletion of tasks past their "until" date, so let's use `auto-purge` to link this semantically to the `task purge` command.

@ryneeverett requesting your review since this is more a question of naming than C++..

Fixes #3402 (again).

### DIFF
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -217,8 +217,8 @@ rc.gc=0 ...), and not permanently used in the .taskrc file, as this
 significantly affects performance in the long term.
 
 .TP
-.B expiration.on-sync=0
-If set, old tasks will be deleted automatically after each synchronization.
+.B auto-purge.on-sync=0
+If set, old tasks will be purged automatically after each synchronization.
 Tasks are identified as "old" when they have status "Deleted" and have not
 been modified for 180 days.
 
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -217,7 +217,7 @@ rc.gc=0 ...), and not permanently used in the .taskrc file, as this
 significantly affects performance in the long term.
 
 .TP
-.B auto-purge.on-sync=0
+.B purge.on-sync=0
 If set, old tasks will be purged automatically after each synchronization.
 Tasks are identified as "old" when they have status "Deleted" and have not
 been modified for 180 days.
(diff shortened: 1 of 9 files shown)


ryneeverett (commented):
expiration->purge is an improvement but I'm not so sure about the "auto-purge". I think "purge.on-sync" is pretty clear. You could add the word "auto" to almost every option or feature of any computer program so I think we have to ask: Is there something *particularly* automatic about this feature?

djmitche:
My thinking was to distinguish this from somehow automatically running `task purge` on every sync. Which doesn't make a whole lot of sense, now that I think of it.

djmitche:
whoops, hold on :)

ryneeverett (approved):

PR #3560, opened 2024-07-19
Warn if an import contains multiple occurrences of the same UUID

### LINKED ISSUE #3548: Warn if an import contains multiple occurrences of the same UUID
If the data stream processed by the `import` command contains the same UUID twice, the effect will be to merge their contents. This might occur due to data corruption in 2.x (the old sync mechanism often duplicated tasks). The result of exporting a task DB containing duplicates and importing into 3.x might be data loss if those duplicated tasks are merged without warning. Warning the user at least gives them a chance to examine the results.

So, when the import input contains the same UUID more than once, provide a warning about that at the end of the input process. Warning when the duplicate is spotted wouldn't be very helpful, since typically `task import` produces a lot of lines of output.

This would be a change in [`CmdImport.cpp`](https://github.com/GothenburgBitFactory/taskwarrior/blob/61c9b486642a386cecf2cba64cce3a8292de96b6/src/commands/CmdImport.cpp).

Necior:


djmitche (approved):
Nice work!

PR #3561, opened 2024-07-20
Fix conversion from `TCStatus::Unknown`

Necior:
Before this patch, the messsage would be "unknown TCStatus 4294967295" (i.e. `u32::MAX`) instead of "unknown TCStatus -1".


djmitche (commented):
Nice fix, thank you!

djmitche (approved):

PR #3566, opened 2024-07-24
Add clang-format to enforce style guide

### LINKED ISSUE #1987: Enforce styleguide, possibly revisit code style and improve code health
[Taskwarrior Coding Style](https://taskwarrior.org/docs/coding_style.html) is honestly pretty vague in terms of describing what good code really looks like. Despite being slightly biased, I think the following actions make sense.

## Adopt an established coding style

As mentioned earlier, currently the coding style is pretty vague and probably somewhat strange. E.g. I find it strange to have both

> Indent code using two spaces, no tabs
> Stick to 80 columns where possible, although exceptions are fine

and

> Surround operators and expression terms with a space

First points make sense, but the latter is not only pretty uncommon (on how it's used: in general spases around operators and expressions are fine, but the project seems to have a convention of using a space before any opening brace e.g. before argument lists, etc: see https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/src/CLI2.h#L39), but takes up a lot of space, which is worth saving because of the 80 column-width limit.

Also, even this completely vague coding style is constantly violated (e.g. https://github.com/GothenburgBitFactory/taskwarrior/blob/2.6.0/src/CLI2.cpp#L592 and in many places in this file, which is the first one I simply took as an example).

My point is that the style generally used in Taskwarrior codebase is very similar to the [Google C++ Codestyle](https://google.github.io/styleguide/cppguide.html). Why not adopt it? It would also greatly decrease the cost of maintaining the codebase and enforcing this codestyle (due to the existence of a great number of tools for its enforcement).

## Use [Clang-Format](https://clang.llvm.org/docs/ClangFormat.html)

Upon fixing a code style it would make sense to start using Clang-Format, which is a very popular tool for enforcing proper formatting. It has a number of plugins for most common text editors, IDEs, tools and useful scripts (e.g. patch formatting [script](https://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting)). It supports a number of popular styles (and [configurable options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)) including Google and LLVM styleguides out of the box.

## Use [Clang-Tidy](http://clang.llvm.org/extra/clang-tidy/)

Clang-Tidy provides a convinient way of ensuring codebase health through a number of [checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html). Frequently running Clang-Tidy over the codebase makes sense to me, continuously generating reports is even better.

Personally, I like Neovim's approach - it provides analysis of the current codebase snapshot after each build in an [accessible form](https://neovim.io/doc/reports/clang/).

Along with https://github.com/GothenburgBitFactory/taskwarrior/issues/1984 it would probably greatly increase the codebase quality.

### LINKED ISSUE #3338: Start using clang-format to format C++ code

felixschurk:
In this PR `pre-commit` is added to automatically format code for python as well as cpp files.
General fixes such as whitespace etc. are enabled for all files.

Down the rabbit hole of changing cpp files, I noticed that `#include <cmake.h>` needs to be the first include header, as otherwise a compilation is breaking.
A comment, as well as a blank line are therefore added to prevent the automatic sorting of clang-format from breaking the build. Another work around would be to exclude that check, but I think having it explicitly stated is better than disabling something very normal.

Currently there should be some request for someone to add the `pre-commit.ci` to this repository, which should then pick up automatically the changed on every PR.

To not completely destroy the `git blame` feature, the commit containing the initial run of all the fixes is added to the `.git-blame-ignore-revs` file which is exlucded when viewing PRs.

Closes #3338.
Also addresses and closes #1987.


felixschurk:
I am not completely sure how we need to take care of the merge/rebase, as when the commit id of https://github.com/GothenburgBitFactory/taskwarrior/pull/3566/commits/7dfe39322167e2995a9c95c78edbe11d4966d521 changes then the `.git-blame-ignore-revs` is not working.

felixschurk:
@tbabej
I mainly added you, as I assume that the GitHub confirmation for the pre-commit will land by you :)

It seams as it already did pick up. 

https://results.pre-commit.ci/run/github/120080377/1721843918.R63gBvLPQsGx1IBTDjX8Bw

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# initial applied pre-commit run
+7dfe39322167e2995a9c95c78edbe11d4966d521
```
tbabej: :+1: Happy to see this, mass refactorings like this wreak havoc on git blame.

tbabej:
> Currently there should be some request for someone to add the pre-commit.ci to this repository, which should then pick up automatically the changed on every PR.

> It seams as it already did pick up.

Correct, I installed this ~20 mins ago upon getting the request.

tbabej:
@felixschurk given the size of the changeset, could you please share the steps to reproduce [7dfe393](https://github.com/GothenburgBitFactory/taskwarrior/pull/3566/commits/7dfe39322167e2995a9c95c78edbe11d4966d521) from [5d3e9c6](https://github.com/GothenburgBitFactory/taskwarrior/pull/3566/commits/5d3e9c6fb6f61d0edd96a314254fd2af49fb05d4)?

This might be relatively straightforward but I would like to reproduce the process and check for any diffs.

Necior:
> I am not completely sure how we need to take care of the merge/rebase, as when the commit id of [7dfe393](https://github.com/GothenburgBitFactory/taskwarrior/commit/7dfe39322167e2995a9c95c78edbe11d4966d521) changes then the `.git-blame-ignore-revs` is not working.

One approach is to merge the PR from the command line using `git merge --ff-only`. This way there will be no changes to the commits (e.g. no date/time changes) and therefore the hashes won't change.

felixschurk:
> @felixschurk given the size of the changeset, could you please share the steps to reproduce [7dfe393](https://github.com/GothenburgBitFactory/taskwarrior/pull/3566/commits/7dfe39322167e2995a9c95c78edbe11d4966d521) from [5d3e9c6](https://github.com/GothenburgBitFactory/taskwarrior/pull/3566/commits/5d3e9c6fb6f61d0edd96a314254fd2af49fb05d4)?
> 
> This might be relatively straightforward but I would like to reproduce the process and check for any diffs.

Yes sure.
When you checkout `5d3e9c6` and have `pre-commit` installed. Then it's step four, to run it against all files. 

[Quick start Guide](https://pre-commit.com/#quick-start)

Necessary commands:
```
pip install pre-commit
pre-commit run --all-files
```

In the main directory.

src/Variant.cpp
```diff
@@ -25,937 +25,955 @@
...
-  {
-  case type_boolean:
-    switch (right._type)
-    {
-    case type_boolean:                             return !left._bool && right._bool;
-    case type_integer:  left.cast (type_integer);  return left._integer  > right._integer;
-    case type_real:     left.cast (type_real);     return left._real     > right._real;
-    case type_string:   left.cast (type_string);   return left._string   > right._string;
-    case type_date:     left.cast (type_date);     return left._date     > right._date;
-    case type_duration: left.cast (type_duration); return left._duration > right._duration;
```
djmitche, line 518: There are a number of snippets like this, where non-standard formatting was used in a way that arguably made the code easier to understand. Should we try to retain that? Is that difficult? Worth the effort?

I don't have a particular opinion -- I just want to be sure that this was considered.

felixschurk: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

Yes, definitely it is possible to turn it off for certain regions.

Enclosing the lines with a comment 
```
// clang-format off 
code without formatting

// clang-format on
```

I also don't have any particular opinion.

djmitche: @tbabej I'm curious what you think, as perhaps you wrote some of those nicely-formatted sections. Maybe there are a few that are particularly helpful that we should preserve?

But without any strong opinions I think we should just let clang-format do its thing.

felixschurk:
Just as its also interfering with other repos (see https://github.com/GothenburgBitFactory/taskchampion/pull/434) we should double check how the `pre-commit.ci` is added to this organisation. 

Unfortunately, I can not do that.

djmitche:
I'm getting ready to work on a pretty major overhaul of the interface with TaskChampion, using cxx instead of hand-rolled FFI. But, that will be a big change to the C++ and conflict mightily with this.

My changes can't land until after the next release of Taskwarrior, but it'd be nice to work on them without worrying about a major conflict-resolution project.

Thoughts on rebasing and landing this soon? I think we haven't heard any strong opinions, so we shouldn't worry about my question above.

Also, I'm not sure what you were referring to regarding pre-commit.ci -- it seems to have passed here!

felixschurk:
I am ready to land it, I can rebase it today and update the `.git-blame-ignore` and we can land.
I would need to force push this branch however, at least to have one clean commit with all the initial changes.

I was referring to the interference with e.g. the taskchampion repo, where the CI gsiñef fue to missing `pre-commit` file. The intention was to add pre-commit only to Taskwarrior, not any other repo of the organisation.

As there are no strong opinions on the formatting of special parts of the code, I would also follow your suggestion - just let `clang-format` do the job.

djmitche:
Sounds good. A force push is fine!

felixschurk:
Done, currently can be merged if everything is alright :)

tbabej:
IMHO we should not block this PR by preserving manually formatted blocks, if so desired, those can be pulled back from the history if someone volunteers to do this in the future.

Good to go from my end!

tbabej (approved):

djmitche (approved):

Necior:
I'm glad this got merged. Mainly because autoformatters lower the entry threshold to development of a project.

Well done!

PR #3568, opened 2024-07-27
Remove encoding declaration in Python files

Necior:
As per
https://docs.python.org/3.11/reference/lexical_analysis.html#encoding-declarations, the default encoding of Python files is UTF-8. In fact, it's been the default encoding since Python 3.0 (released in 2008).


ryneeverett:
> In fact, it's been the default encoding since Python 3.0 (released in 2008).

To be fair, taskwarrior made the switch in 2020 - #2357. That doesn't mean we have to wait another decade to remove the cruft though. :)

djmitche (approved):

PR #3582, opened 2024-08-04
Run all C++ tests from a single executable

### LINKED ISSUE #3579: Faster linking of tests
During development, if you're making changes and trying to run the entire test suite, linking each of the C++ tests takes a long time.  I suspect this is because there are a bunch of libraries to link against (specified twice, for vague reasons).

It'd be great if we could do most of that linking once, such as into a single `test_support` library, and then just link each of the test files against that.

djmitche:
This makes the C++ tests link a lot faster (just once instead of once per test), and as a bonus they run quite a bit faster too!


doc/devel/contrib/development.md
```diff
@@ -52,9 +52,9 @@ cmake --build build-clang
 ## Run the Test Suite:
 For running the test suite [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html) is used.
 Before one can run the test suite the `task_executable` must be built.
-After that also the `build_tests` target must be build, which can be done over:
+After that also the `test_runner` target must be build, which can be done over:
 ```sh
-cmake --build build --target build_tests
```
djmitche, line 57: It'd be nice if we could keep `build_tests` as an over-arching "build everything needed for tests" target, but it seems that [add_custom_target](https://cmake.org/cmake/help/latest/command/add_custom_target.html) makes a target that is _always considered out of date_ and thus would build on every CMake run -- that seems wasteful. Ideas?

djmitche:
Huh, just stumbled across [create_test_sourcelist](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html#handling-a-large-number-of-tests) which seems to do a similar thing. Should I refactor this PR to use it instead?

felixschurk:
Yes I guess it would make sense to use the `create_test_sourcelist` if this is the standard way of combining them into one test. 

I just knew from other projects where we used Google Test Framework, that they also combine them into one main executable and thus link them way faster.

djmitche:
OK, refactored to use `create_test_sourcelist`, which functions as documented!

felixschurk (approved):

PR #3583, opened 2024-08-04
Actually run shell tests

### LINKED ISSUE #3580: Run shell tests
We have a number of tests written in shell:
```
⸩ ls -1 test/*.test.sh
test/tw-1637.test.sh
test/tw-1643.test.sh
test/tw-1688.test.sh
test/tw-1715.test.sh
test/tw-1718.test.sh
test/tw-1804.test.sh
test/tw-1883.test.sh
test/tw-1895.test.sh
test/tw-1938.test.sh
test/tw-2124.test.sh
test/tw-2189.test.sh
test/tw-2257.test.sh
test/tw-2386.test.sh
test/tw-2392.test.sh
test/tw-2429.test.sh
test/tw-2451.test.sh
test/tw-2514.test.sh
test/tw-2530.test.sh
test/tw-2550.test.sh
test/tw-2581.test.sh
test/tw-3102.test.sh
test/tw-3109.test.sh
```
and none of them run by default in `ctest`!

djmitche:
Two of these used EXPFAIL which, because nothing is interpreting the TAP output, does not work. So, that functionality is removed, and the expected-to-fail bits are commented out or removed.

There was a conditional on the filename in `bash_tap.sh` which caused it to not actually do anything and just run the test as a simple shell script. That, too, has been removed.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -226,9 +226,9 @@ set (shell_SRCS
 configure_file(bash_tap.sh bash_tap.sh COPYONLY)
 configure_file(bash_tap_tw.sh bash_tap_tw.sh COPYONLY)
 
-foreach (test ${shell_SRCS})
-  add_test(NAME ${test}
-    COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/${test}
+foreach (shell_Test ${shell_SRCS})
+  add_test(NAME ${shell_Test}
+    COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/${shell_Test}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     )
-endforeach(test)
+endforeach(shell_Test)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -60,9 +60,6 @@ foreach (src_FILE ${test_SRCS})
     )
 endforeach (src_FILE)
 
-configure_file(bash_tap.sh bash_tap.sh COPYONLY)
-configure_file(bash_tap_tw.sh bash_tap_tw.sh COPYONLY)
-
 add_subdirectory(basetest)
 add_subdirectory(simpletap)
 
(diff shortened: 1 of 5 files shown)


felixschurk:
Looks good. 
Initially I think they where not ment to be added to the test suite - however if they run and test something I do not see the reason not having them included.

https://github.com/GothenburgBitFactory/taskwarrior/tree/develop/test#architecture

test/CMakeLists.txt
```diff
@@ -198,3 +195,40 @@ foreach (python_Test ${pythonTests})
...
+  tw-2550.test.sh
+  tw-2581.test.sh
+  tw-3102.test.sh
+  tw-3109.test.sh
+)
+
+configure_file(bash_tap.sh bash_tap.sh COPYONLY)
+configure_file(bash_tap_tw.sh bash_tap_tw.sh COPYONLY)
+
+foreach (test ${shell_SRCS})
```
felixschurk: We could name it
```cmake
foreach (shell_Tests ...)
```
which makes searching maybe a bit easier.

test/tw-2124.test.sh
felixschurk, line 1: https://cmake.org/cmake/help/latest/prop_test/WILL_FAIL.html

Maybe we could still have the "original" test, however mark it as failing?
Anyway thats also a good way. :)

felixschurk (approved):

test/tw-2124.test.sh
djmitche: Either EXPFAIL or WILL_FAIL basically assumes the whole test suite (shell script) will fail, which means we lose any information from the parts of the suite that do pass.

```shell
# Filtering for tag with dash does not work right now
export EXPFAIL=true

task add test +one-two
task +one-two list
```
I don't think a regression test for functionality not-yet-implemented is helping, and
```shell
# this does not work without journal.info
# Check that 2 hour interval is reported by task info
task info | grep -F "Start deleted"
[[ ! -z `task info | grep -F "Start deleted (duration: 2:00:00)."` ]]
```
that one's on me and will hopefully be fixed with introduction of per-task ops.

djmitche:
>  Initially I think they where not ment to be added to the test suite - however if they run and test something I do not see the reason not having them included.

I didn't realize this! But, I agree -- if we've got the tests, let's run them.

PR #3588, opened 2024-08-07
Use TaskChampion 0.7.0, now via cxx instead of hand-rolled FFI

djmitche:
TC 0.7.0 introduces a new `TaskData` type that maps to Taskwarrior's `Task` type more cleanly. It also introduces the idea of gathering lists of operations and "committing" them to a replica.

A consequence of this change is that TaskChampion no longer automatically maintains dependency information, so Taskwarrior must do so, with its `TDB2::dependency_sync` method. This method does a very similar thing to what TaskChampion had been doing, so this is a shift of responsibility but not a major performance difference.

Cxx is .. not great. It is missing a lot of useful things that make a general-purpose bridge impractical:

 - no support for trait objects
 - no support for `Option<T>` (https://github.com/dtolnay/cxx/issues/87)
 - no support for `Vec<Box<..>>`

As a result, some creativity is required in writing the bridge, for example returning a `Vec<OptionTaskData>` from `all_task_data` to allow individual `TaskData` values to be "taken" from the vector.

That said, Cxx is the current state-of-the-art, and does a good job of ensuring memory safety, at the cost of some slightly awkward APIs.

Subsequent work can remove the "TDB2" layer and allow commands and other parts of Taskwarrior to interface directly with the `Replica`.


.cargo/config.toml
```diff
@@ -1 +0,0 @@
-../config.toml
```
djmitche, line 1: No longer needed with the removal of `cargo xtask`

src/CMakeLists.txt
```diff
@@ -28,6 +26,7 @@ add_library (task STATIC CLI2.cpp CLI2.h
                   rules.cpp
                   sort.cpp
                   util.cpp util.h)
+target_link_libraries(task taskchampion-cpp)
```
djmitche, line 29: This means that the include path for `taskchampion-cpp` will be used to compile the sources in `task`.

src/Context.cpp
```diff
@@ -681,6 +683,11 @@ int Context::initialize(int argc, const char** argv) {
     rc = 2;
   }
 
+  catch (rust::Error& err) {
+    error(err.what());
+    rc = 2;
+  }
```
djmitche, line 689: cxx throws `rust::Error` on error.

src/commands/CmdImport.h
```diff
@@ -31,6 +31,7 @@
 #include <JSON.h>
 
 #include <string>
+#include <unordered_map>
```
djmitche, line 34: Another include-what-you-use issue, I think?

src/TDB2.h
```diff
@@ -30,25 +30,21 @@
 #include <FS.h>
 #include <Task.h>
 #include <stdio.h>
-#include <tc/Replica.h>
-#include <tc/WorkingSet.h>
+#include <taskchampion-cpp/lib.h>
 
 #include <map>
+#include <optional>
```
djmitche, line 36: This was included by one of the `tc/*.h`, masking the failure to include it directly in this file ([include-what-you-use](https://include-what-you-use.org/))

djmitche:
Felix, please have a look at the C++ bits -- which is most of it, I suppose!

Ryne, you saw the development of TaskData in Rust, so maybe you can spot some bugs in how I've used it here? Also, hopefully this will be easier to modify than the hand-rolled FFI stuff!

src/taskchampion-cpp/build.rs
```diff
@@ -0,0 +1,6 @@
+#[allow(unused_must_use)]
+fn main() {
+    cxx_build::bridge("src/lib.rs");
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=src/lib.rs");
```
ryneeverett, line 5: What's up with these println's?

src/TDB2.cpp
```diff
@@ -217,59 +224,56 @@ void TDB2::get_changes(std::vector<Task>& changes) {
...
-    // Has the side-effect of freeing undo_ops.
-    replica.commit_undo_ops(undo_ops, NULL);
-  } else {
-    replica.free_replica_ops(undo_ops);
+    if (!replica()->commit_reversed_operations(std::move(undo_ops))) {
+      std::cout << "Could not undo: other operations have occurred.";
+    }
   }
-  replica.rebuild_working_set(false);
+  // Note that commit_reversed_operations rebuilds the working set.
```
ryneeverett: Nitpick: Strikes me as odd to put this comment *after* commit_reversed_operations may or may not have been called.

src/TDB2.cpp
```diff
@@ -217,59 +224,56 @@ void TDB2::get_changes(std::vector<Task>& changes) {
...
-        throw std::string("Can't undo UndoPoint.");
-        break;
-      default:
-        throw std::string("Can't undo non-operation.");
-        break;
+    std::string uuid = static_cast<std::string>(op.get_uuid().to_string());
+    if (op.is_create()) {
+      std::cout << "Create " << uuid;
+    } else if (op.is_delete()) {
+      std::cout << "Delete " << uuid;
```
ryneeverett: Previously this was "Delete {description}" rather than "Delete {uuid}".

src/taskchampion-cpp/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+        /// Get the uuid with the given index, if any exists. Returns the nil UUID if
+        /// there is no task at that index.
+        fn by_index(&self, index: usize) -> Uuid;
+
+        /// Get the index for the given uuid, or zero if it is not in the working set.
+        fn by_uuid(&self, uuid: Uuid) -> usize;
+
+        /// Get the entire working set, using ids as indices into the vector. Empty indices have
+        /// the nil UUID.
+        fn all_uuids(&self) -> Vec<Uuid>;
```
ryneeverett: "using ids as indices into the vector" <-- I don't understand this.

src/taskchampion-cpp/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    #[test]
+    fn expire_tasks() {
+        let mut rep = new_replica_in_memory().unwrap();
+        let mut operations = new_operations();
+        add_undo_point(&mut operations);
+        create_task(uuid_v4(), &mut operations);
+        create_task(uuid_v4(), &mut operations);
+        create_task(uuid_v4(), &mut operations);
+        rep.commit_operations(operations).unwrap();
+        rep.expire_tasks().unwrap();
```
ryneeverett, line 889: I take it we're just testing that expire_tasks doesn't panic?

ryneeverett (approved):

src/commands/CmdImport.h
felixschurk: Most likely yes. :D

felixschurk (commented):
From what I understand of the C++ code it looks good to me. :)

src/taskchampion-cpp/build.rs
djmitche: `build.rs` runs as part of the `cargo build` process, and these are directives to that process to re-run `build.rs` if these files change.

src/TDB2.cpp
djmitche: Good catch!

src/taskchampion-cpp/src/lib.rs
djmitche: Yes, most of the tests here are basically "doesn't panic" or a kind of quick test for the cxx implementation. The `test/tc.test.cpp` does a better job of actually asserting that things are correct.

djmitche:
Thanks! I'll push an update, but I want to hold off landing this for a little while longer until I'm confident we don't need a v3.1.1 release.

src/taskchampion-cpp/build.rs
ryneeverett: So cargo must listen on the stdout of its subprocesses for directives. Seems a little odd but I guess it works.

FYI, I seem to be unable to resolve conversations again.

ryneeverett (approved):

src/taskchampion-cpp/build.rs
djmitche: Just this particular subprocess 

https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

PR #3591, opened 2024-08-09
Update google-cloud-auth to drop ring@0.16.20

jcgruenhage:
ring@0.16.20 doesn't build on ppc and risc-v, and updating
google-cloud-auth pulls in a newer version of jsonwebtoken,
which in turn depends on a newer version of ring that we depend on
already either way.

Readable "diff":
```
taskwarrior on  get-rid-of-old-ring via △ v3.30.2 via 🦀 v1.80.1 
❯ cargo update google-cloud-auth
    Updating crates.io index
     Locking 3 packages to latest compatible versions
    Removing base64 v0.13.1
    Updating google-cloud-auth v0.13.0 -> v0.13.2 (latest: v0.16.0)
    Updating jsonwebtoken v8.3.0 -> v9.3.0
    Updating pem v1.1.1 -> v3.0.4
    Removing ring v0.16.20
    Removing spin v0.5.2
    Removing untrusted v0.7.1
    Removing winapi v0.3.9
    Removing winapi-i686-pc-windows-gnu v0.4.0
    Removing winapi-x86_64-pc-windows-gnu v0.4.0
note: pass `--verbose` to see 90 unchanged dependencies behind latest
```

Chimera Linux has been carrying a patch that contained the changes made by `cargo update` for a while. I've reduced that patch because I'm not sure whether a broad `cargo update` is even wanted.

In my opinion it would make sense to bump all dependencies to their latest compatible versions before each release using `cargo update`, and maybe even to run `cargo upgrade --incompatible` every now and then to get the latest and greatest. That last one some times causes more work though, as it also includes breaking changes of dependencies.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -29,7 +29,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: "1.70.0" # MSRV
+          toolchain: "1.73.0" # MSRV
           override: true
 
       - uses: actions-rs/cargo@v1.0.3
@@ -82,7 +82,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: "1.70.0" # MSRV
+          toolchain: "1.73.0" # MSRV
           override: true
 
       - uses: actions-rs/cargo@v1.0.3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,12 +121,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
-
-[[package]]
-name = "base64"
 version = "0.21.7"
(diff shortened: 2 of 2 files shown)


jcgruenhage:
Ah, oh, I see. MSRV is 1.70.0 which is not supported for these updated deps anymore...

djmitche:
I'm happy to bump the MSRV, given a reason. What version is required?

jcgruenhage:
It would be 1.73.0

djmitche:
That sounds fine. `cargo xtask msrv` should be able to update that.

jcgruenhage:
I'll package cargo xtask for chimera linux later then and go ahead and do that :D

djmitche:
It should just run within the repo where you're making the PR

jcgruenhage:
Indeed, somehow I don't need cargo-xtask installed.

Still, it doesn't do it:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/xtask msrv`
Error: xtask: unknown xtask
```

jcgruenhage:
My short dig seems to point to the xtask being lost when taskchampion got moved out of the taskwarrior repo?

djmitche:
#3592

djmitche (approved):

PR #3598, opened 2024-08-14
Do not count undo operations in the 'would be reverted..' message

### LINKED ISSUE #3596: Undo test failing
https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/10346231185/job/28634586464#step:4:1

It looks like sometimes there are 6 operations to undo?

djmitche:
These undo operations previously were not included in the operations list at all. They are not shown in the list of operations, so do not include them in the count preceding the list. Fixes #3596.


ryneeverett (approved):
How does this differ from `undo_ops.size() - 1`? Could there ever be more than one UndoPoint here?

djmitche:
They should be the same, but the existing loop in the function is written without that assumption, so I did the same.

I am about to file a good-first-bug to improve this display, which will re-write the entire function.

PR #3604, opened 2024-08-25
Add some instructions to the MSRV

### LINKED ISSUE #3592: Figure out how to keep MSRV in sync with taskchampion

djmitche:
There is no easy way to determine the MSRV for TaskChampion, other than somehow pulling the right version of the source and grepping for it. In practice, if we update the `taskchampion` dependency to one that has a higher MSRV, we'll get a build error and find this comment. And if we get an error building Taskwarrior due to an old MSRV (for example if something changes on `crates.io`) then we will also find this comment.

This also removes some superfluous dependency versions from the root workspace. `src/taskchampion-cpp/Cargo.toml` specifies versions directly.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -29,6 +29,9 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
+          # If this version is old enough to cause errors, or older than the
+          # TaskChampion MSRV, bump it to the MSRV of the currently-required
+          # TaskChampion package; if necessary, bump that version as well.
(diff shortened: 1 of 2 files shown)


ryneeverett (approved):
This is an improvement but i do wonder -- would we expect to have frequent rust version related problems if we just set it to 'stable'?

djmitche:
I suspect that `cxx` may occasionally have a dependency on relatively recent Rust features, so it's possible. And, if we don't test it, we wouldn't know until release.

PR #3615, opened 2024-09-02
Refine INSTALL file

tobiolo:
CMake can also abstract the install procedure from the underlying Makefile


djmitche (approved):
Nice, thank you!

PR #3619, opened 2024-09-09
Exclude attributes starting with tag_

### LINKED ISSUE #3607: `task info` shows `tag_..` properties
Taskwarrior stores tags as `tag_<tagname>` properties with a nonempty value `x`. The `task info` command should filter those out and just show the tags, but it does not.

```
✺ sofia ~/p/taskwarrior ⸨issue3605⸩
⸩ task add task with a tag +thetagvalue
TASKRC override: /home/dustin/p/taskwarrior/.taskrc
Created task 6.
✺ sofia ~/p/taskwarrior ⸨issue3605⸩
⸩ task 6
TASKRC override: /home/dustin/p/taskwarrior/.taskrc

Name             Value                                 
ID               6                                     
Description      task with a tag                       
Status           Pending                               
Entered          2024-08-26 21:49:35 (2s)
Last modified    2024-08-26 21:49:35 (2s)              
Tags             thetagvalue
Virtual tags     LATEST PENDING READY TAGGED UNBLOCKED 
UUID             f06b2b7f-a80c-4b79-bf4c-58a0d43d80dc
Urgency          0.8                                   
[tag_thetagvalue x]

    tags    0.8 *    1 =    0.8
                         ------
                            0.8
```

The `[tag_thetagvalue x]` line is the one that shouldn't be there. Just omtting properties beginning with `tag_` should be sufficient to fix this issue.

gagankonana:
Fixes [#3607](https://github.com/GothenburgBitFactory/taskwarrior/issues/3607)

### DIFF
--- a/src/commands/CmdInfo.cpp
+++ b/src/commands/CmdInfo.cpp
@@ -374,7 +374,7 @@ int CmdInfo::execute(std::string& output) {
     // the context.columns map.
     for (auto& att : all) {
       if (att.substr(0, 11) != "annotation_" && att.substr(0, 5) != "tags_" &&
-          att.substr(0, 4) != "dep_" &&
+          att.substr(0, 4) != "dep_" && att.substr(0, 4) != "tag_" &&
           Context::getContext().columns.find(att) == Context::getContext().columns.end()) {
         row = view.addRow();
         view.set(row, 0, '[' + att);
--- a/src/commands/CmdInfo.cpp
+++ b/src/commands/CmdInfo.cpp
@@ -373,8 +373,8 @@ int CmdInfo::execute(std::string& output) {
     // Show any orphaned UDAs, which are identified by not being represented in
     // the context.columns map.
     for (auto& att : all) {
-      if (att.substr(0, 11) != "annotation_" && att.substr(0, 5) != "tags_" &&
-          att.substr(0, 4) != "dep_" && att.substr(0, 4) != "tag_" &&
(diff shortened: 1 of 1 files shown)


src/commands/CmdInfo.cpp
```diff
@@ -374,7 +374,7 @@ int CmdInfo::execute(std::string& output) {
     // the context.columns map.
     for (auto& att : all) {
       if (att.substr(0, 11) != "annotation_" && att.substr(0, 5) != "tags_" &&
-          att.substr(0, 4) != "dep_" &&
+          att.substr(0, 4) != "dep_" && att.substr(0, 4) != "tag_" &&
```
djmitche: At one point we accidentally used `tags_` instead of `tag_`, and I think fixing that bug was what introduced #3607. please remove the `"tags_"` check here.

djmitche (changes requested):
Thanks! I was going to suggest adding a test in https://github.com/GothenburgBitFactory/taskwarrior/blob/develop/test/info.test.py but I suspect that would be `assertNotIn("tag_", ..)` which really isn't that interesting as a test. So, please just address the comment and we can get this merged!

src/commands/CmdInfo.cpp
gagankonana: Got it. Updated the PR.

djmitche (approved):

djmitche:
Thank you!

PR #3621, opened 2024-09-13
Throw error if trying to set end date for a pending task

### LINKED ISSUE #3584: Modifying the end property silently fails
```
$ task add smile
Created task 1.

$ task
ID Age  Description Urg
 1 1s   smile          0

$ task 1 mod end:-7d
Modifying task 1 'smile'.
Modified 1 task.

$ task
ID Age   Description Urg
 1 18s   smile          0
```

I expected that modifying the end property will succeed. Or, alternatively, that it will fail with some explanation and non-zero exit code.

I observed this behavior in both 2.6.2 and 3.0.2.

gagankonana:
Fixes #3584 
-throw an error if we are trying to set an end date for pending tasks
-tests

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -1476,7 +1476,8 @@ void Task::validate(bool applyDefault /* = true */) {
   // Pending tasks cannot have an end date, remove if present
   if ((status == Task::pending) && (get("end") != "")) {
     remove("end");
-    throw format("Could not modify task {1}. You cannot set an end date on a pending task.", this->id);
+    throw format("Could not modify task {1}. You cannot set an end date on a pending task.",
+                 this->id);
   }
 
   // Provide a modified date unless user already specified one.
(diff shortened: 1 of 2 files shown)


src/Task.cpp
```diff
@@ -1474,7 +1474,11 @@ void Task::validate(bool applyDefault /* = true */) {
     setAsNow("end");
 
   // Pending tasks cannot have an end date, remove if present
-  if ((status == Task::pending) && (get("end") != "")) remove("end");
+  if ((status == Task::pending) && (get("end") != "")) {
+    remove("end");
+    throw format("Could not modify task {1}. You cannot set an end date on a pending task.",
+                 this->id);
```
djmitche, line 1480: To answer your [question from the issue](https://github.com/GothenburgBitFactory/taskwarrior/issues/3584#issuecomment-2347969857), in this context I would prefer a warning since tasks might be `pending` but have an `end` from some other cause, and this will cause Taskwarrior to return an error even when the rnuning `task` command did not modify either of these properties.

That said, as you have seen there are lots of other `throw`s in this function, and I think adding another is the right course. We can think about converting them to warnings later.

PR #3622, opened 2024-09-13
Check if end date is being set to a pending task

### LINKED ISSUE #3584: Modifying the end property silently fails

gagankonana:
Fixes #3584 
- throw error if end date is being set to a pending task
- add a test for the bug

### DIFF
--- a/src/commands/CmdModify.cpp
+++ b/src/commands/CmdModify.cpp
@@ -116,6 +116,11 @@ void CmdModify::checkConsistency(Task &before, Task &after) {
 
   if (before.has("recur") && (!after.has("recur") || after.get("recur") == ""))
     throw std::string("You cannot remove the recurrence from a recurring task.");
+
+  if ((before.getStatus() == Task::pending) && (after.getStatus() == Task::pending) &&
+      (after.get("end") != ""))
+    throw format("Could not modify task {1}. You cannot set an end date on a pending task.",
+                 before.identifier(true));
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/test/modify.test.py
+++ b/test/modify.test.py
@@ -56,6 +56,20 @@ class TestBug1763(TestCase):
         self.assertIn("Modified 0 tasks.", out)
 
 
+class TestBug3584(TestCase):
+    def setUp(self):
+        self.t = Task()
+
+    def test_mod_pending_task_end_date(self):
+        """Adding end date for a pending task throws an error"""
+        self.t("add foo")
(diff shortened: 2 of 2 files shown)


src/commands/CmdModify.cpp
```diff
@@ -116,6 +116,11 @@ void CmdModify::checkConsistency(Task &before, Task &after) {
 
   if (before.has("recur") && (!after.has("recur") || after.get("recur") == ""))
     throw std::string("You cannot remove the recurrence from a recurring task.");
+
+  if ((before.getStatus() == Task::pending) && (after.getStatus() == Task::pending) &&
+      (after.get("end") != ""))
+    throw format("Could not modify task {1}. You cannot set an end date on a pending task.",
+                 before.identifier(true));
```
djmitche, line 123: To answer your [question from the issue](https://github.com/GothenburgBitFactory/taskwarrior/issues/3584#issuecomment-2347969857), in this context I would prefer a warning since tasks might be `pending` but have an `end` from some other cause, and this will cause Taskwarrior to return an error even when the rnuning `task` command did not modify either of these properties.

That said, as you have seen there are lots of other `throw`s in this function, and I think adding another is the right course. We can think about converting them to warnings later.

djmitche (approved):

src/commands/CmdModify.cpp
djmitche: Oh, sorry, I admit I copy-pasted this comment from the other PR and didn't notice it's now in `CmdModify`! I think this is a much better fix :)

djmitche (approved):

gagankonana:
@djmitche awesome! Do you know why PR failed one of the checks - tests (Ubuntu 20.04, ubuntu-latest, ubuntu2004)? I am just curious.

djmitche:
I peeked at that failure, and saw some errors in installing Ubuntu packages that the job depended on, so I assume something was wrong there, unrelated to your changes. If I see that failure in another PR, I'll look into what's going on.

gagankonana:
Got it. Thanks for checking!

Issue #3625, opened 2024-09-17
task sync relies on embedded/integrated certificate store and does not obey system store


snicker:
Recently corporate IT implemented TLS inspection using ZScaler and suddently my ability to synchronize tasks with my taskwarrior sync server is failing with an "unknown issuer" error, using Taskwarrior 3.1.0  and the `task sync` command:

```
Failed to synchronize with server: Server Error: https://[redacted]/v1/client/get-child-version/[redacted]: Connection Failed: tls connection init failed: invalid peer certificate: UnknownIssuer
```

output from openssl, showing that the system level certs handle intercepting and modifying trust chain (though this is not something I am really fond of)

```
$ openssl s_client -connect [redacted]:443           
Connecting to [redacted]
CONNECTED(00000005)
depth=3 C=US, ST=California, L=San Jose, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Root CA, emailAddress=support@zscaler.com
verify return:1
depth=2 C=US, ST=California, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Intermediate Root CA (zscalertwo.net), emailAddress=support@zscaler.com
verify return:1
depth=1 C=US, ST=California, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Intermediate Root CA (zscalertwo.net) (t)
verify return:1
depth=0 CN=*.[redacted], O=Zscaler Inc., OU=Zscaler Inc.
verify return:1
---
Certificate chain
 0 s:CN=*.[redacted], O=Zscaler Inc., OU=Zscaler Inc.
   i:C=US, ST=California, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Intermediate Root CA (zscalertwo.net) (t)
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Sep 17 14:16:07 2024 GMT; NotAfter: Sep 28 03:46:25 2024 GMT
 1 s:C=US, ST=California, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Intermediate Root CA (zscalertwo.net) (t)
   i:C=US, ST=California, O=Zscaler Inc., OU=Zscaler Inc., CN=Zscaler Intermediate Root CA (zscalertwo.net), emailAddress=support@zscaler.com
[...]

PR #3626, opened 2024-09-17
Require reqwest feature `rustls-tls-native-roots`

djmitche:
This uses the system's trust roots, rather than those from WebPKI. This is important for enterprise users, where additional enterprise-specific roots may be defined on the system. See #3625.

### DIFF
--- a/src/taskchampion-cpp/Cargo.toml
+++ b/src/taskchampion-cpp/Cargo.toml
@@ -10,6 +10,8 @@ crate-type = ["staticlib"]
 [dependencies]
 taskchampion = "0.7.0"
 cxx = "1.0.124"
+# Not used directly, but adds the native-roots feature. #3625.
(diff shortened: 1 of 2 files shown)


src/taskchampion-cpp/Cargo.toml
```diff
@@ -10,6 +10,8 @@ crate-type = ["staticlib"]
 [dependencies]
 taskchampion = "0.7.0"
 cxx = "1.0.124"
+# Not used directly, but adds the native-roots feature. #3625.
+reqwest = { version = ">= 0.11.27", default-features = false, features = ["rustls-tls-native-roots"] }
```
djmitche, line 14: This "fake dependency" is probably not the best way to solve the issue. Probably, we should mirror this feature from taskchampion into reqwest (on which it depends). But this will do for testing.

djmitche:
See https://github.com/GothenburgBitFactory/taskchampion/issues/460

PR #3627, opened 2024-09-22
Ignore SIGPIPE

### LINKED ISSUE #3624: task sync  crashes with SIGPIPE
With somewhat invalid server sync configuration (`sync.server.url` pointing to a service that is not a taskchampion server) running `task sync` results in the process crashing with unix return code 141 (SIGPIPE), without printing any output to stdout/stderr. I assume the bug is in the external taskwarrior rust lib.

`task` versoin 3.1.0

djmitche:
This replicates what the Rust runtime does, and matches what Rust code expects, for example when writing to a socket which is no longer connected to the remote end.

Fixes #3624.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,6 +29,7 @@
 
 #include <Context.h>
 #include <rust/cxx.h>
+#include <signal.h>
 
 #include <cstring>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


djmitche:
Other bits of the TW code use `signal` instead of `sigaction` so it seems reasonable here.

felixschurk (approved):
Looks clean. :)

Issue #3644, opened 2024-10-11
task delete should also purge


ksandvik:
#### To request a feature...

* Clearly describe the feature.
task delete should have an option or a flag to purge at the same time

* Clearly state the use case. We are only interested in use cases, do not waste time with implementation details or suggested syntax.

In 99 of 100 cases I want to purge a task at the same time as I delete it.  Just now I don't actually have a single work flow where I need to keep a task around after it has been deleted. There's no need to also make this verbal with yes/no responses.

* Please see our notes on [How to request a feature](https://taskwarrior.org/docs/features/)

djmitche:
Huh, I didn't realize `task purge` wouldn't purge a non-deleted task. Maybe we could add an option to it somehow? I'd suggest `-f` but none of the other Taskwarrior commands take `-` style flags. Or, maybe that could be a config option?

There is an option to automatically purge deleted tasks after a period of time -- is that sufficient for you? Why do they need to be purged?

ksandvik:
As mentioned, I have yet to find a case in my taskwarrior workflow where I don't both delete and purge an unnecessary task, doing this all the time. Now I could do do task delete; task purge but then I'm getting all the prompts for asking if I really want to do it (which I want).

I agree that task should never have option flags, rather key words, maybe 
task delete permanently or task delete now or task delete purge or something similar.

djmitche:
Why do you need to purge and not just delete tasks?

smemsh:
I almost always purge after deleting also, but there are some use cases where I wouldn't, such as when doing infrequent "maintenance" or trimming of my task database, I probably wouldn't purge until the end of making such a pass, once I was satisfied I hadn't deleted anything by mistake.

Purge is useful is so they are removed from the database entirely, and don't show up at all anymore, for example for test/junk tasks that should end up like they never existed.  There are some reasons enumerated in #3399 and I was very happy that `purge` was restored in 3.x, as this was a barrier to upgrade for me.

However I would not want purge to happen automatically, just like I don't want gc to happen automatically; it is better off explicit, so I hope we can have this only be an option (if implemented) and not change it for everyone.  The idea of making `purge` automatically force a deletion first might make sense though...

(Also, not sure if still the case with the TaskChampion backend, but note that 2b882605 seems to imply that gc is done automatically upon purge, and that's another reason not to want to auto-purge when deleting, because it would forcibly change the ID numbers.)

djmitche:
[...]

PR #3646, opened 2024-10-13
Issue warnings instead of errors for 'weird' tasks

djmitche:
This gets TW a little closer to the admonition that any set of properties and values should be accepted, even if it doesn't make sense. In other words, if some other TaskChampion replica starts producing "weird" tasks, TW should do not error out while listing tasks, but should do its best to interpret the task as given.

This came up in #3644.


felixschurk (approved):
Sorry, took way too long. Looks good :)

PR #3647, opened 2024-10-13
Install uuid-dev in GitHub action

djmitche:
This started failing in #3646. I'm guessing that the latest Ubuntu image dropped this library, perhaps accidentally.


djmitche:
Causing a lot of PR failures so I'll merge this now.

Issue #3623, opened 2024-09-14
Update libshared to get new weekstart


djmitche:
From https://github.com/GothenburgBitFactory/libshared/pull/81

> The new ISO8601 behavior for Datetime::weekstart == 1 will be a breaking change for users relying on existing week date parsing. This is unavoidable for weekstart == 1. However if the user wants weekstart == 0, the old behavior is still possible if Taskwarrior/Timewarrior were modified to set Datetime::weekstart before libshared calls (which they currently do not). Taskwarrior could propagate its rc.weekstart value (used only for task calendar so far) and propagate this value to libshared, for example.

smemsh:
@djmitche Thanks for picking this up, I was planning to make an issue to bump libshared in Taskwarrior, now that the weekstart change was merged.

Note that the libshared change only affects week number parsing, whereas sonw/sopw/etc should still use Monday regardless of `Datetime::weekstart` (as they did before).  I did not look into tackling that but it would be necessary to make those respect `Datetime::weekstart` as well in order for support to be truly complete.

It looks like all these functions would need changes for that:

- initializeSopw
- initializeSow
- initializeSonw
- initializeEopw
- initializeEow
- initializeEonw
- initializeSopww
- initializeSonww
- initializeSoww
- initializeEopww
- initializeEonww
- initializeEoww

When were you targeting 3.2 ? I might be able to look into doing this, depending on your timeline.

djmitche:
Ah, that sounds like more work than I anticipated! If you can work on it, that would be great. I think it's fine for this to be "done when it's done" so don't rush to get it into 3.2.0.

smemsh:
It's not strictly necessary to get those additional changes in.  Merely updating the libshared peg and propagating `rc.weekstart` to the library would still be a better situation than today, it just would leave the sonw/sopw/etc aliases hard-configured to Monday weeks, but they already are (so at least parsing weeks would be made better, currently they are hard-configured to Sunday weeks).

Let me look at how hard I think it would be to fix the aliases too and I'll get back to you on whether an intermediate release doing the more limited change is warranted in the meantime.  Give me a little bit, thanks.

djmitche:
Sounds good, take your time :)

elig0n:
Before `libshared` there was this code which took `weekstart` into consideration:
https://github.com/GothenburgBitFactory/taskwarrior/commit/903b5b34d41cecf1328c3270868f22a22e5e8827

[...]

Issue #2373, opened 2020-12-15
Year 2038 Problem


tasksoldier:
Hello,

I tried to add a task with a due date after 2038:

_task add "Retirement" due:2048-06-01
'-1820392096' is not a valid date in the '' format._

I expected that a new task is added (at least on the 64 Bit system I am using).

Is there already (I know some time is left) a strategy for the taskwarrior how to cope with the year 2038 problem?

Thank you in advance

kwoot:
Funny. And, also, you must be extremely busy, to already have work to do 18.5 years from now. :smile:

tbabej:
@tasksoldier Thanks for bringing this issue up! We are aware of this problem with TW somehow using 32-bit timestamp values. The mechanism is currently unclear but the problem is tracked in #2017. I'd recommend watching that issue if you're interested in support of 64-bit timestamp values.

tasksoldier:
I debugged the source code and I developed a patch that resolves my problem at least on my 64 Bit system. The regression test passes with my patch. How should I proceed? (I am a new to github)

tbabej:
You could either fork the repository to make a pull request, i.e. (see [the following guide](https://www.freecodecamp.org/news/how-to-make-your-first-pull-request-on-github-3/)) or alternatively format your patch using `git format-patch` and post the patch here.

tbabej:
@tasksoldier See the instructions above. Let us know if we can help!

tasksoldier:
Thank you very much for your kind support. I attached two patches. One for libshared. The other for taskwarrior itself. Both are required to resolve the problem. If you think the patches need improvement, then please let me know.


[0001-Bug-Patch-for-libshared-to-resolve-issue-2373.patch.txt](https://github.com/GothenburgBitFactory/taskwarrior/files/5705108/0001-Bug-Patch-for-libshared-to-resolve-issue-2373.patch.txt)
[0001-Bug-Patch-for-taskwarrior-to-resolve-issue-2373.patch.txt](https://github.com/GothenburgBitFactory/taskwarrior/files/5705109/0001-Bug-Patch-for-taskwarrior-to-resolve-issue-2373.patch.txt)

tasksoldier:
@tbabej I did both. I made two pull requests. And as a fallback I posted the patches here.

tbabej:
Thank you @tasksoldier, I incorporated your changes in 2.6.0.

PR #3651, opened 2024-10-14
Limit the allowed epoch timestamps

### LINKED ISSUE #3444: Bare integers are accepted for due: and wait:, but cause issues later
#### To report a bug...

`task add recur:daily due:10 wait:6 foobar`

The goal was to add a daily task that recurs at 10am.

* What actually happened?

`task list` hangs and thousands of tasks are created while it hangs.

Screenshot 2024-05-06 at 3 27 22 PM

* Paste the output of the ```task diag``` command.

```
task 3.0.2
   Platform: Darwin

Compiler
    Version: Apple LLVM 15.0.0 (clang-1500.3.9.4)
 Compliance: C++17

Build Features
      CMake: 3.29.2
 Build type: Release

Configuration
       File: /Users/jerry/.taskrc (found), 1469 bytes, mode 100644
       Data: /Users/jerry/.task (found), dir, mode 40755
    Locking: Enabled
         GC: Enabled
    $EDITOR: nvim
Hooks
     System: Enabled
   Location: /Users/jerry/.task/hooks
             (-none-)

Tests
   Terminal: 252x117
       Dups: Scanned 8450 tasks for duplicate UUIDs:
             No duplicates found
 Broken ref: Scanned 8450 tasks for broken references:
             No broken references found
```
---

Running `task delete` will go through the tasks one by one and delete them but it takes forever if there are thousands of them. What would the effect be of deleting the rows in the `tasks` table? I didn't try this because I see there are other tables referencing that data, but could that be a 'last resort' way of recovering?

Thanks!

lanker:
The code for parsing epoch timestamps when displaying tasks only supports values between year 1980 and 9999. Previous to this change, it was possible to set e.g., the due timestamp to a value outside of these limits, which would make it impossible to later on show the task.

With this change, we only allow setting values within the same limits used by the code for displaying tasks.

Fixes https://github.com/GothenburgBitFactory/taskwarrior/issues/3444

### DIFF
--- a/src/columns/ColTypeDate.cpp
+++ b/src/columns/ColTypeDate.cpp
@@ -190,7 +190,11 @@ void ColumnTypeDate::modify(Task& task, const std::string& value) {
   if (value != "" && evaluatedValue.get_date() == 0)
     throw format("'{1}' is not a valid date in the '{2}' format.", value, Variant::dateFormat);
 
-  task.set(_name, evaluatedValue.get_date());
+  time_t epoch = evaluatedValue.get_date();
+  if (epoch < EPOCH_MIN_VALUE || epoch >= EPOCH_MAX_VALUE) {
+    throw format("'{1}' is not a valid date.", value);
+  }
+  task.set(_name, epoch);
 }
 
 ////////////////////////////////////////////////////////////////////////////////


lanker:
This PR is depending on https://github.com/GothenburgBitFactory/libshared/pull/83. Not sure how you handle bumping the version of libshared that is used in Taskwarrior...?

djmitche:
Let's wait until that PR is merged, and then include a change to the submodule version in this PR.

That will also affect https://github.com/GothenburgBitFactory/taskwarrior/issues/3623 but I think there's no problem in rolling both of those at once.

lanker:
Just some additional info if anyone ends up here in the future:

Looking at the commit and PR history, I couldn't find any reason for the lower limit when [parsing timestamps](https://github.com/GothenburgBitFactory/libshared/blob/a63fa350413f598b3b55355a6460b7fd2efb2e69/src/Datetime.cpp#L697) (the upper limit was introduced from [this issue](https://github.com/GothenburgBitFactory/taskwarrior/issues/2373)), so my initial approach was to remove it completely. However, it seems like more strings than potential epoch timestamps are sent to `Datetime::parse_epoch`.

For example when doing `task 3 info`, `Datetime::parse_epoch` is called with the value `3`, and must return `false` for it to work. So somehow, we need to differentiate between numbers being a timestamp and numbers being something else, like a task ID. I think that's why there's a lower limit. I guess it would've been nice to only trying to parse strings that we know should be timestamps, but I'm sure there's a good reason for it to work as it does.

(So there might be problems if a task ID would be > 315532800, but let's hope no one ends up there... :))

djmitche:
OK, the libshared update has been merged into Taskwarrior, if you'd like to rebase this PR.

djmitche:
If you're ready for a review on this, please add me and/or @smemsh as a reviewer!

lanker:
> If you're ready for a review on this, please add me and/or @smemsh as a reviewer!

I don't think I have enough permissions to be able to add reviewers. According to the [documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review): "To assign a reviewer to a pull request, you will need write access to the repository.".

So please add yourself and anyone else that should have a look at it, thanks!

smemsh (commented):
Have you considered just using `Datetime::parse_epoch()` itself and checking the return value? Since the range check is already in there.  This way it wouldn't have to be duplicated in both taskwarrior and libshared.  Just an idea...

smemsh:
> Have you considered just using `Datetime::parse_epoch()` itself

But since you're using those defines, and it's just a simple bounds check, it seems valid enough in any case to do it like your current patch already does.

So, the code looks sane to me @djmitche ...

lanker:
> Have you considered just using `Datetime::parse_epoch()` itself and checking the return value? Since the range check is already in there. This way it wouldn't have to be duplicated in both taskwarrior and libshared. Just an idea...

That would actually be a better solution, so I took a brief look at it, not much luck though... Firstly, `Datetime::parse_epoch()` is private to `Datetime`, and I'm not comfortable enough with the code (or C++ in general) to start changing visibilities. Secondly, `Datetime::parse_epoch()` takes a `Pig` as argument. To create a `Pig` you need a `string`, I couldn't find any string that would work for us here.

I think it could've been nice to have a `bool Datetime::check_epoch(time_t epoch)` that could be used both here and in libshared, but might be a bit overkill for just a bounds check.

djmitche (approved):

PR #3655, opened 2024-10-19
Add SECURITY.md

lauft:


SECURITY.md
djmitche, line 1: There should be a link on "here" and the text should mention Taskwarrior, not TC.

djmitche (approved):
thanks!

PR #3666, opened 2024-11-05
Add `color.calendar.scheduled` to `no-color.theme`

d2718nis:
## Issue
The built-in `no-color.theme` file lacks a `color.calendar.scheduled setting`, causing this color to remain active with the default settings rather than being reset.

screenshot-2024-11-05-14-22-25

> There is a `no-color.theme` file that has no color at all, and while this sounds useless, it allows you to start with no color, and add your own. If you are building your own theme, this is what you would start with.

[Documentation](https://taskwarrior.org/docs/themes/#default-theme)

## Other considerations
It may be beneficial to add this setting to other built-in themes as well. However, including `color.calendar.scheduled=` directly would reset any color currently assigned by default. As an alternative, I could add a commented-out placeholder line `# color.calendar.scheduled=` to avoid interfering with existing color settings while providing a hint for future customization. Let me know if this approach would be useful.

### DIFF
--- a/doc/rc/bubblegum-256.theme
+++ b/doc/rc/bubblegum-256.theme
@@ -84,6 +84,7 @@ color.calendar.due=color0 on rgb325
 color.calendar.due.today=color0 on rgb404
 color.calendar.holiday=color15 on rgb102
 color.calendar.overdue=color0 on color5
+color.calendar.scheduled=
 color.calendar.today=color15 on rgb103
 color.calendar.weekend=gray12 on gray3
 color.calendar.weeknumber=rgb104
--- a/doc/rc/dark-16.theme
+++ b/doc/rc/dark-16.theme
@@ -86,6 +86,7 @@ color.calendar.due=white on red
 color.calendar.due.today=bold white on red
 color.calendar.holiday=black on bright yellow
 color.calendar.overdue=black on bright red
+color.calendar.scheduled=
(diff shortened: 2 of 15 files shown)


djmitche (commented):
This fix looks good, but please do add it to the other themes as well.

Could it just be set to the default in those themes?

d2718nis:
Done, added `color.calendar.scheduled=` to all themes which would reset the color to default `foreground on background`.

I also noticed that some of the colors are [assigned](https://github.com/GothenburgBitFactory/taskwarrior/blob/0ff78447322a10e24c18b4ce2950745cf17962a1/src/Context.cpp#L267) during the default `taskrc` generation process. Maybe all color-related configuration entries should be moved to a `default.theme` file and just included via `include default.theme`?

djmitche:
> Done, added `color.calendar.scheduled=` to all themes which would reset the color to default `foreground on background`.
> 
> I also noticed that some of the colors are [assigned](https://github.com/GothenburgBitFactory/taskwarrior/blob/0ff78447322a10e24c18b4ce2950745cf17962a1/src/Context.cpp#L267) during the default `taskrc` generation process. Maybe all color-related configuration entries should be moved to a `default.theme` file and just included via `include default.theme`?

That sounds like a good idea!

djmitche (approved):
Thanks!

Issue #3371, opened 2024-04-16
new undo report not human readable


Andonome:
On version `3.0.0`, I did `task undo`, which produces this:

```
The following 8 operations would be reverted:
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    status: <empty> -> pending
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    scheduled: <empty> -> 1713340800
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    priority: <empty> -> M
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    modified: <empty> -> 1713281078
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    entry: <empty> -> 1713281078
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    due: <empty> -> 1713346200
- Update 90f0b9d3-9ed7-4919-933a-1f3748b1829f
    description: <empty> -> raise taskwarrior issue
- Create 90f0b9d3-9ed7-4919-933a-1f3748b1829f
The undo command is not reversible.  Are you sure you want to revert to theyesevious state? (yes/no) 
``` 

This is a surprising amount of feedback for 'undo last change'.
Of course it's all true, but I feel that the previous behaviour gave more meaningful output.

I've changed the description to 'raise taskwarrior issue'.

ryneeverett:
This issue corresponds to [this TODO](https://github.com/GothenburgBitFactory/taskwarrior/blob/0944c73716e39c3c930fa8514e556b644072de86/src/TDB2.cpp#L243).

PR #3671, opened 2024-11-06
Restore support for` task info` journal

### LINKED ISSUE #3397: task 3 start/end times not saved to task
I just upgraded to version 3.0 from Manjaro linux and finished the import to run task start which added the virtual tag ACTIVE and the Start: field to that task info but afterwards I ran task stop and then task info didn't show the summary of work time like it used show on previous version.

### LINKED ISSUE #3528: Restore support for the journal in `task info`
Per taskrc(5):

```
       journal.info=1
              When  enabled,  this setting causes a change log of each task to be dis‐
              played by the ’info’ command. Default value is "1".
```

However, doing this doesn't show any change log:
```
$ task add test
$ task <id> modify new description
$ task <id> info
```

djmitche:
This was a long time coming!

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -244,8 +244,7 @@ void TDB2::revert() {
 
 ////////////////////////////////////////////////////////////////////////////////
 bool TDB2::confirm_revert(rust::Vec<tc::Operation>& undo_ops) {
-  // TODO Use show_diff rather than this basic listing of operations, though
-  // this might be a worthy undo.style itself.
+  // TODO: convert to Operation and use that type for display, similar to CmdInfo.
 
   // Count non-undo operations
   int ops_count = 0;
--- a/test/tw-1999.test.py
+++ b/test/tw-1999.test.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+###############################################################################
+#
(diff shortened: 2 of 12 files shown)


test/tw-1999.test.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+# Copyright 2006 - 2021, Tomas Babej, Paul Beckingham, Federico Hernandez.
```
djmitche, line 5: This file is reappearing from ddd367232e2dbd7b7666a680e10af8a4094f3f03, so no update to the copyright.

src/TDB2.cpp
```diff
@@ -244,8 +244,7 @@ void TDB2::revert() {
 
 ////////////////////////////////////////////////////////////////////////////////
 bool TDB2::confirm_revert(rust::Vec<tc::Operation>& undo_ops) {
-  // TODO Use show_diff rather than this basic listing of operations, though
-  // this might be a worthy undo.style itself.
+  // TODO: convert to Operation and use that type for display, similar to CmdInfo.
```
felixschurk, line 247: should we directly create an issue for that?

felixschurk (approved):

src/TDB2.cpp
djmitche: It's #3371.

PR #3673, opened 2024-11-08
include bubblegum-256.theme in default .taskrc

### LINKED ISSUE #3586: Update initial generation of taskrc when no one is present
When using the v3.1.0 version without an existing `.taskrc` and following the prompt of generating one, there is the #3505 theme missing as well as files for `completed.data` and `pending.data` are generated.

One could update the initial autogenerated `.taskrc` to reflect the latest status.

I assume it's a good first issue, as one needs to search a bit where the file is generated and then modify it to reflect the up to date changes - nothing to complex I hope.

djmitche:
Fixes #3586.


felixschurk (commented):
Nice :)

PR #3680, opened 2024-11-13
Fix compile with libc++ 18

licy183:
See termux/termux-packages#22242


djmitche (approved):
Thanks for this! I'm looking forward to rocking 3.2.0 on my Pixel :)

I filed #3682 to hopefully prevent this in the future.

PR #3684, opened 2024-11-15
add cpp standard flag to cmake

### LINKED ISSUE #3682: Test with C++ 17 in CI
C++17 is sort of the C++ equivalent of our MSRV, so we should be testing with that version to detect regressions like #3680.

I suspect this means just `--std=c++17` in the clangd command line, but that's a wild guess.

felixschurk:
See documentation in CMake.
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

Fixes #3682


djmitche (approved):
Nice, thanks!

PR #3688, opened 2024-11-17
Set CMake C++ standard

### LINKED ISSUE #3687: Build fails on Apple Clang
The build of v3.2.0 fails on Apple Clang, due to invalid C++ standard compiler flag. The issue is visible in Homebrew/homebrew-core#197503:
```
  cd /tmp/task-20241112-7838-da7q3n/task-3.2.0/build/src/commands && /opt/homebrew/Library/Homebrew/shims/mac/super/clang++  -I/tmp/task-20241112-7838-da7q3n/task-3.2.0 -I/tmp/task-20241112-7838-da7q3n/task-3.2.0/src -I/tmp/task-20241112-7838-da7q3n/task-3.2.0/src/commands -I/tmp/task-20241112-7838-da7q3n/task-3.2.0/src/columns -I/tmp/task-20241112-7838-da7q3n/task-3.2.0/src/libshared/src -I/tmp/task-20241112-7838-da7q3n/task-3.2.0/build/src/taskchampion-cpp/corrosion_generated/cxxbridge/taskchampion-cpp/include -Wall -Wextra -Wsign-compare -Wreturn-type -std=c++17 -stdlib=libc++  -Wno-return-type-c-linkage -O3 -DNDEBUG -std=gnu++11 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk -MD -MT src/commands/CMakeFiles/commands.dir/Command.cpp.o -MF CMakeFiles/commands.dir/Command.cpp.o.d -o CMakeFiles/commands.dir/Command.cpp.o -c /tmp/task-20241112-7838-da7q3n/task-3.2.0/src/commands/Command.cpp
  In file included from /tmp/task-20241112-7838-da7q3n/task-3.2.0/src/CLI2.cpp:34:
  In file included from /tmp/task-20241112-7838-da7q3n/task-3.2.0/src/Context.h:36:
  /tmp/task-20241112-7838-da7q3n/task-3.2.0/src/TDB2.h:80:8: error: no template named 'optional' in namespace 'std'
```

The log shows that the C++ standard is set twice, and the latter takes precedence: `-std=c++17 ... -std=gnu++11`. The reason for this is that Taskwarrior puts `-std=c++17` to `CMAKE_CXX_FLAGS` directly, but its dependency, corrosion, uses `target_compile_features(.. cxx_std_11)`. When CMake evaluates the C++ standard flag to use, it is unaware of the intention to use C++17. As a result, incorrect C++ standard is used and some toolchains/stdlibs do not expose all the needed features. This issue started in commit 4ff63a796.

kietavainen:
Instead of setting `-std` compiler flag directly, set the `CMAKE_CXX_STANDARD` variable. This lets CMake know the required C++ standard and evaluate the final compiler flag correctly, taking into account compile features set by `target_compile_features()`.

This change preserves the existing behavior, where compiler extensions are disabled for other targets than Cygwin.

Fixes #3687.


kietavainen:
Previously, the standard was set with `-std=c++17` flag, except on Cygwin where `-std=gnu++17` was used. I wasn't sure if this was really intentional, so I preserved the earlier behavior. CMake enables compiler extensions by default, so by removing the `CMAKE_CXX_EXTENSIONS` settings, it would use  `-std=gnu++17` on all platforms. Please let me know if that would be acceptable, and I can update the PR.

djmitche:
@felixschurk what do you think? This flag was also set, but just for CI, in 01ced3238e87b71647fcdf8179907524e97ff54a.

felixschurk:
I think that's the better solution which also enables to easily upgrade the standard in one place in the future.

We can then remove the flag introduced in 01ced32.

felixschurk:
In regards to the difference with cygwin, I unfortunately have no clue.

djmitche:
Great! I'll revert that commit then.

djmitche:
> In regards to the difference with cygwin, I unfortunately have no clue.

I think WSL is much more common these days, so maybe this doesn't matter?

PR #3694, opened 2024-11-22
Updating the URL for Migration Documentation

geoffpaulsen:
When I run `task list` with a 2.x`.taskrc` (using taskwarrior 3.1.0), I get the following message:
```
task list
Found existing '*.data' files in /Users/geoffp/.task
  Taskwarrior's storage format changed in 3.0, requiring a manual migration.
  See https://github.com/GothenburgBitFactory/taskwarrior/releases.
No matches.
```

Unfortunately this URL is confusing as I couldn't find any migration data there (it now routes to changelog for 3.2.0).

A better approach is to point the user directly at the user documentation for the 3.0 migration.


djmitche (approved):
Thanks, good fix!

PR #3696, opened 2024-11-22
Use repository owner instead of actor to log into GitHub CR

lauft:
On every successful run of branches `develop` and `stable`, this GitHub action creates a Docker image with the respective Taskwarrior version. Currently the push to the GitHub container registry (ghcr.io) fails because of access rights.

This PR should fix this by using the repository owner instead of the GitHub actor.

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -23,6 +23,10 @@ jobs:
       id-token: write
 
     steps:
+      - name: Create lowercase repository name
+        run: |
+          GHCR_REPOSITORY="${{ github.repository_owner }}"
+          echo "REPOSITORY=${GHCR_REPOSITORY,,}" >> ${GITHUB_ENV}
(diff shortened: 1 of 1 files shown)


felixschurk (approved):
Thanks a lot. :)

lauft:
Currently this image is only (?) used in the [task-timewarrior-hook](https://github.com/GothenburgBitFactory/task-timewarrior-hook) repository, along with the one created for Timewarrior, to test the hook script against the different `stable`/`develop` combinations of `task` and `timew`.

I wonder whether those images could (should?) be used by users outside the GBF organization, e.g. for testing their TW/TI tools... 🤔

felixschurk:
@kdheepak 
I recall reading something about that the last few days in context to `taskwarrior-tui`, where the long build time of taskwarrior was taking most of the CI time. I am sorry if I have that somehow wrong.

PR #3699, opened 2024-11-24
Only nag to read news when there's news to read

### LINKED ISSUE #3499: Do not suggest running `task news` if there's no news to share
Currently, if we make a release with no new news items, users will be reminded to run `task news` because the current version is newer than their `news.version` config. However, when they do so they will be told nothing is new. That's not a great UX - better to just not remind them if there are no new news items.

This will probably require moving the reminder out of `CmdCustom.cpp` and into `CmdNews.cpp`, which can calculate whether there are any news items to show. In fact, it's probably best to factor all of that out into a `src/News.{h,cpp}` which both `CmdCustom.cpp` and `CmdNews.cpp` refer to.

djmitche:
Fixes #3499.

### DIFF
--- a/src/commands/CmdCustom.cpp
+++ b/src/commands/CmdCustom.cpp
@@ -28,6 +28,7 @@
 // cmake.h include header must come first
 
 #include <CmdCustom.h>
+#include <CmdNews.h>
 #include <Context.h>
 #include <Filter.h>
 #include <Lexer.h>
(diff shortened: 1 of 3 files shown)


felixschurk (approved):

felixschurk:
Ups, sorry was not even assigned 😅

djmitche:
> Ups, sorry was not even assigned 😅

I'm always happy for a review :)

PR #3706, opened 2024-11-28
Include prefix to error message regarding opening replica

### LINKED ISSUE #2708: Crash if data.location refers to an unexisting and unwritable path
task 2.6.1 (package with debian unstable). Start with a taskrc containing a single line:

````
data.location=/heyyy/
````

Get a segmentation fault when running:

````
% task next
segmentation fault
````

as task is attempting to create a directory where it has no right (infinitely, when looking at the stack trace?).

task should exit with an error message in such case. ``task diag`` will fail as well.

### PR DESCRIPTION
**Pull request by djmitche**
Fixes #2708.

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -34,6 +34,7 @@
 #include <Table.h>
 #include <format.h>
 #include <main.h>
+#include <rust/cxx.h>
 #include <shared.h>
 #include <signal.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/TDB2.cpp [hunk 1]
```diff
@@ -51,7 +52,11 @@ static void dependency_scan(std::vector<Task>&);
 
 ////////////////////////////////////////////////////////////////////////////////
 void TDB2::open_replica(const std::string& location, bool create_if_missing) {
-  _replica = tc::new_replica_on_disk(location, create_if_missing);
+  try {
+    _replica = tc::new_replica_on_disk(location, create_if_missing);
+  } catch (rust::Error& err) {
+    throw format("Could not open replica at {1}: {2}", location, err.what());
```
Necior, line 58: Would it make sense to modify https://github.com/GothenburgBitFactory/taskwarrior/blob/0b286460b69042b1ed409b981e42ea34c6172574/src/taskchampion-cpp/src/lib.rs#L468 instead so all callers get this extra context in the error message?

Necior (approved):

src/TDB2.cpp
djmitche: Yeah, good point!

djmitche: In fact, it should be fixed in TaskChampion..

PR #3708, opened 2024-11-28
Check Datetime addition when performing recurrence

### LINKED ISSUE #81: [TW-19] no maximum recur leads to an infinite loop in gc from overflow of int
_Scott Kostyshak on 2012-09-01T23:17:31Z says:_ 

 To reproduce:

1. task add brush my teeth recur:70years due:today
2. task

(2) puts me into an infinite loop in generateDueDates.

djmitche:
If the Datetime arithmetic overflows, this just stops recurrence entirely. This means that `recur:70y` will basically not recur. And, it means that a repeated recurrence of a more reasonable period (say, `recur:2d`) will stop in 2038 on 32-bit systems.

Fixes #81.

### DIFF
--- a/src/recur.cpp
+++ b/src/recur.cpp
@@ -51,11 +51,9 @@
 #include <optional>
 #include <sstream>
 
-namespace {
-
 // Add a `time_t` delta to a Datetime, checking for and returning nullopt on integer overflow.
-std::optional<Datetime> checked_add(Datetime& base, time_t delta) {
-  // Datetime::operator+ takes an integer delta, so chek that range
+std::optional<Datetime> checked_add_datetime(Datetime& base, time_t delta) {
+  // Datetime::operator+ takes an integer delta, so check that range
   if (static_cast<time_t>(std::numeric_limits<int>::max()) < delta) {
     return std::nullopt;
   }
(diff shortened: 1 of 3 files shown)


src/recur.cpp
```diff
@@ -47,8 +47,28 @@
...
 #include <iomanip>
 #include <iostream>
+#include <limits>
+#include <optional>
 #include <sstream>
 
+namespace {
+
+// Add a `time_t` delta to a Datetime, checking for and returning nullopt on integer overflow.
+std::optional<Datetime> checked_add(Datetime& base, time_t delta) {
```
Necior: As a pure function, it is is a good candidate for a unit test.

src/recur.cpp
```diff
@@ -47,8 +47,28 @@
...
+  // Datetime::operator+ takes an integer delta, so chek that range
```
Necior: s/chek/check/

Necior (approved):

src/recur.cpp
djmitche: I wish there was a way to test private functions. I'll just make this one public to test it.

Necior (approved):

Issue #3705, opened 2024-11-27
cxxbridge related build errors when building offline (Nix)


doronbehar:
Hello,

On NixOS, when we build packages, we disallow the usage of the internet. Since dependencies of Rust packages such as this and of course the source code itself of packages must be fetched from the internet, we have a mechanism to download in advance all sources required, and use a hash computed from the contents of the fetched sources, to verify these don't change.

So for example the rust dependencies of this package, are all downloaded to a cargo cache, and get a hash. Then this cache is used when we build Taskwarrior (offline, just like any other package).

Weirdly, we get this issue when using the above mechanism (which works well for an uncountable number of packages):

```
error: download of config.json failed
Caused by:
  failed to download from `https://index.crates.io/config.json`
Caused by:
  [6] Could not resolve hostname (Could not resolve host: index.crates.io)
make[2]: *** [src/taskchampion-cpp/CMakeFiles/cxxbridge_v1.0.128.dir/build.make:74: corrosion/cxxbridge_v1.0.128/bin/cxxbridge] Error 101
make[1]: *** [CMakeFiles/Makefile2:617: src/taskchampion-cpp/CMakeFiles/cxxbridge_v1.0.128.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

Do you have an idea what exactly that CMake file is trying to download? We should be able to fix our build if CMake will be able to detect that nothing should be downloaded, as the files are already there - because we'll put them there ourselves.

djmitche:
`index.crates.io` is where the index to crates is stored. The corrosion extension we use is basically calling `cargo build` with some arguments, I think the `--offline` option skips this check. [Corrosion makes `FROZEN` available](https://corrosion-rs.github.io/corrosion/usage.html), which should help in this case - it is equivalent to `--locked --offline`.

It is probably easy to just patch that option into `src/taskchampion-cpp/CMakeLists.txt` but I'd be happy to merge a PR to add a CMake OPTION to do the same, if you'd refer.

doronbehar:
I tried the following:

```diff
diff --git i/src/taskchampion-cpp/CMakeLists.txt w/src/taskchampion-cpp/CMakeLists.txt
index a0d4a42ab..709b1dc3d 100644
--- i/src/taskchampion-cpp/CMakeLists.txt
+++ w/src/taskchampion-cpp/CMakeLists.txt
[...]

PR #3712, opened 2024-12-01
Include cxxbridge-cmd in Cargo.lock, check version consistency

djmitche:
This adds cxxbridge-cmd to Cargo.lock per https://github.com/dtolnay/cxx/issues/1407#issuecomment-2509136343, as linked from #3705.

It adds an MSRV to `src/taskchampion-cpp/Cargo.toml` so that the version of `Cargo.lock` is stil compatible with the MSRV.

It additionally adds a check of the Cargo metadata for all of the cxx* versions agreeing, and for the MSRVs agreeing.

cc: @doronbehar

### DIFF
--- a/.github/workflows/metadata-check.sh
+++ b/.github/workflows/metadata-check.sh
@@ -1,0 +1,73 @@
+#! /bin/bash
+
+# Check the 'cargo metadata' for various requirements
+
+set -e
+
+META=$(mktemp)
+trap 'rm -rf -- "${META}"' EXIT
+
+cargo metadata --locked --format-version 1 > "${META}"
+
+get_version() {
+  local package="${1}"
+  jq -r '.packages[] | select(.name == "'"${package}"'") | .version' "${META}"
+}
+
+get_msrv() {
+  local package="${1}"
+  jq -r '.packages[] | select(.name == "'"${package}"'") | .rust_version' "${META}"
+}
+
+# check that the cxx packages all have the same version
+check_cxx_versions() {
+  local cxx_version=$(get_version "cxx")
+  local cxx_build_version=$(get_version "cxx-build")
+  local cxxbridge_cmd_version=$(get_version "cxx-build")
+  local cxxbridge_flags_version=$(get_version "cxxbridge-flags")
+  local cxxbridge_macro_version=$(get_version "cxxbridge-macro")
+
+  ok=true
+  echo "Found cxx version ${cxx_version}"
+  if [ "${cxx_version}" != "${cxx_build_version}" ]; then
+    echo "Found differing cxx-build version ${cxx_build_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_cmd_version}" ]; then
+    echo "Found differing cxxbridge-cmd version ${cxxbridge_cmd_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_flags_version}" ]; then
+    echo "Found differing cxxbridge-flags version ${cxxbridge_flags_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_macro_version}" ]; then
+    echo "Found differing cxxbridge-macro version ${cxxbridge_macro_version}"
+    ok = false
+  fi
+
+  if ! $ok; then
+    echo "All cxx packages must be at the same version. Fix this in src/taskchampion-cpp/Cargo.toml."
+    exit 1
+  else
+    echo "✓ All cxx packages are at the same version."
+  fi
+}
+
+check_msrv() {
+  local taskchampion_msrv=$(get_msrv taskchampion)
+  local taskchampion_lib_msrv=$(get_msrv taskchampion-lib)
+
+  echo "Found taskchampion MSRV ${taskchampion_msrv}"
+  echo "Found taskchampion-lib MSRV ${taskchampion_lib_msrv}"
+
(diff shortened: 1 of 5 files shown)


.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,48 @@
...
+  ok = false
+fi
+if [ "${cxx_version}" != "${cxxbridge_flags_version}" ]; then
+  echo "Found differing cxxbridge-flags version ${cxxbridge_flags_version}"
+  ok = false
+fi
+if [ "${cxx_version}" != "${cxxbridge_macro_version}" ]; then
+  echo "Found differing cxxbridge-macro version ${cxxbridge_macro_version}"
+  ok = false
+fi
```
doronbehar: For loop?

doronbehar (commented):
Looks pretty good! Thanks for making the effort with that GitHub workflow.

doronbehar:
Haven't tested this locally yet.

.github/workflows/metadata-check.sh
djmitche: Meh, maybe if we add more :)

.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,73 @@
...
+get_msrv() {
+  local package="${1}"
+  jq -r '.packages[] | select(.name == "'"${package}"'") | .rust_version' "${META}"
+}
+
+# check that the cxx packages all have the same version
+check_cxx_versions() {
+  local cxx_version=$(get_version "cxx")
+  local cxx_build_version=$(get_version "cxx-build")
+  local cxxbridge_cmd_version=$(get_version "cxx-build")
```
dtolnay, line 26:
```suggestion
  local cxxbridge_cmd_version=$(get_version "cxxbridge-cmd")
```

.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,73 @@
...
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_cmd_version}" ]; then
+    echo "Found differing cxxbridge-cmd version ${cxxbridge_cmd_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_flags_version}" ]; then
+    echo "Found differing cxxbridge-flags version ${cxxbridge_flags_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_macro_version}" ]; then
```
dtolnay, line 44: This is already enforced by https://github.com/dtolnay/cxx/blob/1.0.124/Cargo.toml#L26.

.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,73 @@
...
+  echo "Found cxx version ${cxx_version}"
+  if [ "${cxx_version}" != "${cxx_build_version}" ]; then
+    echo "Found differing cxx-build version ${cxx_build_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_cmd_version}" ]; then
+    echo "Found differing cxxbridge-cmd version ${cxxbridge_cmd_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_flags_version}" ]; then
```
dtolnay, line 40: This is already enforced by https://github.com/dtolnay/cxx/blob/1.0.124/Cargo.toml#L31.

.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,73 @@
...
+check_cxx_versions() {
+  local cxx_version=$(get_version "cxx")
+  local cxx_build_version=$(get_version "cxx-build")
+  local cxxbridge_cmd_version=$(get_version "cxx-build")
+  local cxxbridge_flags_version=$(get_version "cxxbridge-flags")
+  local cxxbridge_macro_version=$(get_version "cxxbridge-macro")
+
+  ok=true
+  echo "Found cxx version ${cxx_version}"
+  if [ "${cxx_version}" != "${cxx_build_version}" ]; then
```
dtolnay, line 32: This is already enforced by https://github.com/dtolnay/cxx/blob/1.0.124/Cargo.toml#L34.

dtolnay (commented):
Of the 4 cxx version checks in metadata-check.sh, the only 1 that matters is not implemented correctly. That one should be fixed and the other 3 should be removed.

.github/workflows/metadata-check.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "Found differing cxx-build version ${cxx_build_version}"
+    ok = false
+  fi
+  if [ "${cxx_version}" != "${cxxbridge_cmd_version}" ]; then
+    echo "Found differing cxxbridge-cmd version ${cxxbridge_cmd_version}"
+    ok = false
+  fi
```
dtolnay, line 39: I published cxx 1.0.133 which automatically puts a correct version of cxxbridge-cmd into the lockfile of any crate that depends on cxx (https://github.com/dtolnay/cxx/pull/1408). Could you try upgrading, reverting the taskchampion-cpp/Cargo.toml change from this PR, and see if cxxbridge-cmd stays in the lockfile? If so then this check can also be removed.

.github/workflows/metadata-check.sh
djmitche: 🤦🏼

.github/workflows/metadata-check.sh
djmitche: Thanks! I added https://github.com/GothenburgBitFactory/taskwarrior/pull/3714 to try this.

PR #3713, opened 2024-12-01
Rely on cxx to enforce matching versions

dtolnay:
Since 1.0.133, the cxxbridge-cmd version in a Cargo.lock will stay in sync with cxx automatically (https://github.com/GothenburgBitFactory/taskwarrior/pull/3712#pullrequestreview-2471488084).


djmitche:
I guess I should have read all of my notifications before pushing #3714!

djmitche (approved):
#3714 is basically the same thing, which pretty much implies 'approved' :)

PR #3719, opened 2024-12-07
Add instructions for running against customized TaskChampion

djmitche:
This is often useful when doing work that includes changes in both TC and TW.


felixschurk (approved):
Nice, thanks! :)

PR #3724, opened 2024-12-15
Support importing Taskwarrior v2.x data files

### LINKED ISSUE #3720: Support automatically importing 2.x tasks
When a user upgrades from 2.x to 3.x, they are presented with:
```
$ task
Found existing '*.data' files in /home/$USER/.task
  Taskwarrior's storage format changed in 3.0, requiring a manual migration.
  See https://github.com/GothenburgBitFactory/taskwarrior/releases.
No matches.
```

A better link would be https://taskwarrior.org/docs/upgrade-3/, but that's not the point. These instructions require re-installing 2.x, running `task export`, then running `task import` with 3.x. That's pretty annoying!

We could potentially bring back the TF2 support (TF2 is the text-based task format) just for the purpose of importing. Then users could run something like `task import-task2` to import those tasks automatically.

djmitche:
This should ease the pain of upgrading from v2.x to v3.x.

The TF2.{h,cpp} are copied from 2.6.2's `TDB2.{h,cpp}`, slimmed down to support only the most recent file format (which I believe was introduced somewhere in the 2.5.x series?) and only reading, not writing.

Once this lands, I'll update the importing instructions in tw.org.

Fixes #3720.

### DIFF
--- a/doc/man/task.1.in
+++ b/doc/man/task.1.in
@@ -415,6 +415,11 @@ few example scripts, such as:
 .fi
 
 .TP
+.B task import-v2
+Imports tasks from the Taskwarrior v2.x format. This is used when upgrading from
+version 2.x to version 3.x.
+
+.TP
 .B task log <mods>
 Adds a new task that is already completed, to the task list. It is affected by
 the currently set context.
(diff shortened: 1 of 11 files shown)


felixschurk (approved):
Looks good so far. 

Where there any unit test for this format? I skimmed the repo and did not find any. If there are none I think we are good to go, otherwise we can also add them back again.

djmitche:
Yeah, I don't see any tests from the old version, but it's easy enough to add one.

PR #3734, opened 2024-12-23
Consider news read if `news.version` > current version

Tejada-Omar:
Avoids two installations of taskwarrior with differing versions from constantly nagging and rewriting `news.version`

Note that this potentially isn't the best behavior across major versions updates, although in that case `task news` doesn't display any new items anyways so it's probably fine


djmitche (approved):
I'm starting to think this functionality is more trouble than it's worth. But this patch is good!

aricursion:
Is this patch part of 3.3.0 or will be released later?

Yedfree (approved):

PR #3736, opened 2024-12-25
Enable context based filtering for modify commands

rubieking:
Enable context based filtering for modify commands by setting the parameter to true.

This was tested by running the following commands

```
task add main1
task add main2 +tag1
task context define work "project:Work"
task add project:Work work1
task add project:Work work2 +tag1
task context work

task modify +testtag
task +tag1 modify +testtag2

task context none
task list

## Validate that testtag is set only in work1 and work2
## Validate that testtag2 is set only in work2
```

### DIFF
--- a/src/commands/CmdModify.cpp
+++ b/src/commands/CmdModify.cpp
@@ -48,7 +48,7 @@ CmdModify::CmdModify() {
   _read_only = false;
   _displays_id = false;
   _needs_gc = false;
-  _uses_context = false;
+  _uses_context = true;
   _accepts_filter = true;
   _accepts_modifications = true;
   _accepts_miscellaneous = false;


djmitche:
It looks like there's a test failure here -- maybe just needs an update to that test?

@tbabej, this is precisely a revert of 97dbb0cfe4dc5b7bf09fb6a17281112db0b61329 -- what was the context for that change?

tbabej:
Context being truly an overloaded word here :) the context for the change is that - at the time - only read-only context was implemented, so it should not apply to CmdModify.

Since then I also implemented the writeable context, but chose not to apply it to the CmdModify either, simply because I don't think that is a good user experience. It would be potentially injecting the context to any tasks that were bring modified.

Note that when you have the context active, only tasks you see are the ones that already match the context. With writeable context, any newly added task also falls under the context too.

Hence any situation where you're modifying a non-context task must involve you remembering a task ID/UUID of a task that falls outside of the currently active context and referring to it that way. I would argue that bringing that task under the context by injecting the context mods would be unexpected behaviour.

tbabej:
Okay, I see this concerns itself less with mods and more with the filtering. I have to think about it a bit more.

djmitche:
closing as stale. it's still here if we want to re-vitalize it!

Issue #3418, opened 2024-04-29
mtime of taskchampion.sqlite3 got changed when calling `task list`


chenyuntc:
When I run `task list` the ` taskchampion.sqlite3`'s mtime changed.

This is inconvenience as I use the mtime to sync between workstations.

So is there a way to turn off the changes on the sqlite file when operations has no side-effect (i.e. list, filter)?

LeGuipo:
I confirm. Every time taskwarrior fetches some data, the file seems updated. I also suspect this is the cause of the noticeable delays for taskwarrior to display results.

djmitche:
I don't think TW is acutally modifying the data, but it is opening the DB in r/w mode which, from some quick `strace` analysis, does perform a write of the first block.

So, we should find a way to open the DB in a read-only mode.

djmitche:
#3753 got some way toward this, in the sense that it centralizes one of the two "automatic" changes to the task DB.

The problem is, just about every interesting command runs one of GC or handleRecurrence/handleUntil (I'll refer to these collectively as "updates" here). To support that naively, we need to open the DB in read-write mode for just about every invocation -- and in particular for reports, which is the slow part.

I can think of a few ways to fix this:

1. Open in read-only mode by default, but re-open in read-write mode if there's updates needed.
2. Always open in read-write mode unless some rc.read-only=1 is set, in which case open read-only but don't do any updates.
3. Probabalistically open in read-write mode to perform updates.

The first option is more difficult than it seems, as the GC is performed in TaskChampion and does not currently provide a way to "try" with an interpretable error when updates are actually required.

The second option is pretty bad UX, as it basically means if things are slow you need to distribute `rc.read-only=1` all over the place.

The third option would help with performance, but only sometimes, and introduces non-determinacy into the application, making understanding and debugging more difficult.

Anyone have other ideas?

djmitche:
On reflection, a fourth option similar to 2 is to open in read-write mode unless `rc.gc=0`. That's an existing config option that disables garbage collection, and is already suggested for use from scripts, etc.

me-and:
I'm following this ticket as a specific aspect of #3329, which is currently the thing keeping me on Taskwarrior 2. Certainly for me – and I expect a lot of people who routinely have relatively large task lists – I have `gc=0` and `recurrence=0` in my config, and just run commands to override those every so often.

[...]

PR #3746, opened 2025-01-06
Update to TaskChampion 2.0.2

djmitche:
Note that this does not update Taskwarrior to take advantage of the new read-only mode (for #3418).


djmitche:
Waiting on https://github.com/GothenburgBitFactory/taskchampion/pull/528.

felixschurk (approved):
Is it now an update to 2.0.2 or 2.0.1? Nevertheless, looks good.

djmitche:
It will be to 2.0.2 because 2.0.1 gets expiration wrong.

djmitche:
#3749 for the OpenSUSE failure, which I think is not related to this PR.

Issue #3749, opened 2025-01-06
OpenSUSE Tumbleweed builder failing


djmitche:
See https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/12637918529/job/35213605423#step:3:89

Maybe this is just something transient with tumbleweed?

djmitche:
```
#7 [test-opensuse  2/13] RUN zypper install -y python3 awk coreutils git gcc gcc-c++ cmake make libuuid-devel libfaketime curl
#7 0.641 Retrieving repository 'openSUSE-Tumbleweed-Non-Oss' metadata [................done]
#7 3.991 Building repository 'openSUSE-Tumbleweed-Non-Oss' cache [....done]
#7 3.999 Retrieving repository 'Open H.264 Codec (openSUSE Tumbleweed)' metadata [.................done]
#7 7.547 Building repository 'Open H.264 Codec (openSUSE Tumbleweed)' cache [....done]
#7 7.553 Retrieving repository 'openSUSE-Tumbleweed-Oss' metadata [............................done]
#7 11.17 Building repository 'openSUSE-Tumbleweed-Oss' cache [....done]
#7 14.44 Retrieving repository 'openSUSE-Tumbleweed-Update' metadata [...............done]
#7 16.77 Building repository 'openSUSE-Tumbleweed-Update' cache [....done]
#7 16.78 Loading repository data...
#7 16.95 Reading installed packages...
#7 17.29 'coreutils' is already installed.
#7 17.29 No update candidate for 'coreutils-9.5-4.1.x86_64'. The highest available version is already installed.
#7 17.29 'curl' is already installed.
#7 17.29 No update candidate for 'curl-8.11.1-1.1.x86_64'. The highest available version is already installed.
#7 17.29 'python3' not found in package names. Trying capabilities.
#7 17.29 'awk' not found in package names. Trying capabilities.
#7 17.29 Resolving package dependencies...
#7 17.32 
#7 17.32 Problem: 1: the to be installed libuuid-devel-2.40.2-2.1.x86_64 requires 'glibc-devel', but this requirement cannot be provided
#7 17.32 not installable providers: glibc-devel-2.40-2.1.x86_64[repo-oss]
#7 17.32 
#7 17.32  Solution 1: Following actions will be done:
#7 17.32   do not install libuuid-devel-2.40.2-2.1.x86_64
#7 17.32   do not install gcc-14-2.1.x86_64
[...]

PR #3750, opened 2025-01-06
Disable OpenSUSE builder temporarily

djmitche:
See #3749.


djmitche:
Ah, it seems to be working now!

PR #3753, opened 2025-01-11
Handle 'until' and 'recur' simiar to handling of 'gc'

djmitche:
This centralizes updates to recurrence and 'until' in Command, instead of doing so in each individual command implementation.

This is preparatory to opening the TaskChampion replica in read-only mode. Like `gc`, `handleUntil` and `handleRecurrence` are called for commands marked read-only, yet the write to the task DB.

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -852,7 +852,7 @@ int Context::dispatch(std::string& out) {
     Command* c = commands[command];
     assert(c);
 
-    // The command know whether they need a GC.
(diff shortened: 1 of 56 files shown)


felixschurk (approved):
Looks good, sorry for taking so long. 

When reviewing this PR, it made me a bit realize how much similar code we have laying around. At least when looking only at the PR changes. :D

djmitche:
It's true, especially the commands are very formulaic.

PR #3758, opened 2025-01-17
Fix #3571: Added detailed feedback for successful task synchronization

### LINKED ISSUE #3571: No information message given by "task sync" in 3.0
This is more of a feature request than a bug. 

In previous taskwarrior versions, when I synced my tasks, the command line informed me what had been changed. However, in 3.0, the command "task sync" just outputs "Syncing with Sync server at ..". There is no feedback on whether the sync was successful or not.

My feature request is to return a message that shows something like "successful sync: XX tasks synced ..." when one runs a "task sync" successfully.

antek042:
This pull request includes changes to the `CmdSync::execute` method in `src/commands/CmdSync.cpp` to improve logging by adding messages that confirm successful synchronization with different targets.

Logging improvements:

* Added a log message confirming synchronization with the local server directory.
* Added a log message confirming synchronization with the AWS bucket.
* Added a log message confirming synchronization with the GCP bucket.
* Added a log message confirming synchronization with the remote server.

Fixes #3571


djmitche (dismissed):
Thanks!

Some users embed `task sync` in various automation (cron tasks, shell scripts, etc.), and those users may be unhappy to see output unconditionally appear. Please make these new messages appear only when verbose output is enabled, just like the pre-sync messages.

Also, the pre-sync messages include the type of sync, so it seems redundant to repeat that afterward.

Finally, it'd be nice to give the user some additional information about what happened. TaskChampion doesn't expose the number of new operations it _received_ from the server, but the replica method [`num_local_operations`](https://docs.rs/taskchampion/latest/taskchampion/struct.Replica.html#method.num_local_operations) does give a hint as to what will be _sent_ to the server (and after a successful sync, can be assumed to have been sent). So, let's include that value in the output.

src/commands/CmdSync.cpp
```diff
@@ -159,6 +162,11 @@ int CmdSync::execute(std::string& output) {
     context.tdb2.expire_tasks();
   }
 
+  if (verbose) {
+    out << format("Successfully synchronized {1} operations", replica->num_local_operations())
```
djmitche: This will always be zero, right?

djmitche:
@antek5421 will you have a chance to finish this up?

ryneeverett (approved):

PR #3771, opened 2025-02-02
Remove missed handleRecurrence/handleUntil calls

djmitche:
These were missed in #3753.


felixschurk (approved):
Ups, also did not see them in the other PR. :D

PR #3775, opened 2025-02-07
Move color settings from taskrc to default.theme

### LINKED ISSUE #3669: Move color settings from `taskrc` to `default.theme`
## Issue
As previously [discussed](https://github.com/GothenburgBitFactory/taskwarrior/pull/3666#issuecomment-2457243327), it would be beneficial to separate the default color theme from other configuration settings in `taskrc`. This change would ensure that new color settings don’t unintentionally bypass `no-color.theme` in case it omits them by mistake. Separating color settings would also improve the organization of configuration files.

## Solution
Relocate all `color.` related settings from the default `taskrc` [template](https://github.com/GothenburgBitFactory/taskwarrior/blob/0ff78447322a10e24c18b4ce2950745cf17962a1/src/Context.cpp#L267) to a new `default.theme` file. Use `include default.theme` to apply these color settings, like [here](https://github.com/GothenburgBitFactory/taskwarrior/blob/0ff78447322a10e24c18b4ce2950745cf17962a1/src/Context.cpp#L1186).

linsaraj:
Removed all "color." related settings from "Context.cpp" and relocated them into a newly created file called "default.theme" within the same directory as the "no-color.theme" file. 
Included "default.theme" file in "Context.cpp" using "#include default.theme".


src/Context.cpp
```diff
@@ -1195,6 +1123,7 @@ void Context::createDefaultConfig() {
              << "#include solarized-dark-256.theme\n"
              << "#include solarized-light-256.theme\n"
              << "#include no-color.theme\n"
+             << "#include default.theme\n"
```
djmitche: This should be un-commented:
```suggestion
             << "include default.theme\n"
```

doc/rc/default.theme
```diff
@@ -0,0 +1,101 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# https://www.opensource.org/licenses/mit-license.php
+#
+###############################################################################
+
+rule.precedence.color=deleted,completed,active,keyword.,tag.,project.,overdue,scheduled,due.today,due,blocked,blocking,recurring,tagged,uda.
+
+This contains all the default colors.
```
djmitche: This isn't a valid config setting, so it needs to be a comment:
```suggestion
# This contains all the default colors.
```

djmitche (changes requested):
Nice work! I can confirm that the `default.theme` gets installed automatically, without having to be mentioned in `doc/CMakeLists.txt`.

However, this causes a few tests to fail, and I think that's because the removed lines in `src/Context.cpp` are actually the default config (as in, what Taskwarrior uses if no config file is found or a setting doesn't appear there). So when the tests run, there are no color settings at all.

I tried fixing this by putting `include default.theme` in `configurationDefaults` but not surprisingly, this didn't work -- it didn't find that file, since tests run from from the source code and not the installed application.

@d2718nis can correct me if I'm wrong, but I think the idea is to put the theme settings in one place (`doc/rc`) so we can't forget to update them when adding a new color setting. If that's the case, then we just need to find a way to _also_ include that information in the default settings built into the `task` binary. https://stackoverflow.com/questions/410980/include-a-text-file-in-a-c-program-as-a-char/47801116 has some ideas of how to do that, but they look pretty complicated and involve messing with CMake.

@linsaraj and @d2718nis what do you think?

linsaraj:
Screenshot 2025-02-07 at 6 58 43 PM

Perhaps this would work? It's from the stack overflow you linked and doesn't seem to mess with the CMake! It just creates a pointer to the actual file, making it into a string and part of the source code!

djmitche:
I think the issue with that solution is it requires that the file itself have `R"(` and `)"` in the file, which would make it unreadable as an `include` from `.taskrc`.

linsaraj:
Ah, I see. Maybe we can create a function that reads a text file we want (like the "default.theme" file) and outputs it into a string so that we can call that function to include it in the source code? That way, we don't have to edit or add R"()" to the files!

djmitche:
Maybe! Let's try that..

Yedfree (approved):

djmitche:
Closing as stale, but I'm happy to re-open on request.

PR #3795, opened 2025-03-01
Remove unused includes

tobiolo:


djmitche (approved):
Was this done with a tool of some sort?

tobiolo:
LSP-clangd in Sublime Text gave me the indication that they were not used anymore. The rest was change in each file and test if it stills build. After that, commit each change.

djmitche:
Thanks! It would be great to have some standard for when/where to make includes, and a tool to enforce that. I work on Chromium, and there's an "[Include What You Use](https://chromium.googlesource.com/external/github.com/include-what-you-use/include-what-you-use/+/refs/tags/upstream/0.13/README.md)" (IWYU), but it seems pretty half-baked to this day.

I guess this improves compile times -- are there other advantages?

tobiolo:
I currently see the advantage of compile time improvement, tidied code. For the rewrite of the C++ parts to rust code by AI (at least for draft) I think it is also helpful as it avoids confusion.

djmitche:
OK, well, I am happy to merge this as it clearly works and has no cost.

FWIW there's not much appetite to rewrite more of Taskwarrior in Rust without a compelling reason to do so. And if it were done, it would need to be done carefully, not by AI.

A few good tasks for cleaning up the C++:
 - Ensure that definitions in x.cc are declared in x.h. There are especially some odd cases with `main.h`.
 - Use forward declarations to avoid heavy dependency on Task.h, Context.h, and TDB.h. I think this is a large part of the reason for recompiling all of src/commands on just about any change.

PR #3796, opened 2025-03-01
Declare in corresponding header files and dissolve main.h

tobiolo:
Apply include-what-you-use


src/Operation.h
```diff
@@ -24,8 +25,8 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-#ifndef INCLUDED_OPERATIOn
-#define INCLUDED_OPERATIOn
```
djmitche, line 28: Haha, that was a djmitche typo :)

src/feedback.cpp
```diff
@@ -222,8 +222,8 @@ std::string onProjectChange(Task& task, bool scope /* = true */) {
 std::string onProjectChange(Task& task1, Task& task2) {
   if (task1.get("project") == task2.get("project")) return onProjectChange(task1, false);
 
-  std::string messages1 = onProjectChange(task1);
-  std::string messages2 = onProjectChange(task2);
+  std::string messages1 = onProjectChange(task1, true);
+  std::string messages2 = onProjectChange(task2, true);
```
djmitche: The second argument defaults to `true` - any reason to pass it explicitly here?

djmitche (changes requested):
Looks good -- thanks!

Static functions are local to the .cc file and should not be in the header. Similar for static variables (gsColor, etc.).

src/feedback.cpp
tobiolo: Thanks for your review! No reason here, so I corrected it.

djmitche:
Please flag me for review again when it's ready for another look

tobiolo:
> Looks good -- thanks!
> 
> Static functions are local to the .cc file and should not be in the header. Similar for static variables (gsColor, etc.).

Thanks for the review! I am not so familiar and got confused with `static` in context of classes and just functions. This is fun, thanks! I incorporated your review comments.

djmitche (approved):
Nice work -- thanks!

tobiolo:
You are welcome!

PR #3797, opened 2025-03-02
Remove header files out of CMakeLists

tobiolo:
It does not make any difference.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,20 +6,20 @@ include_directories (${CMAKE_SOURCE_DIR}
                      ${CMAKE_SOURCE_DIR}/src/libshared/src
                      ${TASK_INCLUDE_DIRS})
 
-add_library (task STATIC CLI2.cpp CLI2.h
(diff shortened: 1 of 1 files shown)


djmitche:
I can't find any indication in the CMake docs as to whether header files should or should not be included in add_library's `<sources>`. Do you have links to why this makes no difference?

tobiolo:
As far as I can tell from a little Google search it helps IDEs using CMake to show the header files more prominently. But for the compilation itself header files indicated in add_library will be ignored. And I think they can find the header files via include_directories. No need for redundancy here.

djmitche:
Got it -- Felix, what do you think?

felixschurk:
https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#static-libraries

I personally also did not knew this, but in the documentation they also have not included them. So sounds good.

Always nice to learn something new :)

felixschurk (approved):

djmitche:
Thanks!

PR #3802, opened 2025-03-05
Fix errors finding Rust toolchain

djmitche:


djmitche:
OK, I can reproduce this locally now, with
```
docker compose build test-ubuntu2004
```
I suspect something has changed with `rustup`?

djmitche:
Ah, `rustup` output now includes
```
(active, default) /root/.rustup/toolchains/stable-x86_64-unknown-linux-gnu
```
and that `(active, default)` is obviously not part of the path!

djmitche:
The fix was basically, update corrosion :)

felixschurk (approved):

PR #3803, opened 2025-03-06
Make updates after releasing

djmitche:
Right after making a release is a good time to update dependencies -- that gives adequate time to test those updates and uncover any issues.


felixschurk:
Are the failing tests related to #3802?

djmitche:
> Are the failing tests related to #3802?

Yes.

PR #3804, opened 2025-03-06
Update task.1.in

yongli-abc:
Clearly this is an error. The example command should use foo.


djmitche (approved):
Good spot -- thank you!

PR #3811, opened 2025-03-10
Release 3.4.0

djmitche:
I thought about making this 3.3.1, since it's not a lot of user-facing change. But I think I'd prefer to leave the last digit to bugfix-style patches, and this contains more than bugfixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ enable_testing()
 set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 project (task
-  VERSION 3.3.0
+  VERSION 3.4.0
   DESCRIPTION "Taskwarrior - a command-line TODO list manager"
   HOMEPAGE_URL https://taskwarrior.org/)
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,10 +1,31 @@
 ------ current release ---------------------------
 
+3.4.0 -
+
+- Where possible, the task DB is now opened in read-only mode, which improves
+  performance. This is the case for reports (task lists) only when the `gc`
+  config is false.
+- An updated version of corrosion fixes build errors trying to find the Rust
+  toolchain.
+
+Thanks to the following people for contributions to this release:
+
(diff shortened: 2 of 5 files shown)


felixschurk:
Should we update the `taskrc` man page?
Currently it still has
```
       locking=1
              Determines  whether  to  use  file  locking  when accessing the pending.data and com‐
              pleted.data files. Defaults to "1". Solaris users who store the data files on an  NFS
              mount may need to set locking to "0". Note that there is danger in setting this value
              to  "0" - another program (or another instance of task) may write to the task.pending
              file at the same time.

       gc=1   Can be used to temporarily suspend garbage collection (gc), so that  task  IDs  don't
              change.  Note  that  this should be used in the form of a command line override (task
              rc.gc=0 ...), and not permanently used in the .taskrc file, as this significantly af‐
              fects performance in the long term.
```

Where `locking=1` I think is obsolete and `gc=1` is now a bit contradicting with the release notes if I get it right.

felixschurk:
Should we update the `taskrc` man page?

Currently it still states
```
       locking=1
              Determines  whether  to  use  file  locking  when accessing the pending.data and com‐
              pleted.data files. Defaults to "1". Solaris users who store the data files on an  NFS
              mount may need to set locking to "0". Note that there is danger in setting this value
              to  "0" - another program (or another instance of task) may write to the task.pending
              file at the same time.

       gc=1   Can be used to temporarily suspend garbage collection (gc), so that  task  IDs  don't
              change.  Note  that  this should be used in the form of a command line override (task
              rc.gc=0 ...), and not permanently used in the .taskrc file, as this significantly af‐
              fects performance in the long term.
```
Where `locking=1`, I think is now obsolete and `gc=1` is a bit contradicting with the release notes or not?

djmitche:
It looks like `locking` has already been removed:
https://github.com/GothenburgBitFactory/taskwarrior/blob/f73b42d23f6318d8e14d2c7113fcda0865964e4c/doc/man/taskrc.5.in#L208-L220

But you're right, `gc` can be updated.

PR #3812, opened 2025-03-10
Allow dur/dur division

Ram-Z:
It is perfectly valid to divide two durations, it is a ratio with no unit.

### DIFF
--- a/src/Variant.cpp
+++ b/src/Variant.cpp
@@ -75,7 +75,6 @@
 #define STRING_VARIANT_DIV_DUR_BOOL "Cannot divide duration by Boolean"
 #define STRING_VARIANT_DIV_DUR_STR "Cannot divide durations by strings"
 #define STRING_VARIANT_DIV_DUR_DATE "Cannot divide durations by dates"
-#define STRING_VARIANT_DIV_DUR_DUR "Cannot divide durations by durations"
 #define STRING_VARIANT_MOD_BOOL "Cannot modulo Booleans"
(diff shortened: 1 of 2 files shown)


Ram-Z:
One other missing duration operations I found is `dur%dur`.

There are also some duration/date operations that are implemented that don't make much sense. I'm reluctant to remove those as that would be a breaking change.

Here's a couple I found: 
* `int/dur`, `real/dur`, `sqrt(dur)`, `sqrt(date)`,... 
* `<=>` operators are very fluid: allowing `date<dur`, `bool<date`, ...
* `+-` operators too: `bool+dur->dur`, `date-int->date`,...
  * AFAICS, all dates/duration have a resolution of seconds, so these casts will add/sub seconds

djmitche:
Is this operation usable from the command line?

Ram-Z:
Yes.

```
./build-Linux-x86_64-Debug/src/task calc "1w / 2d"
3.5
./build-Linux-x86_64-Debug/src/task calc "(1w / 2d) * 3d"
P10DT12H
```

djmitche:
Oh, I completely forgot about `task calc`. Thanks!

Ram-Z:
@djmitche Any thoughts on the other operations? https://github.com/GothenburgBitFactory/taskwarrior/pull/3812#issuecomment-2712170861

Is there a deprecation path? Should these be kept as grand-fathered "features"?

djmitche:
I think there's no harm in leaving nonsense operations. A PR adding % would be accepted!

PR #3817, opened 2025-03-13
Fix suppressing news nag after reading the news

kietavainen:
The news nag suppression regressed again in 5c67d22. That commit intended to remove the sponsorship outro from the news, but also removed the bookkeeping that marks the news as read. This commit reverts that part back to its previous state.


djmitche (approved):
Ugh, how am I so bad at this very particular thing?

PR #3824, opened 2025-03-18
Initial changes for Windows compatibility

Foadsf:
# Windows Compatibility: Initial Changes

This PR introduces initial changes to make Taskwarrior build on Windows with MSVC:

## Changes made:
- Modified CXXSniffer.cmake to correctly detect C++17 support on MSVC
- Modified CMakeLists.txt to use Windows native UUID APIs instead of libuuid
- Made initial changes to header files for cross-platform compatibility

## Next steps (not part of this PR):
- Create platform-specific implementations for POSIX functionality
- Add Windows-specific equivalents for Unix headers
- Address remaining platform-specific code patterns

This is the first step toward making Taskwarrior truly cross-platform and compiler-agnostic. These changes do not affect functionality on Unix/Linux platforms.


src/recur.h
```diff
@@ -36,11 +36,20 @@
...
 #include <Lexer.h>
 #include <format.h>
+
+// Platform-specific includes
+#ifdef _WIN32
+#include <windows.h>
+// Windows-specific user info implementation will be needed
+// Create a pwd.h equivalent for Windows
+#else
 #include <pwd.h>
```
djmitche, line 46: This, and unistd.h, appears to be unused and could just be removed.

src/util.h
```diff
@@ -30,16 +30,25 @@
...
-#if defined(FREEBSD) || defined(OPENBSD)
+
+// Platform-specific UUID handling
+#ifdef _WIN32
+#include <rpc.h>
+#include <windows.h>
+// Define uuid_t for Windows compatibility
+typedef UUID uuid_t;
+#ifndef uuid_unparse_lower
+void uuid_unparse_lower(uuid_t uu, char* out);
```
djmitche, line 43: This shouldn't be necessary if HAVE_UUID_UNPARSE_LOWER is unset correctly.

djmitche (commented):
The libshared update is to a revision that doesn't exist, so I'm not sure what the idea is there?

This looks good so far. We should definitely add a Windows GitHub action to check we don't break this support.

djmitche:
@Foadsf will you be able to continue this work?

djmitche:
Closing the PR, but we can always re-open or post a new PR to continue this.

PR #3825, opened 2025-03-18
Update dependabot.yml

### LINKED ISSUE #3813: Use dependabot to update submodules
We should keep both corrosion and libshared (both imported as submodules) up to date.

It seems like [dependabot supports this](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-).

felixschurk:
Closes #3813.

I have no clue if this is already it, so lets see.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,11 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Let's see!

PR #3827, opened 2025-03-18
Add uuid UDA type

Ram-Z:
Mainly so that UDAs that refer to another task can be formated as "short".

### DIFF
--- a/test/columns.test.py
+++ b/test/columns.test.py
@@ -109,29 +109,26 @@ class TestUUIDFormats(TestCase):
     def setUpClass(cls):
         """Executed once before any test in the class"""
         cls.t = Task()
-        cls.t.config("report.xxx.columns", "id,uuid")
+        cls.t.config("report.xxx.columns", "uuid")
         cls.t.config("verbose", "nothing")
 
         cls.t("add zero")
         code, out, err = cls.t("_get 1.uuid")
         cls.uuid = out.strip()
 
-    def setUp(self):
-        """Executed before each test in the class"""
-
     def test_uuid_long(self):
         """Verify formatting of 'uuid.long' column"""
-        code, out, err = self.t("xxx rc.report.xxx.columns:id,uuid.long")
-        self.assertIn(self.uuid, out)
+        code, out, err = self.t("xxx rc.report.xxx.columns:uuid.long")
+        self.assertEqual(self.uuid, out.strip())
 
     def test_uuid_short(self):
         """Verify formatting of 'uuid.short' column"""
-        code, out, err = self.t("xxx rc.report.xxx.columns:id,uuid.short")
-        self.assertIn(self.uuid[:7], out)
+        code, out, err = self.t("xxx rc.report.xxx.columns:uuid.short")
+        self.assertEqual(self.uuid[:8], out.strip())
 
     def test_uuid_format_unrecognized(self):
         """Verify uuid.donkey formatting fails"""
-        code, out, err = self.t.runError("xxx rc.report.xxx.columns:id,uuid.donkey")
+        code, out, err = self.t.runError("xxx rc.report.xxx.columns:uuid.donkey")
         self.assertEqual(err, "Unrecognized column format 'uuid.donkey'\n")
 
 
--- a/test/columns.test.py
+++ b/test/columns.test.py
@@ -482,6 +482,45 @@ start       active*           ✓
         """
 
 
+class TestUDAUUIDFormats(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """Executed once before any test in the class"""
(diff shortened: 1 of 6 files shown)


Ram-Z:
Missing updates to man page.

djmitche (commented):
This is cool, thanks!

> Missing updates to man page.

Yep, and some tests for this new functionality.

Ram-Z:
Have added a test, ~~but when run locally it didn't seem to pick up the modified executable, and failed. Maybe I'm missing something for running the tests.~~ OK... was being silly and was only rebuilding the `test_runner` rather than all executables.

test/columns.test.py
```diff
@@ -482,6 +482,41 @@ def test_uda_format_unrecognized(self):
...
+        cls.t.config("report.xxx.columns", "id,uda_uuid")
+
+        cls.t("add zero")
+        code, out, err = cls.t("_get 1.uuid")
+        cls.t("add uda_uuid:{} one".format(out))
+        code, out, err = cls.t("_get 2.uda_uuid")
+        cls.uda_uuid = out.strip()
+
+    def setUp(self):
+        """Executed before each test in the class"""
```
djmitche: This method can be omitted if not used

test/columns.test.py
```diff
@@ -482,6 +482,41 @@ def test_uda_format_unrecognized(self):
...
+
+    def test_uda_uuid_long(self):
+        """Verify formatting of 'uda_uuid.long' column"""
+        code, out, err = self.t("xxx rc.report.xxx.columns:id,uda_uuid.long")
+        self.assertIn(self.uda_uuid, out)
+
+    def test_uda_uuid_short(self):
+        """Verify formatting of 'uda_uuid.short' column"""
+        code, out, err = self.t("xxx rc.report.xxx.columns:id,uda_uuid.short")
+        self.assertIn(self.uda_uuid[:7], out)
```
djmitche: This would also match the long format. Maybe assert `self.uda_uuid[:7] + '\n'` is in output?

djmitche (commented):
Looks good! Could you also add some tests for an invalid uuid in the task DB? I suspect other UDA types will crash in this sort of situation, but let's make sure the uuid type is robust.

Ram-Z:
Addressed both comments and aligned the TestUUIDFormats class with suggestions since I simply copied that as base for my test.

djmitche (commented):
I think this bit was missed?

> Could you also add some tests for an invalid uuid in the task DB

Ram-Z:
It was indeed. On it.

Ram-Z:
Although not entirely sure where to start, where are these task DB tests?

I'm adding a test for `task add uda_uuid:invalid`, but I'm not sure that is what you mean.

djmitche:
I think one way to do this would be to define the UDA as type=string, add a value, and then reconfigure it with type uuid.

Ram-Z:
Sorry, I need to stop for today and will be AFK for the next 2 weeks. Feel free to take over if you don't want it sitting here for that time. Or I'll pick it up again when I'm back.

Ram-Z:
> I think one way to do this would be to define the UDA as type=string, add a value, and then reconfigure it with type uuid.

What exactly do you want to happen in that case? The current behaviour is to simply leave the string value as is. The only difference is that this string can now be formatted as 'short' (only showing the 8 first char).

If that uda with an invalid uuid were then to be used as an uuid in a hook or something, there would be an error. Which allows the user to rectify the content then.

I've added a test that shown the above behaviour.

test/columns.test.py
```diff
@@ -482,6 +479,70 @@ def test_uda_format_unrecognized(self):
...
+
+        cls.t("add zero")
+        code, out, err = cls.t("_get 1.uuid")
+        cls.t("add uda_uuid:{} one".format(out.strip()))
+        code, out, err = cls.t("_get 2.uda_uuid")
+        cls.uda_uuid = out.strip()
+
+    def test_uda_uuid_invalid_fails(self):
+        """Verify adding invalid uuid fails"""
+        code, out, err = self.t.runError("add uda_uuid:shrek three")
```
djmitche, line 500: 🤣

djmitche (approved):
Nice work - thanks!!

djmitche:
> > I think one way to do this would be to define the UDA as type=string, add a value, and then reconfigure it with type uuid.
> 
> What exactly do you want to happen in that case? The current behaviour is to simply leave the string value as is. The only difference is that this string can now be formatted as 'short' (only showing the 8 first char).

Not crashing is the desired behavior :)

We've had a long series of bugs where if some property has an unexpected value every `task` command ends up failing with a vague error like "invalid date" and no other output. I'd like to avoid introducing more such errors!

Issue #3826, opened 2025-03-18
Durations/periods are not handled consistenly


Ram-Z:
A duration or period may be ambiguous/vague, f.ex. 1 month may be 28/29/30/31 days. However when doing calculations on dates, taskwarrior will "normalise" a duration into seconds then add that to the date.

This can easily be verified with `task calc`, but it also applies to `task add due:...`.

```
task calc 2025-03-18+1mo
2025-04-17T01:00:00
```

This should be `2025-04-18T00:00:00`. I didn't plan this, but this also included a DST change, hence the extra unexpected hour difference in addition to the day difference.

Taskwarrior seems to handle this correctly in at least one place, recurrence. It is also well documented in [docs/durations](https://taskwarrior.org/docs/durations/)[^1]. However it doesn't behave like that everywhere.

[^1]: although it misses that 1 day is sometimes imprecise too due to DST.

In recurrence, it simply stores the period as a string and reparses it every time a template is created, taking into account the context of the date (`src/recur.cpp`). For all other locations it seems to convert to seconds first using some "arbitrary" value (i.e. 30 days in a month), and stores it as a single `time_t` field (`src/libshared/src/Duration`).

Ram-Z:
One solution would be to always store a duration as separate fields, i.e. `duration { Y, M, D, h, m, s }` and never normalise them. This properly differentiates 1M from 30D, 2M from 60D, etc. And should be able to be added/subtracted to dates easily by first converting the date to a local datetime in the same format. This is pretty much what `recur.cpp` does.

However, some operations between durations become complicated.
```
2M>=1M  // always true
2M>=58D // always true (28+30)
2M>=59D // also always true Feb is surrounded by two months of 31
2M>=60D // true or false depending on what months
2M>=61D // true or false
2M>=62D // true or false
2M>=63D // always false
1Y/2!=6M // twice a year, not the same as 6M
```

djmitche:
Date and time handling is an endlessly complex task, but Taskwarrior only needs to accomplish what is relevant to personal task-tracking. Getting recurrence "intuitively" correct is important, but rigorously precise comparison of durations is not.

There's a general sense that recurrence needs an overhaul, and I think that's one of the only places that we store a duration. So, perhaps part of the proposal for that project could be more precise recurrence intervals. Maybe they aren't durations at all -- for example, maybe things like "first Tuesday of the month" or "every Tuesday and Friday" are more useful for task-tracking.

[...]

PR #3829, opened 2025-03-24
Replace inacurate comment

Ram-Z:
As discovered in https://github.com/GothenburgBitFactory/taskwarrior/issues/3826#issuecomment-2746468031, the duration UDA is not stored in raw form like the recur column is.


djmitche (approved):
Thanks, nice spot!

PR #3845, opened 2025-04-10
Add note about uninstallation

awilkins:
If you remove the source folder as per the 6-step instructions for installation, you will remove the installation manifest and be unable to use it to remove Taskwarrior later.


djmitche (approved):
Thanks!

PR #3847, opened 2025-04-14
Redact HTTP credentials from "Syncing…" message (#3846)

### LINKED ISSUE #3846: Don't show HTTP credentials from `sync.server.url` when syncing
Suppose I have Basic Authentication set up in front of my `taskchampion-sync-server`, and so I set `sync.server.url` to `https://USER:PASS@example.com/`. Those credentials - or at least, the password - should not be revealed on the command line when syncing.

In other words, I currently see this:
```
$ task sync
Syncing with sync server at https://USER:PASS@example.com/
```
…and I think I should instead see something like this:
```
$ task sync
Syncing with sync server at https://example.com/
```
…or maybe one of:
```
Syncing with sync server at https://USER@example.com/
```
```
Syncing with sync server at https://USER:****@example.com/
```

nickgrim:
This PR updates the message shown when running `task sync` to redact any HTTP credentials within `sync.server.url` - the password is replaced in the message with `****`. Closes #3846.

I couldn't see any syncing-related unit tests to bundle the test with, so I created a new test file.


test/sync_message.test.py
nickgrim, line 1: This test passes locally but fails in CI, presumably because it's _actually_ connecting to `example.com`.

My Python is Not Strong, and I couldn't see any tests mocking an HTTP sync, and didn't especially want to pull in a new dependency to do that. In the end I figured the best option might be to leave the test file but don't run it (as per `test/CMakeLists.txt` :point_up:), but feel free to disagree.

src/commands/CmdSync.cpp
```diff
@@ -78,6 +79,10 @@ int CmdSync::execute(std::string& output) {
     out << "sync.server.origin is deprecated. Use sync.server.url instead.\n";
   }
 
+  // redact credentials from `server_url`, if present
+  std::regex remove_creds_regex("(https?://.+):(.+)@(.+)");
```
djmitche:
```suggestion
  std::regex remove_creds_regex("^(https?://.+):(.+)@(.+)");
```

test/sync_message.test.py
djmitche: I wonder if you could make the test pass by just not expecting the sync to succeed? So it fails, but does output the right message?

djmitche (commented):
This looks good! If you can get the test working, that's great -- but if not, I think it would be better to just omit it. This is "leaf" functionality that nothing else really depends on, so even if we accidentally regressed it, we could just fix it again.

src/commands/CmdSync.cpp
nickgrim: Ah, yes, of course.  :D

test/sync_message.test.py
nickgrim: I'm unsure how I'd go about doing that, tbh.

If I deliberately fail a sync locally (e.g. with incorrect credentials) I get a different error message printed to ~stdout~ stderr and this particular message is never printed. In CI the issue seems to be caused by a timeout that leads to a `SIGABRT` ([here](https://github.com/GothenburgBitFactory/taskwarrior/pull/3847/checks?sha=12a5e5185cca1b5e368014af981d09da421c6af4#step:6:299)) rather than a well-handled sync-failure, iyswim.

nickgrim: …but, honestly, I'm happy to just remove the test if you are.  :D

djmitche: Interesting. I suppose the "syncing.." message isn't flushed out of the pipeline before it crashes. In that case, removing the test is fine!

djmitche (approved):

djmitche:
Woo, thank you!

PR #3849, opened 2025-04-16
change implementation from has

### LINKED ISSUE #3848: task diag hangs up due to memory high memory usage
When running `task diag` on my own taskdatabase (around 8177 tasks are there in) the execution is killed by the operation system due to high memory usage. (16GB of ram available)
Playing a bit around with it, I do think that it is due to the lines highlighted below, the checking for the recurrence.

https://github.com/GothenburgBitFactory/taskwarrior/blob/15bb71764e7820a9df4ad5cd5b13963bc90fd6ad/src/commands/CmdDiagnostics.cpp#L240-L268

I also tried it with the "performance" database, there is no crashing - however there are also no dependencies available. My assumption is that the traversing/searching in the SQL database is somehow not optimal and thus tries to load everything into the memory - but that is only a guess.

felixschurk:
Change from `get(uuid, task)` to `get_task_data` to always have a full uuid match and to improve read performance for the diagnostics command.

Closes #3848.

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -354,8 +354,7 @@ bool TDB2::get(const std::string& uuid, Task& task) {
 ////////////////////////////////////////////////////////////////////////////////
 // Locate task by UUID, wherever it is.
 bool TDB2::has(const std::string& uuid) {
-  Task task;
-  return get(uuid, task);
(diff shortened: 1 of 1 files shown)


djmitche (approved):
That is a simple improvement for a big difference in performance!

felixschurk:
Oh yes, from crashing due to memory issue to running in a few micro seconds.

PR #3859, opened 2025-04-30
Do not decode a non-JSON value

### LINKED ISSUE #3858: Trailing `\` in annotation causes spurious NUL character
This is the only output I get when I run `task` or `task <tasknumber>`. It takes several seconds to get this output, so it looks like it runs into a timeout.

However, `task proj:<projectname>` works.

I am using version 3.4.1 on WSL Arch Linux.

Are there any logs, or is there anything else, I can do to further analyse, what could be the problem?

djmitche:
None of the other task modifications (modify, prepend, append) treat the input as JSON, so this one shouldn't either. This works around https://github.com/GothenburgBitFactory/libshared/issues/95. Fixes #3858.

### DIFF
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -928,7 +928,7 @@ void Task::addAnnotation(const std::string& description) {
     ++now;
   } while (has(key));
 
-  data[key] = json::decode(description);
+  data[key] = description;
   ++annotation_count;
   recalc_urgency = true;
 }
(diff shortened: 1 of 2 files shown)


djmitche:
(I looked at the blame for this line and it's from 10 years ago!)

felixschurk (approved):
Good catch :)

PR #3867, opened 2025-05-05
use cached urgency() for json exports

### LINKED ISSUE #3866: exported urgency with urgency.inherit=1 is calculated inefficently
If I interpret the code correctly, TaskWarrior is currently "recalculating" the urgencies on an export by using `urgency_c()`, ignoring the cache implemented by `urgency()`.
I have found no comment, neither in the code, nor in the introducing commit 0918a7c79adc580b5f9a84b2758f44f3622874f6 as to about why the cache is ignored here.
Because I see no reason for that, I decided to open this issue.

To be precise, I mean the call in following code snippet:
https://github.com/GothenburgBitFactory/taskwarrior/blob/36e5f8895d8508374abb4f0df4ef8baf3926f2c9/src/Task.cpp#L895-L898

I am unsure yet whether you could just change that to `urgency()` without introducing any issues (esp. because I am neither into C++ nor into this code base), but I will attempt to test that on my systems for the time being.

Zocker1999NET:
speeding up the export of tasks for cases with urgency.inherit=1

fixes #3866 

untested, PR submitted for running GitHub Actions as requested in #3866

### DIFF
--- a/src/Hooks.cpp
+++ b/src/Hooks.cpp
@@ -275,7 +275,7 @@ void Hooks::onAdd(Task& task) const {
 // - all emitted non-JSON lines are considered feedback or error messages
 //   depending on the status code.
 //
-void Hooks::onModify(const Task& before, Task& after) const {
+void Hooks::onModify(Task& before, Task& after) const {
   if (!_enabled) return;
 
   Timer timer;
--- a/src/Hooks.h
+++ b/src/Hooks.h
@@ -40,7 +40,7 @@ class Hooks {
   void onLaunch() const;
   void onExit() const;
   void onAdd(Task&) const;
-  void onModify(const Task&, Task&) const;
+  void onModify(Task&, Task&) const;
   std::vector<std::string> list() const;
 
  private:
--- a/src/Task.cpp
+++ b/src/Task.cpp
@@ -777,7 +777,7 @@ void Task::parseLegacy(const std::string& line) {
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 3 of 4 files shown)


djmitche:
```
/home/runner/work/taskwarrior/taskwarrior/task-3.4.1/src/Task.cpp:897:56: error: passing ‘const Task’ as ‘this’ argument discards qualifiers [-fpermissive]
  897 |   if (decorate) out << ',' << "\"urgency\":" << urgency();
      |                                                 ~~~~~~~^~
```
So it looks like the issue is that the export method is const, but `urgency` isn't (because it modifies the cache).

djmitche:
I removed the const-ness from the call chain, so let's see how this works..

djmitche:
Is this noticeably faster?

Zocker1999NET:
okay, got it finally built (take a while to learn that taskwarrior3 now requires submodules to be fetched, explaining why just updating the tag & hashes in its [package.nix from nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ta/taskwarrior3/package.nix) was not working)

running `time task export >/dev/null`:
- with v3.1.0 (latest nixpkgs stable) took 3m21.014s (what I was using until now)
- with the current develop branch took 0m8.691s
- with v3.4.1 (latest nixpkgs unstable) took 0m8.860s

I think this PR made some speedup, albeit being not that noticable ...
more important to me, through this I discovered that taskwarrior became way more efficient since v3.1.0 ... anyway thanks for your fast help & merging :sweat_smile:

PR #3874, opened 2025-05-09
Make libuuid not mandatory as internal implementation is available

tobiolo:


djmitche:
Could we just get rid of use of libuuid entirely, then? What's the advantage to using it when it's available?

tobiolo:
Hello @djmitche 
I doublechecked and unfortunately the implementation still relies partly on `libuuid` as library function `uuid_generate` is used.

Issue #3899, opened 2025-06-22
the Taskwarrior hook examples should use double quotes when emitting JSON


bughunter2:
Here's a partial snippet from the `on-modify` example hook provided by Taskwarrior itself.

```
...
...
# Output:
# - JSON, modified or unmodified.
# - Optional feedback/error.
echo $modified_task
echo 'on-modify'
...
...
```

Instead of `echo $modified_task`, it should use **`echo "$modified_task"`**

**Why:** because otherwise all kinds of undesirable things may occur, like losing multiple spaces (they would be compressed to a single space by the shell). Even **shell expansion** can occur if the task contains one or more * characters (asterisks). For example, a task with `hello * world` as its description could be expanded
to `hello <all_files_and_directories_in_the_current_directory> world`.

Of course the hook scripts provided by Taskwarrior are just examples, but this is an easy and small improvement, providing a solid foundation upon which to build hooks without such errors.

The following Taskwarrior hook examples are affected (output format: filename:line_number:content):

**Original:**
```
on-add:13:echo $new_task
on-add.the:5:if (echo $new_task | grep -qE '[tT]eh');
on-add.the:7:  new_task=$(echo $new_task | sed -r 's/([tT])eh/\1he/g')
on-add.the:11:echo $new_task
on-exit.shadow-file:30:  echo Could not create $SHADOW_FILE
on-exit.shadow-file:34:echo Shadow file $SHADOW_FILE updated.
on-modify:15:echo $modified_task
```

**Fixed:**
```
on-add:13:echo "$new_task"
on-add.the:5:if (echo "$new_task" | grep -qE '[tT]eh');
on-add.the:7:  new_task=$(echo "$new_task" | sed -r 's/([tT])eh/\1he/g')
on-add.the:11:echo "$new_task"
on-exit.shadow-file:30:  echo "Could not create $SHADOW_FILE"
on-exit.shadow-file:34:echo "Shadow file $SHADOW_FILE updated."
on-modify:15:echo "$modified_task"
```

djmitche:
Thanks! Do you want to make a PR to address this?

bughunter2:
Although I never clone repos or make PRs, I can provide you with a `diff` that can be applied using the `patch` command.

Feel free to credit me (not required) if that's something you normally do for PRs.

[...]

PR #3909, opened 2025-07-01
Fix and test handling of backslashes in hooks

### LINKED ISSUE #3900: escape sequences handled incorrectly when Taskwarrior sends/receives tasks to/from hook scripts
If a task has a description such as `hello \t world`, then the escape sequence may get lost (or converted) when the task is passed to a hook script and received from the hook script. We created the task using `task add` and then `task edit` to insert the escape sequence (which was preserved properly). Hook scripts were not used when creating the task, to ensure the problem wasn't caused by hook scripts.

The problem happens even if the hook script simply passes the input as-is (unmodified) back to Taskwarrior.

For example, if we modify the example hook `on-modify` script to make it parse the input as-is (using `read -r`) and pass it back unmodified, using `echo "$modified_task"`, then the above problem still occurs.

Here's a hook script that shows us what happens:

```sh
#!/bin/sh

# Input (using read -r to ensure escape sequences are preserved):
read -r original_task
read -r modified_task

# Output (using double quotes to prevent shell expansion):
echo "$modified_task"
echo "DEBUGGING: Original task: $original_task" >&2

exit 0
```

When we use `task export`, we can see that its output differs from the output produced by the `echo "$modified_task"` line in the hook. Note that the `task export` output properly escapes the `\t` as `\\t`. But that's not what we receive in the hook script when Taskwarrior calls it (there, we receive just `\t`, even though we used `read -r` to read the input).

**This seems to indicate that Taskwarrior itself is the source of this behavior.**

Example outputs:

```
$ task export
[
{"id":1,"description":"hello \\t world","entry":"20250622T171235Z","modified":"20250622T171241Z","status":"pending","uuid":"813a6b02-e860-41bc-be07-6881f53dc9d8","urgency":0}
]
```

```
$ task edit 1
Launching 'editor "task.813a6b02.task"' now...
Editing complete.
Edits were detected.
DEBUGGING: Original task: {"description":"hello \t world","entry":"20250622T171235Z","modified":"20250622T171241Z","status":"pending","uuid":"813a6b02-e860-41bc-be07-6881f53dc9d8"}
Description modified.
```

It seems that `task export` does NOT have the escaping problem explained above.

The problem only occurs when a task is passed through a hook script, even if the task remains unmodified, and even if the hook script simply emits the same output that it received from Taskwarrior.

Even when we force the hook script to emit the **exact same task line** that **`task export`** shows, then the escape sequence is **lost again** when Taskwarrior **receives** (or interprets) the output generated by the hook script.

In other words, it seems Taskwarrior does not properly escape a task when (a) passing it to a hook script, and (b) receiving it from the hook script.

djmitche:
This adds a test for #3900, and as it turns out the issue isn't TaskWarrior, but the shell implementation of the hook.  ~~Even with the `read -r` and escaping the input to `echo` added in #3899, it still interprets the backslash escape. This is illustrated by converting the hook to a Python script (in the last commit), fixing the test.~~

As discussed in #3899, using `read -r` and `printf "%s"` avoids these issues. This PR updates all hooks (tests and examples) and adds a test which failed with the old shell syntax.

### DIFF
--- a/scripts/hooks/on-add
+++ b/scripts/hooks/on-add
@@ -10,7 +10,7 @@ read -r new_task
 # Output:
 # - JSON, modified or unmodified.
 # - Optional feedback/error.
-echo "$new_task"
+printf "%s\n" "$new_task"
 echo 'on-add'
 
 # Status:
--- a/scripts/hooks/on-add.the
+++ b/scripts/hooks/on-add.the
@@ -2,11 +2,11 @@
 
 read -r new_task
 
-if (echo "$new_task" | grep -qE '[tT]eh');
+if (printf "%s" "$new_task" | grep -qE '[tT]eh');
 then
-  new_task=$(echo "$new_task" | sed -r 's/([tT])eh/\1he/g')
+  new_task=$(printf "%s" "$new_task" | sed -r 's/([tT])eh/\1he/g')
   echo "Auto-corrected 'teh' --> 'the'"
 fi
 
-echo "$new_task"
+printf "%s" "$new_task"
 exit 0
--- a/scripts/hooks/on-modify
+++ b/scripts/hooks/on-modify
@@ -12,8 +12,8 @@ read -r modified_task
 # Output:
 # - JSON, modified or unmodified.
 # - Optional feedback/error.
-echo "$modified_task"
-echo 'on-modify'
+printf "%s" "$modified_task"
+printf "%s" 'on-modify'
 
 # Status:
 # - 0:     JSON accepted, non-JSON is feedback.
--- a/test/basetest/task.py
+++ b/test/basetest/task.py
@@ -8,7 +8,13 @@ import tempfile
 import unittest
 from .exceptions import CommandError
 from .hooks import Hooks
-from .utils import run_cmd_wait, run_cmd_wait_nofail, which, task_binary_location
+from .utils import (
+    run_cmd_wait,
+    run_cmd_wait_nofail,
+    which,
+    task_binary_location,
+    CMAKE_BINARY_DIR,
+)
 from .compat import STRING_TYPE
 
 
@@ -299,6 +305,22 @@ class Task(object):
         if faketime is not None:
             # Use advanced time format
             self._command = [cmd, "-f", faketime] + self._command
+
+    def make_tc_task(self, **props):
+        """Create a task directly in TaskChampion, bypassing TaskWarrior
+        entirely, and returning the UUID. The properties are not interpreted by
+        the shell.
+        """
+        # Generate the path to the `make_tc_task` binary, which is a dependency of the
+        # test runner.
+        make_tc_task = os.path.abspath(
+            os.path.join(CMAKE_BINARY_DIR, "test", "make_tc_task")
+        )
+        cmd = [make_tc_task, self.datadir]
+        for p, v in props.items():
+            cmd.append(f"{p}={v}")
+        _, out, _ = run_cmd_wait(cmd)
+        return out.strip()
 
 
 # vim: ai sts=4 et sw=4
--- a/test/hooks.on-modify.test.py
+++ b/test/hooks.on-modify.test.py
@@ -176,6 +176,34 @@ class TestHooksOnModify(TestCase):
         hook.assertTriggeredCount(1)
         hook.assertExitcode(0)
 
+    def test_onmodify_escaped_backslash(self):
+        """on-modify-accept - a well-behaved, successful, on-modify hook."""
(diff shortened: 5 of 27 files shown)


djmitche:
To clarify, I tested this with the changes from #3899 applied to the test hook, and it didn't fix the test. It turns out, this has to do with the difference between `bash` and `dash` -- apparently `read -r` isn't a standardized POSIX shell thing:
```
⸩ bash -c "echo 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
foo\tbar
⸩ dash -c "echo 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
foo     bar
```

We should leave #3900 open to talk about how to adjust the documentation to help avoid this issue.

bughunter2:
> To clarify, I tested this with the changes from #3899 applied to the test hook, and it didn't fix the test. It turns out, this has to do with the difference between `bash` and `dash` -- apparently `read -r` isn't a standardized POSIX shell thing:
> 
> ```
> ⸩ bash -c "echo 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
> foo\tbar
> ⸩ dash -c "echo 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
> foo     bar
> ```
> 
> We should leave #3900 open to talk about how to adjust the documentation to help avoid this issue.

@djmitche: Sorry, but I think the problem is with Taskwarrior. Not the shell. The situation is more nuanced. Here's why:

Try this instead:

```sh
$ bash -c "printf '%s' 'foo\tbar' | ( read -r foo; set | grep '^foo='; )"
foo='foo\tbar'
$ dash -c "printf '%s' 'foo\tbar' | ( read -r foo; set | grep '^foo='; )"
foo='foo\tbar'
```

The outputs are identical now.

Similarly, this also produces identical outputs:

```sh
$ bash -c "echo -e 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
foo     bar
$ dash -c "echo 'foo\tbar' | ( read -r foo; echo \"\${foo}\"; )"
foo     bar
```

The difference is in the shell's implementation of the `echo` command, NOT in the `read -r` command. Whereas `bash` supports `echo -e` (and disables `-e` it by default) to control escape interpretation, `dash` instead doesn't support the `echo -e` option and always interprets escape sequences.

In other words, the changes as I proposed in #3899 are simply a **part** of the **solution**, but the **rest** of the problem must be resolved within Taskwarrior itself, as discussed in #3900.

djmitche:
It seems like the bug in the shell hooks isn't `read -r` as I thought, but the `echo` -- in a non-bash shell, that `echo`, even with `"${modified_task}"` in quotes, interprets the `\t`. Perhaps the advice should be to not write hooks in shell, or to always use bash?

In any case, that is still a bug in the hook, and not in Taskwarrior.

/cc @bpeetz as this might be an issue we could avoid in hooks v2?

bughunter2:
@djmitche There's an easy solution, in that case. Observe the difference between the following commands and outputs:

Before (the problematic situation):

```sh
a="hello\tworld"
echo "$a"
```

In `bash`, that outputs: `hello\tworld`
In `dash`, it outputs: `hello        world` with an actual tab character in between

After (fixed):

```sh
a="hello\tworld"
printf '%s\n' "$a"
```

In both `bash` and `dash`, that outputs: `hello\tworld`

Hope this helps. Cheers.

Perhaps in bug #3900 I could have been a bit more explicit, because I had used `printf` while debugging things (before I made the report) without mentioning it explicitly, but it's what I meant when I said (quote):
> Even when we force the hook script to emit the exact same task line that task export shows, then the escape sequence is lost again when Taskwarrior receives (or interprets) the output generated by the hook script.

That's why I'm sure Taskwarrior is messing something up as well. It is not just the hook scripts, it's both the hook scripts and Taskwarrior that cause the behavior in #3900.

djmitche:
OK, so it seems like we should update the example hooks to use `read -r` and `printf` instead.

bughunter2:
@djmitche Agreed. I have provided a new patch in #3899.

Note that, contrary to what I wrote earlier, I am now unable to reproduce the issue where `\t` gets interpreted (provided that we use the new patch with the `printf` approach), so it looks like the hook scripts were all that needed to be fixed. Perhaps I misremembered my `printf` debugging, and perhaps I used `printf` with `%b` rather than `%s` (as we have just agreed upon, `%s` is correct, keep it that way).

Can you confirm this as well? From my end all looks well with the new patch I provided in #3899.

felixschurk (approved):

PR #3923, opened 2025-07-31
Update task-sync(5) to highlight AWS support

### LINKED ISSUE #3921: task-sync(5) says only GCP supported, when AWS is as well
Howdy!

I was reading `task-sync(5)` today and noticed this section on configuration:

> CONFIGURATION
>        Taskwarrior provides several options for synchronizing your tasks:
> 
>         - To a server specifically designed to handle Taskwarrior data.
>         + To a cloud storage provider. Currently only GCP is supported.
>         - To a local, on-disk file.

(From: 
https://github.com/GothenburgBitFactory/taskwarrior/blob/7a5d222f90326645b86153c8a3536788bec3ff76/doc/man/task-sync.5.in#L47 )

But that's not true!  AWS/S3 has been supported since https://github.com/GothenburgBitFactory/taskchampion/issues/368

taskwarrior is underselling itself :)

travisby:
In #3723 we've added AWS (S3) sync support to taskwarrior.  We even updated most of the docs (all the hard parts, on what actual config settings to set).

In #3921 we noticed one tiny section that still only mentioned GCP support though.  This fixes that!

Fixes #3921

### DIFF
--- a/doc/man/task-sync.5.in
+++ b/doc/man/task-sync.5.in
@@ -44,7 +44,7 @@ periodically, such as via
 Taskwarrior provides several options for synchronizing your tasks:
 
  - To a server specifically designed to handle Taskwarrior data.
- + To a cloud storage provider. Currently only GCP is supported.
+ + To a cloud storage provider. Currently both AWS and GCP are supported.
  - To a local, on-disk file.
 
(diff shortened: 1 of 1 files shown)


djmitche (approved):

djmitche:
Thanks!!

PR #3924, opened 2025-08-03
Add support for s3-compatible storage

travisby:
Utilizes https://github.com/GothenburgBitFactory/taskchampion/pull/601

Will rewrite PR description and move out of draft

### DIFF
--- a/src/commands/CmdSync.cpp
+++ b/src/commands/CmdSync.cpp
@@ -98,9 +98,10 @@ int CmdSync::execute(std::string& output) {
         Context::getContext().config.get("sync.aws.secret_access_key");
     std::string aws_default_credentials =
         Context::getContext().config.get("sync.aws.default_credentials");
-    if (aws_region == "") {
-      throw std::string("sync.aws.region is required");
-    }
+    std::string aws_endpoint_url =
+        Context::getContext().config.get("sync.aws.endpoint_url");
+    bool aws_force_path_style =
+		Context::getContext().config.getBoolean("sync.aws.force_path_style");
     if (encryption_secret == "") {
       throw std::string("sync.encryption_secret is required");
     }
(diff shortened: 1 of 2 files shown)


src/commands/CmdSync.cpp
```diff
@@ -98,9 +98,9 @@ int CmdSync::execute(std::string& output) {
         Context::getContext().config.get("sync.aws.secret_access_key");
     std::string aws_default_credentials =
         Context::getContext().config.get("sync.aws.default_credentials");
-    if (aws_region == "") {
-      throw std::string("sync.aws.region is required");
-    }
+    std::string aws_endpoint_url = Context::getContext().config.get("sync.aws.endpoint_url");
+    bool aws_force_path_style =
+        Context::getContext().config.getBoolean("sync.aws.force_path_style");
```
djmitche, line 103: What value does this need to take when it and `aws_endpoint_url` are omitted? I assume this call will default to `false` when the config is not provided.

djmitche: We'll also need to update the docs as to when `region` is required, and allow it to be empty in other cases.

djmitche (commented):
Looking good. Thanks for hacking on this!

Have you been able to built Taskwarrior with the updated TaskChampion?

travisby:
> Have you been able to built Taskwarrior with the updated TaskChampion?

Yeah!  I had a successful sync to a rook-ceph object store by using the instructions in: https://github.com/GothenburgBitFactory/taskwarrior/blob/bba38b00f8d4735dc260deea3c05a302fb878474/doc/devel/contrib/development.md#using-a-custom-version-of-taskchampion

---

I should be able to get to the feedback today or tomorrow!

djmitche:
I'm really busy too, so take your time but please let me know if you feel like "tomorrow" slips into "never" :)

src/commands/CmdSync.cpp
travisby: Indeed, this defaults to `false`.

I believe AWS will actually work with both `aws_force_path_style = false` and `aws_force_path = true`!

:ack: on the docs here, I need to start editing ./doc/man/task-sync.5

djmitche:
I think #4176 just covered this!

PR #3925, opened 2025-08-10
Update Taskwarrior Docker file

lauft:
This PR mainly wants to upgrade the Taskwarrior Dockerfile to a more recent base image (Ubuntu 24.04).
I upgraded the similar Timewarrior Dockerfile, but this broke the creation of the combined image within the Task-Timewarrior-Hook (👉🏻 https://github.com/GothenburgBitFactory/task-timewarrior-hook/actions/runs/16856764703).

~~@djmitche @felixschurk The upgrade to `ubuntu:24.04` somehow broke the Rust installation.
I succeeded to fix it, but it somehow seems ugly.
As my expertise with Rust is limited, maybe you have some input here.~~

As a side quest, I also modernized the INSTALL document.

### DIFF
--- a/docker/task.dockerfile
+++ b/docker/task.dockerfile
@@ -1,8 +1,8 @@
-FROM ubuntu:22.04 AS base
+FROM ubuntu:24.04 AS base
 
 FROM base AS builder
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
     apt-get install -y \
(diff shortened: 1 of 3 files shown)


docker/task.dockerfile
```diff
@@ -23,14 +23,23 @@ ADD .. /root/code/
 WORKDIR /root/code/
 
 # Setup Rust
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh && \
-    sh rustup.sh -y --profile minimal --default-toolchain stable --component rust-docs
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
djmitche: I'm curious why you removed these arguments? I suspect that without these arguments the toolchain isn't being installed, since `cmake` doesn't call `cargo` but invokes `rustc` directly.

lauft: Well, at first the compilation of Taskwarrior failed when building the Docker image, so I tried to "fix" it. Now I tried the original setup again, and it works... and I have no idea what went wrong the first time. 😳 

So the changes here have become quite minimal.

djmitche (approved):

djmitche:
Thanks! I'll retry the failing CI, which looks like an intermittent.

Issue #3915, opened 2025-07-19
Back up GitHub content


djmitche:
Just for belt-and-suspenders purposes, we should back up all of the useful content on GitHub that is not in the core Git repository -- issues, pull requests comments, release binaries, etc.

I'm sure there's a tool to do this. Backups could be as simple as dumping it all into a Git repository that is pushed to another hosting service.

djmitche:
Once #3929 lands and seems to work, remaining work is
 - Apply this to other repos in the org
 - Consider whether to try to capture additional information (releases, discussions (actually, let's check if these are captured - they are numbered with issues), artifacts)
 - Consider automating a periodic push to another Git hosting service (codeberg?). Various people cloning the repo should be enough to ensure at least one backup exists, but it can't hurt to do better.

djmitche:
This does not appear to be working! This may be partially due to API limits, which may not be an issue if we didn't download the HTML. It fails with logs ending with
```
Done downloading JSON
repo: GothenburgBitFactory/taskwarrior
entries: 3940
Download issues and pull requests HTML sequentially
Error: Process completed with exit code 1.
```

kwoot:
Maybe this is an easy/easier approach? https://docs.codeberg.org/advanced/migrating-repos/#migrating-from-services

djmitche:
Thanks! I don't think we want to migrate away from GitHub (yet, anyway -- a few more Copilot nags and I may change my tune, but hopefully AI will crash soon). It doesn't look like that's something we could easily automate to happen e.g., weekly.

djmitche:
See https://github.com/maxlath/backup-github-repo/pull/13

djmitche:
That wasn't enough -- still exceeding rate limits. I think the GH JS client has a way to add a bit of middleware that watches for rate-limit errors and backs off appropriately. I'll try that.

djmitche:
It wasn't exceeding rate-limits. It just had a bug which maxlath fixed :)

djmitche:
This is now failing with
```
remote: warning: File repo-backup/data.json is 83.35 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB        
```

I think I've put more effort into this than it justifies.

djmitche:
I have reverted the job.

PR #3929, opened 2025-08-18
Perform weekly repo metadata backups

djmitche:
This will backup the issues and pull requests using the format of the API responses for fetching such things.

The tool also fetches HTML images of each issue, but this data is not stable, so would change each week even if no changes occur to the issues. Since we really just want a copy of the data for posterity, the JSON should be sufficient.

This addresses #3915, but does not capture discussions, releases, or release artifacts.

### DIFF
--- a/.github/workflows/backup.yml
+++ b/.github/workflows/backup.yml
@@ -1,0 +1,58 @@
+name: backup
+
+on:
+  schedule:
+    # Mondays are backup days
+    - cron: '0 0 * * 1'
+
+jobs:
+  # This job runs `backup-github-repo` which captures all issues and pull requests,
+  # including their comments. Note that this does not capture discussions or releases.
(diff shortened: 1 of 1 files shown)


djmitche:
I tested this out [over here](https://github.com/djmitche/dervire/commits/main/).

felixschurk (approved):

djmitche:
Hm, that's failing with
`[get request error: will retry in -1s] 403 Forbidden {"message":"API rate limit exceeded for installation. If you reach out to GitHub Support for help, please include the request ID DC01:1849:230CCD7:7E533A9:68B50080 and timestamp 2025-09-01 02:10:08 UTC.","documentation_url":"https://docs.github.com/rest/overview/rate-limits-for-the-rest-api","status":"403"}`

PR #3940, opened 2025-08-29
Support sorting by 'random'

### LINKED ISSUE #3865: Feature Request: Random command
Wouldn't be cool to have the following command: `task random` which basically shows you the random task. Of course we can probably do that with unix commands, but anyway that would be much easier to do with TaskWarrior. 
Also we can get the random task by project: `task random project:foo` 
Any thoughts?

djmitche:
This adds a new 'random' sort key, which is stable within a single run of `task` but will produce a different result for each run. This allows, for example, selecting a random task with `task all sort:random limit:1`.

Fixes #3865. I used Gemini for this, but have reviewed the results carefully.

### DIFF
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -1279,6 +1279,10 @@ grouping.
 A special sort value of "none" indicates that no sorting is required, and tasks
 will be presented in the order (if any) in which they are selected.
 
+A special sort value of "random" indicates that the tasks will be sorted in a
+random order. Each time a report with 'sort:random' is run, a new random
+order will be generated.
+
 .TP
 .B report.X.filter
 This adds a filter to the report X so that only tasks matching the filter
--- a/src/sort.cpp
+++ b/src/sort.cpp
@@ -36,13 +36,16 @@
 #include <util.h>
 
 #include <algorithm>
+#include <functional>
 #include <list>
 #include <map>
+#include <random>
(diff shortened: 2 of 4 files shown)


felixschurk:
Looks good :)
Do we also need to update the autocompletions to catch the new way? Would `sort:ran` already work?

I will try to check it out and test it locally.

felixschurk:
Locally, the test are running fine and everything. However with my personal setup (which should be kind of like the default) I am unable to use the `sort:random` key.

```
❯ ./build/src/task all sort:random
No matches.
```
However, without the `sort:random` I have a list of all the tasks in the db. 

I will try to figure it out over the next days whats going on there..

djmitche:
Sorry, I mislead you! `sort:..` is not a command-line option. Instead, you need to configure a report with this sort. The tests use `rc.report.all.sort:random` and that does work. 

Weirdly, `limit:1` doesn't work (at all, with or without random) with the `all` report by default. I'm not sure what that's about but seems unrelated.

djmitche:
@felixschurk what do you think?

felixschurk:
Sorry, I forget to look at it again. I will do this evening and then report. :)

felixschurk:
Okay, now I managed to check :D
It is working as expected - once its configured as one should :D

realihorrud:
Much appreciated!

PR #3948, opened 2025-09-11
Remove optional spaces before interpreter in shebangs

Necior:
The remaining ~200 shebangs in this repo are space-less.

### DIFF
--- a/.github/workflows/metadata-check.sh
+++ b/.github/workflows/metadata-check.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/bin/bash
 
 # Check the 'cargo metadata' for various requirements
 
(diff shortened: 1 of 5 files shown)


djmitche (approved):
If this distinction is important, should we add some sort of check so that we don't re-introduce spaces?

Necior:
I'm not sure writing a check for that is worth it. This PR is just a consistency thing -- the whitespace after `#!` is optional but it is technically correct.

Having shebangs written the same way across a project makes grepping for patterns easier. For example, I'll be creating a follow-up PR that will avoid `#!/bin/bash`-like shebangs (for reasons that will be explained in that PR) and it's just slightly more convenient to write `grep -rnE '^#!/bin/bash'` instead of `grep -rnE '^#!\s*/bin/bash'`.

djmitche:
Ah, it makes sense that this is leading to a followup :)

PR #3949, opened 2025-09-12
Remove COPYING, use LICENSE only

Necior:
Currently, there are two license files in the repo, namely COPYING and LICENSE. That confuses GitHub's UI that shows "MIT, Unknown licenses found" on the project's page. When one goes into details, LICENSE is recognized as MIT and COPYING is the "unknown" one.


djmitche:
Oh, interesting -- I wonder how things got in that state..

PR #3951, opened 2025-09-13
Use `/usr/bin/env` in shebangs

Necior:
Please see commit descriptions for extra context.

### DIFF
--- a/.github/workflows/metadata-check.sh
+++ b/.github/workflows/metadata-check.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Check the 'cargo metadata' for various requirements
 
--- a/doc/rc/refresh
+++ b/doc/rc/refresh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 for i in holidays*rc
 do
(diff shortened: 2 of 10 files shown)


djmitche (approved):
I didn't realize we had perl scripts!

djmitche:
Would it make sense to add Nix to the GitHub actions? We have a bunch of other distros already.

Necior:
> I didn't realize we had perl scripts!

One can find many surprises by casually browsing project's source code :)

> Would it make sense to add Nix to the GitHub actions? We have a bunch of other distros already.

Hm, maybe? I'm not really familiar with GitHub CI but feel free to file an issue. (Small terminology nit: Nix is a package manager while NixOS is a distro. Some people use Nix on, for example, Debian or macOS.)

Necior:
I compiled Taskwarrior on GhostBSD (a FreeBSD-based OS) today and noticed that this OS also uses different paths for Bash and Perl: `/usr/local/bin/bash` and `/usr/local/bin/perl`.

`/usr/bin/env` is available so this PR would be helpful not only on NixOS.

I updated the PR title from "Use NixOS friendly paths in shebangs" to "Use `/usr/bin/env` in shebangs".

djmitche:
Oh, I'm sorry -- I thought I merged this already. Thanks for the note to bring my attention back to it :)

PR #3953, opened 2025-09-13
Remove asserts that do not necessarily hold

### LINKED ISSUE #3952: Flaky test: random_sort.test.py
Example runs where it failed:

* https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/17700870161/job/50306375493
* https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/17702618062/job/50310350531

Example runs where it succeeded:

* https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/17702618062/job/50310350530
* https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/17702618062/job/50310350529

Necior:
Different seeds might -- but do not have to -- produce different orders.

Similarly, sorting a list by random values might produce the same order as sorting by the ids.

This patch fixes GothenburgBitFactory/taskwarrior#3952.

### DIFF
--- a/test/random_sort.test.py
+++ b/test/random_sort.test.py
@@ -47,36 +47,16 @@ class TestRandomSort(TestCase):
     def test_random_sort_deterministic(self):
         """Verify that 'sort:random' with different seeds produces different orderings."""
         # With a fixed seed, the order is always the same.
-        code, out1_seed1, err = self.t(
+        code, out1_seed, err = self.t(
             "rc.debug.random.seed=123 all rc.report.all.sort:random"
         )
-        code, out2_seed1, err = self.t(
+        code, out2_seed, err = self.t(
             "rc.debug.random.seed=123 all rc.report.all.sort:random"
         )
         self.assertEqual(
-            out1_seed1,
-            out2_seed1,
+            out1_seed,
+            out2_seed,
             "Random sort with the same seed should produce the same order",
-        )
-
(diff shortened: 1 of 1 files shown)


test/random_sort.test.py
```diff
@@ -47,38 +47,18 @@ def setUp(self):
...
-        code, out2_seed1, err = self.t(
+        code, out2_seed, err = self.t(
             "rc.debug.random.seed=123 all rc.report.all.sort:random"
         )
         self.assertEqual(
-            out1_seed1,
-            out2_seed1,
+            out1_seed,
+            out2_seed,
             "Random sort with the same seed should produce the same order",
```
djmitche, line 59: I agree the existing tests were flaky and flaky tests are bad! I thought using a fixed seed would make them deterministic but it seems not.

After removing the second half, this is a pretty weak test! But I suppose it's at least exercising the code so better than nothing..

djmitche (approved):

test/random_sort.test.py
Necior: The code under test uses `std::hash(uuid + seed)`. https://en.cppreference.com/w/cpp/utility/hash.html says:

> Hash functions are only required to produce the same result for the same input within a single execution of a program; this allows salted hashes that prevent collision denial-of-service attacks.

Perhaps a deterministic method for shuffling the tasks could be used in the future.

PR #3954, opened 2025-09-14
Remove `index.html`

Necior:
It's been in its current, almost empty, form since its creation. Nothing in this repo refers to this file.


djmitche (commented):
You are finding some strange things with your tools, `ls` and `cat`!

PR #3968, opened 2025-10-07
Remove test for random sort

djmitche:
This is quite difficult to test, and has caused a lot of flakiness. Let's just get rid of the test.

@Necior I don't think we're getting much value from this test and it's not worth further debugging. What do you think? An example failure is [here](https://github.com/GothenburgBitFactory/taskwarrior/actions/runs/18271028166/job/52013592227?pr=3965).


Necior:
LGTM

PR #3969, opened 2025-10-09
Update CMake minimum required version to 3.10

Daviey:
CMake 4 no longer supports versions older than 3.5, and versions between 3.5-3.10 are deprecated. This PR updates the minimum required version to 3.10 to ensure compatibility with modern CMake versions.

## Background

This issue is already affecting NixOS users (see https://github.com/NixOS/nixpkgs/issues/449826) and will likely impact other distributions as they upgrade to CMake 4.x in the coming months.

## Changes

- Updated `cmake_minimum_required (VERSION 3.0)` to `cmake_minimum_required (VERSION 3.10)`

## Testing

This change has been tested and builds successfully with both CMake 3.x and CMake 4.x.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.10)
 set (CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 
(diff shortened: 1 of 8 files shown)


djmitche:
I think you have an extremely old base version? We made 3.22 the minimum version in 18a8bd2997ddcaa9bfc6154da43c2b650ec8633d (March 2024) and bumped to 3.24 in 236b57f321374f3556aeaf8baf9f231388497fc3 (July 2025).

PR #3976, opened 2025-10-22
feature: update CMakeFiles for corrosion fetch

daeho-ro:
Download `corrosion` only if `SYSTEM_CORROSION` is OFF


djmitche (approved):
That makes sense!

PR #3998, opened 2025-11-24
Allow for arbitrary CMake build tree locations

gonsoos:
* Fix paths based on the assumption that CMAKE_BINARY_DIR is `${CMAKE_SOURCE_DIR}/build`.

Gentoo Linux, for example, builds CMake projects with a directory structure where the source and build trees are side-by-side to test that an out-of-source build is possible.


djmitche (approved):

djmitche:
Thanks!!

PR #4005, opened 2025-12-07
Pass color setting to table component

lauft:
The TTY check has been removed from libshared's `Table` component in [3da15345](https://github.com/GothenburgBitFactory/libshared/commit/3da15345b1bdd2bc69de7231e578bd5d530fc92c).
Instead, the color settings from the main application are passed to it.

### DIFF
--- a/src/commands/CmdCalendar.cpp
+++ b/src/commands/CmdCalendar.cpp
@@ -345,6 +345,7 @@ int CmdCalendar::execute(std::string& output) {
       holTable.width(Context::getContext().getWidth());
       holTable.add("Date");
       holTable.add("Holiday");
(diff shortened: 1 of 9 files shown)


djmitche:
It looks like this is not passing tests?

lauft:
I am still working on it, but I did not want to rely on the runs on my local machine.

lauft:
Test are green now. I had to adapt some test in `stats.test.py` and `commands.test.py` to actually test for ANSI color codes. Those seem to have been suppressed before by the (now removed) `Table`-internal TTY check.

djmitche (approved):
Looks good!

PR #4019, opened 2026-01-05
Update to TaskChampion 3.0.1

djmitche:
The `Replica` type is now generic over the storage, which means supporting both Sqlite and in-memory storage would lead to monomorphization for both types, leading to unnecessary code bloat for a storage type (in-memory) only used in tests. These tests are updated to use an on-disk DB instead.

This uses 3.0.1 to get the fix from https://github.com/GothenburgBitFactory/taskchampion/issues/660.


felixschurk (approved):
Looks good!

PR #4034, opened 2026-01-19
test/README.md: Clarify how to run python tests

ryneeverett:
The changes involved in the switch to cmake/ctest broke the ability to run python tests directly from the source.


djmitche (approved):
Nice, thanks!

PR #4036, opened 2026-01-23
Move fish completion to vendor's directory

### LINKED ISSUE #3023: Fish completions are installed in wrong folder
Taskwarrior installs fish completions in
```
/usr/share/fish/completions
```
As stated in https://fishshell.com/docs/current/completions.html

> A directory for third-party software vendors to ship their own completions for their software, usually /usr/share/fish/vendor_completions.d;

Thus, completions should be installed in `/usr/share/fish/vendor_completions.d/`

Werkov:
I noticed that task completions won't work recently but it can be "fixed" by symlinking the completions scripts from vendor_completions.d directory as described by fish docs [1].
Use this directory by default.

Fixes #3023

Notice: ~Not tested~ CI tested, I'm open to adjustments if this is out of conventional style.

[1] https://fishshell.com/docs/current/completions.html

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -1,6 +1,11 @@
 cmake_minimum_required (VERSION 3.22...4.0)
-install (DIRECTORY bash fish vim hooks
+
+set(FISH_COMPLETIONSDIR share/fish/vendor_completions.d)
+
+install (DIRECTORY bash vim hooks
          DESTINATION ${TASK_DOCDIR}/scripts)
+install (FILES fish/task.fish
+         DESTINATION ${FISH_COMPLETIONSDIR})
 install (FILES zsh/_task
          DESTINATION share/zsh/site-functions)
 install (DIRECTORY add-ons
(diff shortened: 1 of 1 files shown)


Werkov:
This should fix #3023 but some futher work may need to be done in distros (e.g. add dependency on `fish-devel` + not move the completions file in [openSUSE](https://build.opensuse.org/projects/openSUSE:Factory/packages/taskwarrior/files/taskwarrior.spec?expand=1))

djmitche:
This looks OK to me, but I'd appreciate some feedback either from a package maintainer or a fish user before merging it!

Werkov:
Heh, I'm a fish user, moving the completions file [recovers the completions for me](https://github.com/GothenburgBitFactory/taskwarrior/issues/3023#issuecomment-3789670525).

djmitche:
Fair enough. @felixschurk if this looks OK to you, let's merge!

felixschurk:
Yes, looks fine. :)

PR #4113, opened 2026-05-20
Add iterative task handling to TaskWarrior

carmiac:
Most of the changes are moving some of the task handling from raw TaskChampion TaskData to the Task interface.


djmitche:
Since it will take a bit to staibilize the proposed TaskChampion changes, what do you think of extracting some of the refactors into separate PRs that we could land earlier?

carmiac:
Sure. There's basically four groups of changes:
- Adding the partial Task FFI
- Moving some commands from raw TaskData -> Task
- Moving some built in reports from `status:pending` to `+PENDING`
- Adding `Iterative` handling. 

If I were to break this up, I'd probably want to make the first two more complete as separate PRs and see just how much can be moved from TaskData to Task. Make sense?

djmitche:
Makes lots of sense, yep!

PR #4117, opened 2026-06-03
Add a (mostly) full Task FFI from TaskChampion

carmiac:
This adds a mostly complete Task FFI from TaskChampion. It also adds the Replica functions dealing with Tasks.

It does not include:

- deprecated functions or features
- `Task::into_task_data`
- `Replica::dependancy_map` 

Nor does it actually connect any of this to the TaskWarrior logic.

### DIFF
--- a/src/taskchampion-cpp/src/lib.rs
+++ b/src/taskchampion-cpp/src/lib.rs
@@ -137,6 +137,24 @@ mod ffi {
         /// Get the operations for a task task by its UUID.
         fn get_task_operations(&mut self, uuid: Uuid) -> Result<Vec<Operation>>;
 
+        /// Create a new `Task` with the given UUID. If a task with that
+        /// UUID already exists, the existing task is returned.
+        fn create_task(&mut self, uuid: Uuid, ops: &mut Vec<Operation>) -> Result<Box<Task>>;
+
+        /// Get an existing `Task` by UUID. Returns None if no task with that UUID exists.
+        fn get_task(&mut self, uuid: Uuid) -> Result<OptionTask>;
+
+        /// Get all `Task` values in the replica.
+        ///
+        /// This contains `OptionTaskData` to allow C++ to `take` values out of the vector and use
+        /// them as `rust::Box<TaskData>`. Cxx does not support `Vec<Box<_>>`. Cxx also does not
+        /// handle `HashMap`, so the result is not a map from uuid to task. The returned Vec is
+        /// fully populated, so it is safe to call `take` on each value in the returned Vec once .
+        fn all_tasks(&mut self) -> Result<Vec<OptionTask>>;
+
+        /// Simiar to all_task_data, but returing only pending tasks (those in the working set).
+        fn pending_tasks(&mut self) -> Result<Vec<OptionTask>>;
+
         /// Return the operations back to and including the last undo point, or since the last sync if
         /// no undo point is found.
         fn get_undo_operations(&mut self) -> Result<Vec<Operation>>;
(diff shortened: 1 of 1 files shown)


djmitche:
I apologize for my tardiness in looking at this!

src/taskchampion-cpp/src/lib.rs
```diff
@@ -276,6 +294,194 @@ mod ffi {
...
+        /// Mark the task as started.
+        fn start(self: &mut Task, ops: &mut Vec<Operation>) -> Result<()>;
+        /// Mark the task as stopped.
+        fn stop(self: &mut Task, ops: &mut Vec<Operation>) -> Result<()>;
+        /// Mark the task as completed.
+        fn done(self: &mut Task, ops: &mut Vec<Operation>) -> Result<()>;
+
+        /// Add a tag to the task. The tag is parsed from a string, so can fail.
+        fn add_tag(self: &mut Task, tag: &CxxString, ops: &mut Vec<Operation>) -> Result<()>;
+        /// Remove a tag from the task. The tag is parsed from a, so can fail.
```
djmitche:
```suggestion
        /// Remove a tag from the task. The tag is parsed from a string, so can fail.
```

src/taskchampion-cpp/src/lib.rs
```diff
@@ -137,6 +137,24 @@ mod ffi {
...
+        /// UUID already exists, the existing task is returned.
+        fn create_task(&mut self, uuid: Uuid, ops: &mut Vec<Operation>) -> Result<Box<Task>>;
+
+        /// Get an existing `Task` by UUID. Returns None if no task with that UUID exists.
+        fn get_task(&mut self, uuid: Uuid) -> Result<OptionTask>;
+
+        /// Get all `Task` values in the replica.
+        ///
+        /// This contains `OptionTaskData` to allow C++ to `take` values out of the vector and use
+        /// them as `rust::Box<TaskData>`. Cxx does not support `Vec<Box<_>>`. Cxx also does not
```
djmitche:
```suggestion
        /// This contains `OptionTask` to allow C++ to `take` values out of the vector and use
        /// them as `rust::Box<Task>`. See the comments on `all_task_data`.
```

src/taskchampion-cpp/src/lib.rs
```diff
@@ -899,7 +1161,343 @@ impl TaskData {
...
+
+    fn get_priority(&self) -> String {
+        self.0.get_priority().into()
+    }
+    fn get_entry(&self) -> i64 {
+        if let Some(t) = self.0.get_entry() {
+            t.timestamp()
+        } else {
+            0
+        }
```
djmitche: A helper function for this -- inverse of optional_timestamp -- might be useful.

src/taskchampion-cpp/src/lib.rs
```diff
@@ -899,7 +1161,343 @@ impl TaskData {
...
+        &mut self,
+        status: ffi::Status,
+        ops: &mut Vec<Operation>,
+    ) -> Result<(), CppError> {
+        let status = match status {
+            ffi::Status::Pending => tc::Status::Pending,
+            ffi::Status::Completed => tc::Status::Completed,
+            ffi::Status::Deleted => tc::Status::Deleted,
+            ffi::Status::Recurring => tc::Status::Recurring,
+            _ => unreachable!("ffi::Status variants are exhaustive"),
```
djmitche: This isn't actually unreachable, though -- `t.set_status(ffi::Status::Unknown, ops)` will reach this point. So I don't think it should panic. It might make sense to treat this as an error, since we can't round-trip the `Unknown(value)` variant from Rust.

src/taskchampion-cpp/src/lib.rs
```diff
@@ -1220,4 +1818,182 @@ mod test {
...
+        let mut rep = new_replica_on_disk(path, true, true).unwrap();
+        let uuid = uuid_v4();
+        let mut ops = new_operations();
+        rep.create_task(uuid, &mut ops).unwrap();
+        rep.commit_operations(ops).unwrap();
+        let task = rep.get_task(uuid).unwrap().take();
+        (rep, task)
+    }
+
+    #[test]
```
djmitche, line 1826: Woo, tests!

djmitche (commented):
Again, I'm sorry for taking well over a month to review this. It looks great, with just a few very minor updates below. Otherwise I'm happy with this, but I would like to understand what the next step is before doing so. What's the benefit of this approach over the lower-level TaskData type?

carmiac:
> Again, I'm sorry for taking well over a month to review this. It looks great, with just a few very minor updates below. Otherwise I'm happy with this, but I would like to understand what the next step is before doing so. What's the benefit of this approach over the lower-level TaskData type?

Oh I get it, it's been a rough month for me.  This is one part of splitting up https://github.com/GothenburgBitFactory/taskwarrior/pull/4113 

Iterative tasks need to be handled by `Task::set_status` and other `Task` functions, rather than directly manipulating TaskData so that they are always handled the same way.  I threw in the rest of the Task interface since I was already there.

djmitche:
Gotcha - thanks!

djmitche (approved):
Awesome, thanks!

PR #4125, opened 2026-06-13
fix: don't round urgency in _urgency command

Ram-Z:
Rounding causes sorting output of _urgency to not match the sorting of any reports.


djmitche (approved):
Looks good - any chance you could add a test, at least for the sorting?

Ram-Z:
I don't think that `_urgency` allows sorting, what I was referring to was sorting it externally, i.e. `task _urgency | sort -k4n`.

djmitche:
Got it, thanks!

PR #4136, opened 2026-06-27
Burndown performance overhaul

### LINKED ISSUE #3412: slow burndown.daily report
### Discussed in https://github.com/GothenburgBitFactory/taskwarrior/discussions/3307

<div type='discussions-op-text'>

<sup>Originally posted by **gour** March 28, 2024</sup>
Hello!

While anticipating 3.0 to arrive into my distribution, I wonder if it should improve time to complete `burndown` reports?

For instance, here is what I currently get wit h2.6.2:

```
Executed in  167.89 secs    fish           external
   usr time   56.11 secs  307.00 micros   56.11 secs
   sys time  111.29 secs  144.00 micros  111.29 secs
```

which is useless for some interactive shell scripting...
 </div>

ashprice:
Closes https://github.com/GothenburgBitFactory/taskwarrior/issues/3412 (planned PRs will likely help as well.)

# Overview of changes

I have tried to make the code as self-documenting as possible with comments. This said, here is a tldr:
- The original `scan()` function used loops that checked `while (from < now)` on every period between a task's entry date and the present, which is slow. Now we pre-quantize tasks into epochs, and then use a map to look them up instead of scanning.
- The original burndown report, as far as I understand, was DST-unsafe. This has been fixed (as far as I can tell). Tested using `libfaketime`. No more gaps if you just so happen to be using TW at 1:30AM on 2 days of the year. 
- Quantization is now done via two routes. First, we count tasks for the bars using the quantization of the chart mode (daily/weekly/monthly). Second, we track the peak count quantized by day (because the fix rate is always calculated by day). Although the peak count is quantized by day, it should be noted that the final sum calculates against an unquantized `now` - ie. the hours, minutes, and seconds since midnight also count. 
- Peak calculation uses a difference array which is prefix-summed in the `finalize()` function. 
- Bars are iterated based on the bars that are visible, not the _n_k possible epochs. 
- Ultimately two passes made sense, so we do quantization once and cache it. 
- Threw in an annual burndown report, because why not. Looks a bit silly though, should probably change it to show less bars if there's nothing in them. 

There are quite a lot of changes here, so please ask questions if anything is unclear. My first attempt at this caused a lot of regressions, as far as I can tell the report is equivalent in functionality now and is lacking in bugs. The only difference from the previous report on the user-side is that the fix rate can be rounded slightly differently, leading to a single decimal point difference.

### DIFF
--- a/doc/man/task.1.in
+++ b/doc/man/task.1.in
@@ -166,6 +166,10 @@ the 'burndown.weekly' report. Is affected by the context.
 Shows a graphical burndown chart, by month. Is affected by the context.
 
 .TP
+.B task <filter> burndown.annual
+Shows a graphical burndown chart, by year. Is affected by the context.
+
+.TP
 .B task calendar [due|<month> <year>|<year>] [y]
 Shows a monthly calendar with due tasks marked.  Shows one horizontal line of
 months.  If the 'y' argument is provided, will show at least one complete year.
--- a/src/commands/CmdBurndown.cpp
+++ b/src/commands/CmdBurndown.cpp
@@ -25,21 +25,19 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include <cmake.h>
-// cmake.h include header must come first
-
+// cmake.h include must come first.
 #include <CmdBurndown.h>
 #include <Context.h>
 #include <Datetime.h>
 #include <Duration.h>
 #include <Filter.h>
 #include <format.h>
-#include <math.h>
-#include <shared.h>
 #include <string.h>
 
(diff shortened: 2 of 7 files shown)


djmitche (approved):
This looks great!

I don't profess to understand the algorithm at any level of detail, but nothing here has me worried. And I can confirm that not only do tests pass, but this renders the same burndown as the latest release version for my personal task DB (and does so much more quickly).

There are a bunch of C++-level fixes and modernizations mixed in here, too, and those all look great. I especially appreciate the use of closures for simplifying the core of the implementation.

PR #4157, opened 2026-07-23
use TaskChampion Task when creating/setting setting status

carmiac:
This is part 3 of the split of https://github.com/GothenburgBitFactory/taskwarrior/pull/4113

### DIFF
--- a/src/TDB2.cpp
+++ b/src/TDB2.cpp
@@ -51,8 +51,8 @@ static DependencyGraph build_dependency_graph(const std::vector<Task>&,
                                               const std::unordered_map<std::string, size_t>&);
 
 ////////////////////////////////////////////////////////////////////////////////
-// Map the C++ on-disk status string to the typed `tc::Status` enum used by the
-// taskchampion-cpp bridge.
+// Map the C++ representation of status string to the typed `tc::Status` enum
+// used by the taskchampion-cpp bridge.
 static tc::Status statusFromString(const std::string& s) {
   if (s == "pending") return tc::Status::Pending;
   if (s == "completed") return tc::Status::Completed;
@@ -153,8 +153,7 @@ void TDB2::modify(Task& task) {
   }
   auto tctask = maybe_tctask.take();
 
-  std::string deferred_status;
-  bool deferred_status_changed = false;
+  std::optional<std::string> deferred_status;
   std::unordered_set<std::string> seen;
   for (auto k : task.all()) {
     // ignore task keys that aren't stored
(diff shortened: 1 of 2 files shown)


src/TDB2.cpp
```diff
@@ -50,6 +50,17 @@ static void dependency_scan(std::vector<Task>&, const std::unordered_map<std::st
 static DependencyGraph build_dependency_graph(const std::vector<Task>&,
                                               const std::unordered_map<std::string, size_t>&);
 
+////////////////////////////////////////////////////////////////////////////////
+// Map the C++ on-disk status string to the typed `tc::Status` enum used by the
```
djmitche:
```suggestion
// Map the C++ representation of status to the typed `tc::Status` enum used by the
```

src/TDB2.cpp
```diff
@@ -128,48 +147,51 @@ void TDB2::modify(Task& task) {
...
+  auto maybe_tctask = replica()->get_task(tcuuid);
   if (maybe_tctask.is_none()) {
     throw std::string("task no longer exists");
   }
   auto tctask = maybe_tctask.take();
 
-  // Perform the necessary `update` operations to set all keys in `tctask`
-  // equal to those in `task`.
+  std::string deferred_status;
+  bool deferred_status_changed = false;
```
djmitche: You could use `std::optional<std::string>` here instead of two variables

djmitche (commented):
This looks good - just some minor suggestions. And as always I'm sorry for getting to it so late.

src/TDB2.cpp
carmiac: done!

djmitche (approved):
Thanks!

djmitche:
Hm, it might be worth increasing the timeout for that test:
```patch
diff --git test/ids.test.py test/ids.test.py
index 4dc30935e6..39826696a4 100755
--- test/ids.test.py
+++ test/ids.test.py
@@ -127,17 +127,17 @@
     def generate_tasks(self, n):
         """Generates n tasks for testing purposes"""
         with tempfile.NamedTemporaryFile(mode="w") as f:
             f.write(
                 "\n".join([f'{{"description": "test task {i+1}"}}' for i in range(n)])
             )
             f.flush()
             # use a long timeout here, because import is quite slow
-            code, out, err = self.t(f"import {f.name}", timeout=100)
+            code, out, err = self.t(f"import {f.name}", timeout=300)
 
     def test_single_digit_range(self):
         """Test that parsing single digit ID range works"""
         self.generate_tasks(20)
         code, out, err = self.t("4-6 count")
         self.assertEqual("3", out.strip())
 
     def test_double_digit_range(self):
```
this is set to auto-merge so once that tests passes it should merge.

PR #4163, opened 2026-07-31
Fix build on Haiku

carmiac:
I've started playing around with HaikuOS again, so I'll be opening a small series of PRs. This one gets taskwarrior to build on Haiku, tests on a recent nightly and also on linux to make sure it didn't break anything. With this, it passes 166 of the 178 tests on Haiku.  I'll be putting together a PR to deal with those failures separate.


djmitche (approved):
Nice! Is it possible to add GH support for testing this?

PR #4164, opened 2026-08-01
Fix config paths on Haiku

carmiac:
This fixes the default config and data locations in Haiku, which uses a non-standard (but reasonably sane) location for such things and provides a nice API for getting them.

Also, this fixes two call sites that weren't getting the default correctly. All of the other calls that I found already use this format.

I'm still working on a CI build/test for Haiku.


djmitche (approved):
Nice!

PR #4167, opened 2026-08-14
feat: add annotation.info setting

### LINKED ISSUE #4158: info: annotations render both under Description and in the journal table... intentional? (proposal: an annotation.info-style toggle mirroring journal.info)
**Taskwarrior version:** 3.4.2

## Question

In `task <id> info`, a task's annotations are rendered **twice**:

1. Appended as indented dated lines under the **Description** cell.
2. Again in the bottom **Date | Modification** journal table (as `Annotation of '<full text>' added` rows), when `journal.info=1`.

For a task with many annotations this produces a large duplicated block. Minimal repro:

```
$ task add "Example task"
$ task 1 annotate "first note"
$ task 1 annotate "second note"
$ task 1 info
```

The two annotations appear once under Description and again in the journal table below.

Is the under-Description rendering in `info` intentional? I couldn't find documentation explaining its purpose, and for tasks used as small "projects" (where annotations are a running log), the duplication is a lot of visual noise.

## What I found

- `journal.info=0` removes the bottom journal table but keeps the under-Description annotations. There is no symmetric setting for the reverse.
- `indent.annotation` and `dateformat.annotation` affect only formatting, not inclusion.
- `information` is a static, non-modifiable report, so `report.info.*` does not apply.

So the two renderings are not independently controllable today.

## Proposal (RFE)

Add a boolean config, e.g. `annotation.info` (default `on` for backward compatibility), that gates the under-Description annotation block in the `info`/`information` output, mirroring the existing `journal.info` pattern that already gates the journal table. This would let users pick either representation without losing the other.

Possibly related: #3914 (annotation indentation in reports) touches the same rendering area.

Happy to help test or take a look at the change if there is interest.

floze-the-genius:
## Summary

- add an `annotation.info` setting that controls whether annotations appear below the Description field in `task info`
- keep the existing output by default and leave the journal table independently controlled by `journal.info`
- document the setting and recognize it in `task show` and the Vim taskrc syntax

## Tests

- `cmake --build build-4158 --target task_executable test_runner -j 6`
- `ctest --test-dir build-4158 -R '^(info|annotate|show)\.test\.py$' --output-on-failure` (3/3 passed)
- `ctest --test-dir build-4158 -j 6 --output-on-failure` (178/178 passed)

Closes #4158

## AI assistance

Developed with assistance from OpenAI Codex. I reviewed and tested the resulting changes.

### DIFF
--- a/doc/man/taskrc.5.in
+++ b/doc/man/taskrc.5.in
@@ -362,6 +362,11 @@ Setting this to '0' means that it is an error to use a write command with no
 filter.
 
 .TP
+.B annotation.info=1
+Determines whether annotations are displayed below the description field by the
+'info' command. The default value is "1".
+
+.TP
 .B indent.annotation=2
 Controls the number of spaces to indent annotations when shown beneath the
 description field. The default value is "2".
--- a/scripts/vim/syntax/taskrc.vim
+++ b/scripts/vim/syntax/taskrc.vim
@@ -39,6 +39,7 @@ syn match taskrcGoodKey '^\s*\Vabbreviation.minimum='he=e-1
 syn match taskrcGoodKey '^\s*\Vactive.indicator='he=e-1
 syn match taskrcGoodKey '^\s*\Valias.\S\{-}='he=e-1
 syn match taskrcGoodKey '^\s*\Vallow.empty.filter='he=e-1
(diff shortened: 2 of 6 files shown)


djmitche (approved):
Very nice!

PR #4173, opened 2026-08-17
scripts: install bash completion just like fish

doronbehar:
Noticed you finally started installing fish completion properly, so I thought it would make sense to install the Bash completion properly too.


djmitche (approved):
Thanks!

Issue #4178, opened 2026-08-22
modify depends reads a database fully per each non-pending task


9484e0dd:
Not sure how much this is related to #4121, and if this will be of any interest to @ashprice.

`modify depends` reads a database fully per each non-pending task.

`dependencyIsCircular()` calls `tdb2.get(dep, current)` for each depends edge.

Since #4127, `TDB2::get()` resolves pending UUIDs through `_pending_index`, but a non-pending UUID requires `all_task_data()`.

My database has 1217 such edges from completed tasks out of 2087. Total size is 5578 tasks, 54 MB. Adding one dependency to a task with 69 completed tasks referenced through `depends` peaks at ~1 GB RSS over ~9.5 s. I suppose nothing is freed between dependency modifications in a command: one dependency 1005 MB, two 1935 MB, three 2945 MB. Fifty-four dependencies in a single modify therefore requires ~50 GB, which bricks my machine.

The script below creates a database with completed chains and a pending control, then measures a `modify depends:` against each.

```
import json
import os
import subprocess
import tempfile
import time
import uuid

TASK_BIN = os.environ.get("TASK", "task")
FILLER_COUNT = int(os.environ.get("FILLER", 2000))
ENTRY_DATE = "20200101T000000Z"


def new_uuid():
    return str(uuid.uuid4())


all_tasks = []
chain_heads = []


# Linear chain
def add_chain(length, status, name):
    link_uuids = []
    for i in range(length):
        link_uuids.append(new_uuid())

    for i in range(length):
        task = {
            "uuid": link_uuids[i],
            "description": name + str(i),
            "status": status,
            "entry": ENTRY_DATE,
        }
        if status == "completed":
            task["end"] = ENTRY_DATE
        if i + 1 < length:
            task["depends"] = [link_uuids[i + 1]]
        all_tasks.append(task)

    head_uuid = new_uuid()
    all_tasks.append({
        "uuid": head_uuid,
        "description": name,
        "status": "pending",
        "entry": ENTRY_DATE,
        "depends": [link_uuids[0]],
    })
    chain_heads.append((name, length, status, head_uuid))


for length in (10, 20, 40, 80):
    add_chain(length, "completed", "done" + str(length))
add_chain(80, "pending", "live80")

spare_uuid = new_uuid()
all_tasks.append({
    "uuid": spare_uuid,
    "description": "SPARE",
    "status": "pending",
    "entry": ENTRY_DATE,
})

# Filler tasks
[...]

PR #4179, opened 2026-08-24
Use the pending tasks cache in the circular dependency check.

ashprice:
Introduced ahead of the other work to address #4178 . 

Main changes:
1. the dependency graph is traversed using a stack of UUIDs.
2. The traversal first looks to the modified task, so that modified dependencies are included. 
3. Other dependencies are all checked against the cache with `find_pending()`.
4. Traversal stops at completed, deleted, or missing dependencies.
5. Visited UUIDs are marked, so each dependency is only visited once.
6. The cache is initialized where necessary.

Three tests are introduced:
- `test_completed_not_circular()` checks that completed tasks do not cause a positive circularity check.
- `test_circular_with_visited_sibling()` makes sure that visited siblings do not get overwritten when the unvisited sibling is popped. 
- `test_dependency_modify_with_cold_pending_cache()` checks that the pending cache is initialized during modifications. It uses an empty filter because most modify commands warm the cache, but this one doesn't. 

There are still remaining issues with empty filters, bulk modifications of completed tasks, and bulk modifications of dependencies in other regards, I plan to address these in a future PR.  

Also, this won't bring broader performance improvement on its own, because it doesn't touch the stuff relevant for eg. `task ready`.

### DIFF
--- a/src/dependency.cpp
+++ b/src/dependency.cpp
@@ -33,48 +33,44 @@
 #include <shared.h>
 
 #include <iostream>
-#include <stack>
+#include <vector>
 
 #define STRING_DEPEND_BLOCKED "Task {1} is blocked by:"
 
 ////////////////////////////////////////////////////////////////////////////////
 // Returns true if the supplied task adds a cycle to the dependency chain.
+// Only dependencies within the pending task cache are traversed.
 bool dependencyIsCircular(const Task& task) {
   // A new task has no UUID assigned yet, and therefore cannot be part of any
   // dependency chain.
-  if (task.has("uuid")) {
-    auto task_uuid = task.get("uuid");
+  if (!task.has("uuid")) return false;
 
-    std::stack<Task> s;
-    s.push(task);
+  const auto& task_uuid = task.get_ref("uuid");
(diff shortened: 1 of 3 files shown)


src/dependency.cpp
```diff
@@ -33,48 +33,44 @@
...
-          }
-        }
-      }
+  if (!task.has("uuid")) return false;
+
+  const auto& task_uuid = task.get_ref("uuid");
+  auto& tdb2 = Context::getContext().tdb2;
+
+  std::unordered_set<std::string> visited{task_uuid};
+  std::vector<std::string> stack;
```
djmitche, line 52: I'm curious why the switch from stack to vector?

djmitche (approved):
This looks great!

src/dependency.cpp
ashprice: If there was a reason for this I have forgotten it, I guess I thought it might be more memory efficient at the time but I don't see any reason to think that. 

I wrote the original code this PR was based on in early July or late June so I might have had another reason and forgotten about it, it seems I had originally used stack.

PR #4190, opened 2026-09-14
Fix typo “depenency” in summary message

swardman:
When removing a dependency from a task the message “Depenency on {taskid} will be removed” is displayed. This fixes the spelling. This appears to be the only occurrence.


djmitche:
Thanks! My tping is terrible.

